PR #12, opened 2019-11-18
Work towards creating first Isolate (#12)

piscisaureus:


.vscode/c_cpp_properties.json
```diff
@@ -0,0 +1,22 @@
+{
```
ry: Does this file have some function?

PR #20, opened 2019-11-30
make internal symbols private

ry:


ry:
closed in favor of #21 ... windows is doing an uncached rebuild when the version is bumped.

PR #38, opened 2019-12-05
Fix mac build

ry:


piscisaureus:
@ry 
FYI -- it turns out that my windows system also requires `symbol_level = 1`, otherwise the linker produces a corrupt executable.
I have no idea why this is not an issue on CI but I can't seem to find a way around it.

ry:
Let's just revert all the changes I made to .gn 

They were arrived through much trial and error and I was hoping it could just be simplified - but apparently not. We can try to simplify in the future. It's really not so important right now.

PR #39, opened 2019-12-05
add String::length method

bartlomieju:


src/string.rs
```diff
@@ -78,6 +80,10 @@ impl String {
     }
   }
 
+  pub fn length(&self) -> usize {
```
ry: Can you copy the docs from v8.h here as a triple slash comment? I know this isn't sustainable - and that we must automate this somehow - but documentation in docs.rs will improve the quality of this binding

https://cs.chromium.org/chromium/src/v8/include/v8.h?l=2899&rcl=69827db645fcece065bf16a795a4ec8d3a51057f

PR #45, opened 2019-12-07
add v8::Context::Scope

bartlomieju:
I'm not really sure that porting `v8::Context::Scope` is needed... wouldn't it be equivalent to do it like this?

```rust
impl Context {
  pub fn new<'sc>(scope: &mut HandleScope<'sc>) -> Local<'sc, Context> {
    // TODO: optional arguments;
    let context = unsafe { Local::from_raw(v8__Context__New(scope.cxx_isolate())).unwrap() };
    unsafe { v8__Context__Enter(&mut context) };
    context
  }
}

impl Drop for Context {
  pub fn drop(&mut self) {
    unsafe { v8__Context__Exit(self) };
  } 
}
```
then the test could be rewritten as:
```rust
#[test]
fn script_compile_and_run() {
  setup();
  let mut params = v8::Isolate::create_params();
  params.set_array_buffer_allocator(
    v8::array_buffer::Allocator::new_default_allocator(),
  );
  let mut isolate = v8::Isolate::new(params);
  let mut locker = v8::Locker::new(&mut isolate);

  v8::HandleScope::enter(&mut locker, |s| {
    let context = v8::Context::new(s);
    let source =
      v8::String::new(s, "'Hello ' + 13 + 'th planet'", Default::default())
        .unwrap();
    let mut script = v8::Script::compile(s, context, source, None).unwrap();
    source.to_rust_string_lossy(s);
    let result = script.run(s, context).unwrap();
    // TODO: safer casts.
    let result: v8::Local<v8::String> =
      unsafe { std::mem::transmute_copy(&result) };
    assert_eq!(result.to_rust_string_lossy(s), "Hello 13th planet");
    drop(context);
  });
}
```

The same could be applied to `HandleScope` I guess?

EDIT: I guess it's a good idea after all.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -32,13 +32,11 @@ impl Context {
 
   pub fn scope<'sc>(
     handle_scope: &mut HandleScope<'sc>,
-    mut f: impl FnMut(Local<Context>, &mut HandleScope<'sc>) -> (),
+    mut f: impl FnMut(&mut HandleScope<'sc>, Local<Context>) -> (),
   ) {
-    let mut context = unsafe {
-      Local::from_raw(v8__Context__New(handle_scope.cxx_isolate())).unwrap()
-    };
+    let mut context = Context::new(handle_scope);
     context.enter();
-    f(context, handle_scope);
+    f(handle_scope, context);
     context.exit();
   }
 }
--- a/src/context.rs
+++ b/src/context.rs
@@ -29,4 +29,16 @@ impl Context {
     // TODO: enter/exit should be controlled by a scope.
     unsafe { v8__Context__Exit(self) };
   }
+
+  pub fn scope<'sc>(
+    handle_scope: &mut HandleScope<'sc>,
+    mut f: impl FnMut(Local<Context>, &mut HandleScope<'sc>) -> (),
+  ) {
+    let mut context = unsafe {
+      Local::from_raw(v8__Context__New(handle_scope.cxx_isolate())).unwrap()
+    };
+    context.enter();
+    f(context, handle_scope);
+    context.exit();
+  }
 }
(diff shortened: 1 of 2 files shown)


piscisaureus:
I'm not sure yet I want to continue using closures for controlling the lifetime of C++ RAII objects.
As such this seems acceptable, but V8 has many more RAII types and they are not necessarily entered or used in a fixed order.
I also want to avoid a situation where we end up with something like:

```rust
MicroTaskPolicy::scope(MicroTaskPolicy::AUTO, |isolate, context, try_catch, backup_incumbent, micro_task_policy| {
  // your code here
});
```

which seems pretty unwieldy.

piscisaureus:
So I haven't found the ideal pattern for dealing with RAII objects yet, and I want to punt on it until we have at least one more type that we need (which I think is going to be TryCatch). Which would have us support the following ones:

* Locker (this one is pretty trivial)
* HandleScope
* TryCatch
* Context::Scope

ry:
> I'm not sure yet I want to continue using closures for controlling the lifetime of C++ RAII objects 

I'm also in favor of removing the closure helpers for RAII. It seems to go against the principal of exposing a Rust V8 API as close as possible to the C++ API. Closures for lifetime management is neither used in C++, nor is it particularly Rusty.

piscisaureus:
> I'm also in favor of removing the closure helpers for RAII bindings. It seems to go against the principal of exposing a Rust V8 API as close as possible to the C++ API. Closures for lifetime management is neither used in C++, nor is it particularly Rusty.

Just FYI, it was never my favorite solution, but it is not obvious how to do it in rust without violating safety rules.

ry:
> but it is not obvious how to do it in rust without violating safety rules.

 I hadn't considered that... That's a good reason to keep them.

bartlomieju:
> So I haven't found the ideal pattern for dealing with RAII objects yet, and I want to punt on it until we have at least one more type that we need (which I think is going to be TryCatch). Which would have us support the following ones:
> 
> * Locker (this one is pretty trivial)
> * HandleScope
> * TryCatch
> * Context::Scope

I'm in favor :+1:

piscisaureus:
So long story short, I'll take this issue.

PR #53, opened 2019-12-09
Add v8::MaybeLocal

bartlomieju:


bartlomieju:
Meh, I see we already have `maybe_local_to_ptr` and `ptr_to_maybe_local` and use `Option<Local<>>` to handle that :+1:

PR #56, opened 2019-12-10
add v8::Promise

bartlomieju:


src/support.h
```diff
@@ -60,4 +60,9 @@ inline static v8::MaybeLocal<T> ptr_to_maybe_local(T* ptr) {
   return *reinterpret_cast<v8::MaybeLocal<T>*>(&ptr);
 }
 
+template <class T>
+inline static T maybe_to_value(v8::Maybe<T> maybe, T default_value) {
+  return maybe.FromMaybe(default_value);
+}
+
```
bartlomieju: @ry @piscisaureus could you take a look at this `v8::Maybe` type? I'm not sure I handled it correctly

src/promise.rs
```diff
@@ -0,0 +1,171 @@
...
+
+  /// Returns true if the promise has at least one derived promise, and
+  /// therefore resolve/reject handlers (including default handler).
+  pub fn has_handler(&mut self) -> bool {
+    unsafe { v8__Promise__HasHandler(&mut *self) }
+  }
+
+  /// Returns the content of the [[PromiseResult]] field. The Promise must not
+  /// be pending.
+  pub fn result<'sc>(&mut self) -> Local<'sc, Value> {
```
piscisaureus: There are several methods on Promise that return a Local handle, such as this one.
With local handles you have to be careful that their lifetime is properly bound to the HandleScope in which they live.
This function doesn't do that. `'sc` is effectively an unbound lifetime here and thus the returned Local will happily outlive its HandleScope, which is unsafe.

You can fix this by adding a HandleScope parameter to the function signature (even if the argument is unused).
So:
`pub fn result<'sc>(&mut self, _scope: &mut HandleScope<'sc>) -> Local<'sc, Value> {`

src/support.h
ry: Looks good to me...

PR #57, opened 2019-12-10
Add v8::Function

bartlomieju:


bartlomieju:
FYI `FunctionCallbackInfo` is half-baked any doesn't support anything besides getting number of args passed.

Issue #51, opened 2019-12-08
Need automated conversion for Local<String> to Local<Value>


ry:
Here is an attempted solution which doesn't quite work.

```rust
/// For example, to allow automated conversion from Local<String> to
/// Local<Value>
impl<'sc, T, U> From<Local<'sc, T>> for Local<'sc, U>
where
  T: Deref<Target = U>,
{
  fn from(v: Local<'sc, T>) -> Local<'sc, U> {
    unsafe { std::mem::transmute(v) }
  }
}
```

In C++ this is done by this bit of code: https://github.com/v8/v8/blob/63da8397bc967232f91fea07b1861a50d20ca717/include/v8.h#L192-L201

manyuanrong:
I tried all kinds of ways. It seems that they are not feasible. I gave up for the time being.
Do we currently implement "From trait" for each "Local<T>" struct?

manyuanrong:
We can use `macro_rules` to explicitly define the conversion relationships for these subtypes
```rust
macro_rules! impl_into_type {
    ($child:ident, $parent:ident) => {
        impl From<$child> for $parent {
            fn from(child: $child) -> $parent {
                unsafe { mem::transmute(child) }
            }
        }
    }
}
```

```rust
impl_into_type!(String, Value);
impl_into_type!(Boolean, Value);
impl_into_type!(Interger, Value);
```

ry:
https://github.com/denoland/rusty_v8/commit/9a72f62bd661890088b1ea582e1f70ddd9fd76bd

PR #61, opened 2019-12-13
automatically convert from Local<T> to Local<Value>

afinch7:
This seems to work fine, but it's not exactly ideal. The rust compiler doesn't allow me to be more implicit  here sadly.

ref #51

### DIFF
--- a/src/local.rs
+++ b/src/local.rs
@@ -33,6 +33,8 @@ impl<'sc, T> DerefMut for Local<'sc, T> {
   }
 }
 
+// TODO make it possible for targets other than Local<Value>. For example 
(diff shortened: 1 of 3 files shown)


src/local.rs
```diff
@@ -31,3 +32,12 @@ impl<'sc, T> DerefMut for Local<'sc, T> {
     unsafe { self.0.as_mut() }
   }
 }
+
+impl<'sc, T> From<Local<'sc, T>> for Local<'sc, Value>
```
ry: This is certainly an improvement - but please add a TODO to make it possible for targets other than `Local<Value>`.. For example `Local<String>` should be able to be down cast to `Local<Name>`.

PR #63, opened 2019-12-14
fix: maybe_to_value -> maybe_to_ptr

bartlomieju:
Discussed with @piscisaureus, `Maybe<T>` should be returned as `Option<T>` instead of casting it to default value.

### DIFF
--- a/src/support.h
+++ b/src/support.h
@@ -64,6 +64,7 @@ template <class T>
 inline static T* maybe_to_ptr(v8::Maybe<T> maybe) {
   T* ptr = new T();
   if (maybe.IsJust()) {
+    delete ptr;
     *ptr = maybe.FromJust();
   }
   return ptr;
--- a/src/support.h
+++ b/src/support.h
@@ -61,8 +61,12 @@ inline static v8::MaybeLocal<T> ptr_to_maybe_local(T* ptr) {
 }
 
 template <class T>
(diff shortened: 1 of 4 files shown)


src/support.h
```diff
@@ -61,8 +61,12 @@ inline static v8::MaybeLocal<T> ptr_to_maybe_local(T* ptr) {
 }
 
 template <class T>
-inline static T maybe_to_value(v8::Maybe<T> maybe, T default_value) {
-  return maybe.FromMaybe(default_value);
+inline static T* maybe_to_ptr(v8::Maybe<T> maybe) {
+  T* ptr = new T();
```
piscisaureus: This looks like a memory leak. When is this thing deleted?

bartlomieju: Probably never... I added delete before assigning, is that correct? I need to sharpen my cpp 😅

bartlomieju: Okay after some pondering I think `Maybe<T>` should be rewritten. 

In V8 APIs `Maybe<T>` is coupled with `V8_WARN_UNUSED_RESULT`. That suggests that `Maybe<T>` must always be consumed (or have `#[must_use]` attribute in Rust). 

So I guess I should return pointer to `Maybe<T>` to Rust and then consume its methods in Rust idiomatic way.

piscisaureus: This is not going in the right direction. Let me take a stab at it.

PR #65, opened 2019-12-14
add v8::PromiseRejectMessage

bartlomieju:


src/promise.rs
```diff
@@ -176,3 +186,33 @@ impl PromiseResolver {
...
+  pub fn get_event(&self) -> PromiseRejectEvent {
+    unsafe { v8__PromiseRejectMessage__GetEvent(self) }
+  }
+
+  pub fn get_value(&self) -> Local<'_, Value> {
+    unsafe {
+      Local::from_raw(v8__PromiseRejectMessage__GetValue(self)).unwrap()
+    }
+  }
+}
```
bartlomieju: This one is tricky, it returns `Local`s, but is there's no way to get current isolate/context, whatever to get scope, because there's no such method on `PromiseRejectMessage`.

I took a look at [tests in v8](https://cs.chromium.org/chromium/src/v8/test/cctest/test-api.cc?sq=package:chromium&g=0&l=15570) and they have a helper to obtain current isolate.

bartlomieju:
#63 should be fixed before landing this one, marking as WIP

bartlomieju:
@ry PTAL

src/isolate.rs
```diff
@@ -16,6 +17,10 @@ extern "C" {
     caputre: bool,
     frame_limit: i32,
   );
+  fn v8__Isolate__SetPromiseRejectCallback(
+    isolate: &mut CxxIsolate,
+    callback: extern "C" fn(PromiseRejectMessage),
```
ry: Add a type alias for this

```rust
type PromiseRejectCallback = extern "C" fn(PromiseRejectMessage);
```

src/object.rs
```diff
@@ -59,6 +59,11 @@ impl Object {
       .unwrap()
     }
   }
+
+  /// Return the isolate to which the Object belongs to.
+  pub fn get_isolate(&self) -> &mut CxxIsolate {
+    unsafe { v8__Object__GetIsolate(self) }
```
ry: This is fine for now, but I think we need a way to return a reference to the public `Isolate` without exposing `CxxIsolate`. I think maybe something like this:
```rust
pub fn get_isolate(&mut self) -> std::cell::RefCell<Isolate>
```
But just leave it as such for now. We can fix this later.

PR #68, opened 2019-12-16
add v8::PropertyCallbackInfo

bartlomieju:
Stripped down version allowing only for `v8::PropertyCallbackInfo<v8::Value>` 

Depends on #67


src/binding.cc
```diff
@@ -366,24 +366,38 @@ v8::Promise* v8__Promise__Then2(v8::Promise* self,
...
+
+v8::Object* v8__PropertyCallbackInfo__This(
+    const v8::PropertyCallbackInfo<v8::Value>* self) {
+  return local_to_ptr(self->This());
+}
+
+v8::ReturnValue<v8::Value>* v8__PropertyCallbackInfo__GetReturnValue(
+    const v8::PropertyCallbackInfo<v8::Value>* self) {
+  v8::ReturnValue<v8::Value>* rv =
+      new v8::ReturnValue<v8::Value>(self->GetReturnValue());
```
piscisaureus: Actually does not look good after all (memory leak)

bartlomieju: Yikes, we already landed the same leak in #67

ry: What’s the leak? We need a LSAN check...

PR #80, opened 2019-12-19
Add Local::is_empty and Local::clear

ry:


src/local.rs
```diff
@@ -15,21 +15,36 @@ impl<'sc, T> Clone for Local<'sc, T> {
...
+    self.0.is_null()
+  }
+
+  /// Sets the handle to be empty. IsEmpty() will then return true.
+  pub fn clear(&mut self) {
+    self.0 = ptr::null_mut();
   }
 }
 
 impl<'sc, T> Deref for Local<'sc, T> {
```
ry: I suggest that since Locals can hold nullptr we do not implement Deref but instead implement something like this
https://doc.rust-lang.org/std/primitive.pointer.html#method.as_ref 
```rust
impl<'sc, T> Local<'sc, T> {
  unsafe fn as_ref<'a>(self) -> Option<&'a T> {
    if let Some(ref) = unsafe { self.0.as_ref() } {
      Some(ref)
    } else {
      None
    }
  }
}
```
What do you think Bert?

piscisaureus: But in many cases Locals can't hold nullptrs. Basically every time a Local is returned V8 will use a MaybeLocal if it is potentially empty.
Some V8 APIs do accept empty locals as parameters. But I think we can just model those as Option<Local>'s as well.

ry:
> I would rather avoid this. Can you just use Option<Local> where the local might be empty?

There are many places in the API where empty locals are used...

piscisaureus:
> There are many places in the API where empty locals are used...

But then what's the problem with passing `None` if you want to pass an empty local?

ry:
I understand where you're coming from `Option<Local>` is safer and more Rusty - but I worry that we're changing the semantics too much and will run into problems later. Imagine a situation where you get a callback from V8, say `v8::MessageCallback`, which gets two Local handles as arguments. It's valid for V8 to call that callback with empty handles. So we're forced to change essentially all APIs which previously take `Local` to `Option<Local>`. I think it will cause less confusion if `Local` more closely matched how it's used in V8. We just need to be more careful about how they are dereferenced.

piscisaureus:
> It's valid for V8 to call that callback with empty handles

Is it really? That seems odd, because they've introduced MaybeLocal for these cases.

But APIs that weren't updated really aren't a problem. iterally all you have to do is change the function signature from Local<T> to Option<Local<T>> and you're golden.

ry:
@piscisaureus You're saying that empty local handles are a legacy mistake of the V8 API and the intention is that they switch to the use of MaybeLocal everywhere a Local handle can be empty? If that's the case, then I agree with you we should use Option<Local>; with the caveat that we document this fact.

PR #83, opened 2019-12-19
Unsplit isolate

ry:


ry:
closed in favor of https://github.com/denoland/rusty_v8/pull/84

Isolate isn't being dropped in this patch

PR #84, opened 2019-12-19
Cxxisolate should be NonNull

ry:


ry:
Actually I prefer https://github.com/denoland/rusty_v8/pull/85 now

PR #85, opened 2019-12-19
Rename CxxIsolate to Isolate and Isolate to OwnedIsolate

ry:


ry:
There's still more to do here wrt LockedIsolate - but what do you think about this rename?

src/isolate.rs
```diff
@@ -99,26 +84,51 @@ impl Isolate {
...
   ) {
-    unsafe { v8__Isolate__SetPromiseRejectCallback(self.0, callback) }
+    unsafe { v8__Isolate__SetPromiseRejectCallback(self, callback) }
+  }
+
+  /// Disposes the isolate.  The isolate must not be entered by any
+  /// thread to be disposable.
+  pub unsafe fn dispose(&mut self) {
+    v8__Isolate__Dispose(self)
+  }
```
ry: Note this is unsafe. Most users will not call this directly.

src/locker.rs
```diff
@@ -18,12 +17,7 @@ extern "C" {
 }
 
 #[repr(C)]
-pub struct Locker<'a> {
-  has_lock: bool,
-  top_level: bool,
-  isolate: &'a mut Isolate,
-  phantom: PhantomData<&'a Isolate>,
-}
+pub struct Locker<'sc>([usize; 2], PhantomData<&'sc mut ()>);
```
ry: This struct layout is not verified anywhere and may cause problems. It's better to use an opaque buffer like this, at least we can assert it's the right size.

PR #88, opened 2019-12-20
Remove locked isolate

ry:
depends on #87


ry:
I think it's progress - we need to simplify as much as possible until we're complete with the bindings.

PR #94, opened 2019-12-20
Add note about libglib2.0-dev dependency to readme

piscisaureus:


hayd:
Would this be needed if #49? (`--target x86_64-unknown-linux-musl`)

PR #95, opened 2019-12-20
Add Local lifetimes back

piscisaureus:


src/number.rs
```diff
@@ -1,29 +1,33 @@
...
 /// A JavaScript number value (ECMA-262, 4.3.20)
 #[repr(C)]
 pub struct Number(Opaque);
 
 impl Number {
-  pub fn new(isolate: &Isolate, value: f64) -> Local<Number> {
+  pub fn new<'sc>(
+    scope: &mut HandleScope<'sc>,
+    value: f64,
+  ) -> Local<'sc, Number> {
```
ry: Doesn't match V8's API

piscisaureus: Because in v8 the handlescope is implicit to the active isolate. Here it's explicit because we want to know it's lifetime. What's the problem?

PR #98, opened 2019-12-21
add v8::Module::instantiate_module

ry:


src/module.rs
```diff
@@ -1,10 +1,22 @@
...
 use crate::support::Opaque;
 use crate::Context;
 use crate::Local;
+use crate::String;
 use crate::Value;
 
-extern "C" {}
+// Ideally the return value would be Option<Local<Module>>... but not FFI-safe
+type ResolveCallback =
+  extern "C" fn(Local<Context>, Local<String>, Local<Module>) -> *mut Module;
```
ry: @piscisaureus Have you thought through how this works with respect to lifetimes?

piscisaureus: Yes. This looks good.

ry:
I'm landing this even though it's not complete - there's no test - it's only half done. But I'm going to bed and I want to allow @bartlomieju to work on Module further - I'll probably be offline tomorrow.

PR #109, opened 2019-12-21
allow building with compatible system clang

chrmoritz:
This PR is a WIP to allow building with a compatible system `clang` instead of the one provided by Chromium.

To find a compatible `clang`/`llvm` installation it checks both the `$CLANG_BASE_PATH` env variable and looks for `clang` in the `$PATH`. If it finds a clang version there it compares the version output against known compatible version (we might need to update these constants over time when we upgrade V8) using regex's borrowed from node.js configure (added a build-dependencies on the regex crate).
Also it allows specifying additional arguments to gn by setting the `$GN_ARGS` similar to current `$DENO_BUILD_ARGS`.

Please keep in mind that I'm a beginner in rust, so let me know if I can improve the code somehow.

Also note, that this will switch over the current macOS Github Action build to use Apple clang 11 instead of Chromium's clang.

Currently I'm doing some compatibility testing to find out the compatible `clang`/`llvm` version with travis (because it has more images options than Github Actions):
* Apple Clang (only lates v11 supported, v10- fails): https://travis-ci.com/chrmoritz/rusty_v8/builds/142017554
* LLVM on macOS (via homebrew, works with v8.0+, v7.0- fails): https://travis-ci.com/chrmoritz/rusty_v8/builds/142017979
* LLVM 7 on linux (failed): https://travis-ci.com/chrmoritz/rusty_v8/builds/142018120
* linux build with official llvm 8.0 binaries: https://travis-ci.com/chrmoritz/rusty_v8/builds/142023043


build.rs
```diff
@@ -110,19 +129,57 @@ fn need_gn_ninja_download() -> bool {
     && env::var_os("GN").is_some())
 }
 
+fn is_compatible_clang_version(clang_path: &Path) -> bool {
```
ry: Cool!

chrmoritz: We'd just need to update the compatible versions constants roughly once per year during a V8 upgrade once Google decides to break compatibility with older `clang`'s again.

Another side node: It looks like a lot of linux distributions ship their versioned `clang` packages with a version suffix (e.g. `clang-8`), which is unfortunately not compatible with Chromium's build scripts (and I guess we don't want to patch them too heavily). So this approve is only compatibly with: Apple clang (on macOS), homebrew's llvm packages (on macOS and linux via Homebrew on Linux?), the official llvm binaries (from http://releases.llvm.org/, all platforms) and maybe the default `clang` package on Linux (but they are often too old anyway).

Cargo.toml
```diff
@@ -55,6 +55,7 @@ bitflags = "1.2.1"
 [build-dependencies]
 cargo_gn = "0.0.13"
 which = "3.1.0"
+regex = "1.3.1"
```
ry: Is it possible to accomplish this without adding the regex dependency? It seems like a large addition for a relatively simple task.

chrmoritz: I don't know. Parsing all these different `clang` version string seems to be quite a difficult task and the regex's from Node.js have been proven over time. You have to distinguish between Apple `clang` (which uses XCode version and had different version string formats over time) and llvm `clang` (where also distributions like to modify the format of the version string).

build.rs
ry: it might be worthwhile to add some of this in a comment

PR #110, opened 2019-12-21
add v8::EscapableHandleScope

bartlomieju:


src/handle_scope.rs
```diff
@@ -55,3 +70,61 @@ impl<'sc> AsMut<Isolate> for HandleScope<'sc> {
...
+  fn drop(&mut self) {
+    unsafe { v8__EscapableHandleScope__DESTRUCT(self) }
+  }
+}
+
+impl<'sc> AsRef<EscapableHandleScope<'sc>> for EscapableHandleScope<'sc> {
+  fn as_ref(&self) -> &Self {
+    self
+  }
+}
```
ry: what is this for?

src/handle_scope.rs
```diff
@@ -55,3 +70,61 @@ impl<'sc> AsMut<Isolate> for HandleScope<'sc> {
...
+      scope.assume_init()
+    }
+  }
+
+  /// Pushes the value into the previous scope and returns a handle to it.
+  /// Cannot be called twice.
+  pub fn escape<'parent>(
+    &mut self,
+    mut value: Local<'sc, Value>,
+  ) -> Local<'parent, Value> {
```
ry: I guess this is right. @piscisaureus ?

tests/test_api.rs
```diff
@@ -93,6 +93,63 @@ fn test_string() {
...
+  params.set_array_buffer_allocator(v8::Allocator::new_default_allocator());
+  let mut isolate = v8::Isolate::new(params);
+  let mut locker = v8::Locker::new(&isolate);
+  isolate.enter();
+  v8::HandleScope::enter(&mut locker, |scope1| {
+    // After dropping EscapableHandleScope, we should be able to 
+    // read escaped values.  
+
+    let mut escapable_scope = v8::EscapableHandleScope::new(scope1);
+    let mut scope: v8::HandleScope = unsafe { std::mem::transmute_copy(&escapable_scope) };
```
ry: :/

tests/test_api.rs
```diff
@@ -93,6 +93,63 @@ fn test_string() {
...
+    let l2: Local<v8::Integer> = cast(escapable_scope.escape(l2));
+    drop(escapable_scope);
+    assert_eq!(l2.value(), 456);
+
+    let mut escapable_scope = v8::EscapableHandleScope::new(scope1);
+    let mut scope: v8::HandleScope = unsafe { std::mem::transmute_copy(&escapable_scope) };
+    let l3: Local<v8::Value> = cast(v8::Number::new(&mut scope, 78.9));
+    let l3: Local<v8::Number> = cast(escapable_scope.escape(l3));
+    drop(escapable_scope);
+    assert_eq!(l3.value(), 78.9);
```
ry: There are 3 v8::Number::new tests - maybe there can be just one

tests/test_api.rs
bartlomieju: Worst thing is this is very bad... `HandleScope` has size 3 and `EscapabaleHandleScope` has size 4. I think methods should require some kind of `impl` or `dyn` for scope argument, I'll look into it

src/handle_scope.rs
bartlomieju: Just copy-pasted from `HandleScope` so they have same API

tests/test_api.rs
```diff
@@ -93,6 +93,63 @@ fn test_string() {
...
+  let mut locker = v8::Locker::new(&isolate);
+  isolate.enter();
+  v8::HandleScope::enter(&mut locker, |scope1| {
+    // After dropping EscapableHandleScope, we should be able to 
+    // read escaped values.  
+
+    let mut escapable_scope = v8::EscapableHandleScope::new(scope1);
+    let mut scope: v8::HandleScope = unsafe { std::mem::transmute_copy(&escapable_scope) };
+    let l1: Local<v8::Value> = cast(v8::Integer::new(&mut scope, -123));
+    let l1: Local<v8::Integer> = cast(escapable_scope.escape(l1));
```
ry: maybe don't write over the previous variable but use a different one

maybe put this all in a block to test scoping

tests/test_api.rs
bartlomieju: Done and done

tests/test_api.rs
bartlomieju: Done, now it looks super clean, I guess we could rewrite `HandleScope` to work the same way?

src/handle_scope.rs
piscisaureus: 'parent is unbound, so as far as the borrow checker is concerned the returned local can go anywhere. 
In reality it is only valid in the parent HandleScope.

PR #113, opened 2019-12-22
V8  escapable handle scope

bartlomieju:


bartlomieju:
@ry I think problems with lifetimes still hold. Do you want to go ahead and land this now or shall we fix them first?

ry:
@bartlomieju What are the lifetime problems?

bartlomieju:
@ry example by Bert that should fail, but it compiles fine:
```rust
  let local: Local<v8::String>;
  {
    let scope = &mut v8::HandleScope::new(&mut locker);
    local = v8::String::new(scope, "bla").unwrap();
  }
```
Note that it uses `HandleScope` but the same rules apply to `EscapableHandleScope`

PR #120, opened 2019-12-23
first pass implementation for v8::Module

ry:
going to just land these basic bindings before trying to address the ResolveCallback

### DIFF
--- a/src/module.rs
+++ b/src/module.rs
@@ -5,17 +5,48 @@ use crate::Context;
 use crate::Local;
 use crate::String;
 use crate::Value;
+use std::mem::MaybeUninit;
 
-// Ideally the return value would be Option<Local<Module>>... but not FFI-safe
 type ResolveCallback =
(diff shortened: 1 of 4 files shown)


src/module.rs
```diff
@@ -5,17 +5,48 @@ use crate::Context;
...
+    this: *mut Module,
+    context: *mut Context,
+  ) -> *mut Value;
+  fn v8__Location__GetLineNumber(this: &Location) -> int;
+  fn v8__Location__GetColumnNumber(this: &Location) -> int;
+}
+
+#[repr(C)]
+/// A location in JavaScript source.
+pub struct Location([usize; 2]);
```
bartlomieju: I think I screwed up `Location`. 

```
// fail
static_assert(sizeof(v8::Location) == sizeof(size_t) * 2,
              "Location size mismatch");
// success
static_assert(sizeof(v8::Location) == sizeof(size_t) * 1,
              "Location size mismatch");
```

PR #122, opened 2019-12-23
first pass at snapshots

bartlomieju:


src/snapshot.rs
```diff
@@ -45,7 +45,7 @@ impl SnapshotCreator {
   /// - `external_references` - a null-terminated array of external references
   ///        that must be equivalent to CreateParams::external_references.
   /// - `existing_blob` - existing snapshot from which to create this one.
-  pub fn new() -> Self {
+  pub fn default() -> Self {
```
ry: impl Default for SanpshotCreator ?

tests/test_api.rs
```diff
@@ -921,3 +921,44 @@ fn equality() {
...
+  let isolate = v8::Isolate::new(params);
+  let mut locker = v8::Locker::new(&isolate);
+  v8::HandleScope::enter(&mut locker, |scope| {
+    let mut context = v8::Context::new(scope);
+    context.enter();
+    let source = v8::String::new(scope, "a === 'Hello 13th planet'").unwrap();
+    let mut script = v8::Script::compile(scope, context, source, None).unwrap();
+    let result = script.run(scope, context).unwrap();
+    let true_val: Local<v8::Value> = cast(v8::new_true(scope));
+    assert!(result.same_value(true_val));
```
ry: 👍

src/binding.cc
```diff
@@ -610,6 +619,33 @@ void v8__PropertyCallbackInfo__GetReturnValue(
...
+
+v8::StartupData v8__SnapshotCreator__CreateBlob(
+    v8::SnapshotCreator* self,
+    v8::SnapshotCreator::FunctionCodeHandling function_code_handling) {
+  printf("working fine here\n");
+  auto a = self->CreateBlob(function_code_handling);
+  printf("working fine here too\n");
+  return a;
+}
+
```
bartlomieju: @ry it crashes CI in this binding on 
```
auto a = self->CreateBlob(function_code_handling);
```
It works for me in debug build though; I wonder if it might have something to do with `SerializeInternalFieldsCallback` 🤔

PR #134, opened 2019-12-26
add v8::SharedArrayBuffer

afinch7:
It's technically possible to drop the `SharedBuf` returned by `SharedArrayBuffer::new`(and end up with some strange results). Maybe I should rework this to use the internal allocation/backing instead.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -579,6 +579,10 @@ v8::PrimitiveArray* v8__ScriptOrModule__GetHostDefinedOptions(
   return local_to_ptr(self.GetHostDefinedOptions());
 }
 
+v8::SharedArrayBuffer* v8__SharedArrayBuffer__New(v8::Isolate* isolate, void *data, size_t byte_length) {
+  return local_to_ptr(v8::SharedArrayBuffer::New(isolate, data, byte_length, v8::ArrayBufferCreationMode::kExternalized));
+}
+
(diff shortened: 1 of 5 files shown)


ry:
@afinch7 we’ve recently added more support for BackingStore in ArrayBuffer. Can you make use of that?

afinch7:
@ry I have a working version of this using the internal backing, but it might make sense to add support for external backing via `BackingStore` in a future PR.

src/binding.cc
```diff
@@ -682,6 +682,26 @@ v8::PrimitiveArray* v8__ScriptOrModule__GetHostDefinedOptions(
...
+  return local_to_ptr(v8::SharedArrayBuffer::New(isolate, byte_length));
+}
+
+size_t v8__SharedArrayBuffer__ByteLength(v8::SharedArrayBuffer& self) {
+  return self.ByteLength();
+}
+
+void v8__SharedArrayBuffer__GetContents(v8::SharedArrayBuffer& self, v8::SharedArrayBuffer::Contents* out) {
+  *out = self.GetContents();
+}
```
ry: SharedArrayBuffer::GetContents is deprecated:
https://cs.chromium.org/chromium/src/v8/include/v8.h?l=5698&rcl=4c53f9a51444393133ff303952f1296603d44ab7

afinch7: Didn't see that will refactor to use `BackingStore`.

PR #135, opened 2019-12-26
Add shared_ptr wrapper and implement ArrayBuffer::get_backing_store() + new_with_backing_store()

piscisaureus:
TODO: add test that the C++ reference count actually increases/decreases as expected, and that memory is released when the reference count gets to zero.

### DIFF
--- a/src/array_buffer.rs
+++ b/src/array_buffer.rs
@@ -1,6 +1,10 @@
+use crate::support::long;
 use crate::support::Delete;
 use crate::support::Opaque;
+use crate::support::Shared;
+use crate::support::SharedRef;
 use crate::support::UniqueRef;
+use crate::InIsolate;
 use crate::Isolate;
 use crate::Local;
 use crate::ToLocal;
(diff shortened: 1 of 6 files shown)


ry:
Nice! Looking good.

Probably it would be good to bind this `ArrayBuffer` constructor too:
```c++
  static Local<ArrayBuffer> New(Isolate* isolate,
                                std::shared_ptr<BackingStore> backing_store);
```
and maybe also this one (tho not sure if it's useful)
```c++
  static std::unique_ptr<BackingStore> NewBackingStore(
      void* data, size_t byte_length, BackingStoreDeleterCallback deleter,
      void* deleter_data);
```

piscisaureus:
> TODO: add test that the C++ reference count actually increases/decreases as expected, 

Done.

> Probably it would be good to bind this `ArrayBuffer` constructor too:
> 
> ```c++
>   static Local<ArrayBuffer> New(Isolate* isolate,
>                                 std::shared_ptr<BackingStore> backing_store);
> ```

Added.
 
> and maybe also this one (tho not sure if it's useful)
> 
> ```c++
>   static std::unique_ptr<BackingStore> NewBackingStore(
>       void* data, size_t byte_length, BackingStoreDeleterCallback deleter,
>       void* deleter_data);
> ```

Given that it is not straightforward, I'm skipping that for now.

piscisaureus:
We probably need something like `impl<T> From<UniqueRef<T>> for SharedRef<T>` but that can be done in a separate patch too.

Issue #139, opened 2019-12-26
add hello world code example to main docs page


ry:

ry:
f2cc688439a8be50225ad13af9845b7cea000b21

PR #142, opened 2019-12-27
add hello world example

manyuanrong:
#139

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -3,35 +3,34 @@
 //! # Hello World Example
 //!
 //! ```
-//! use rusty_v8::{Isolate, V8, Allocator, HandleScope, Script, Context, Locker, Local};
-//! use rusty_v8::platform;
+//! use rusty_v8::{Isolate, V8, HandleScope, Script, Context, Locker, Local};
+//! use rusty_v8::{platform, new_default_allocator};
 //!
 //! let platform = platform::new_default_platform();
 //! V8::initialize_platform(platform);
 //! V8::initialize();
 //!
 //! let mut create_params = Isolate::create_params();
-//! create_params.set_array_buffer_allocator(Allocator::new_default_allocator());
+//! create_params.set_array_buffer_allocator(new_default_allocator());
 //! let isolate = Isolate::new(create_params);
 //! let mut locker = Locker::new(&isolate);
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1,5 +1,40 @@
...
+//! let platform = platform::new_default_platform();
+//! V8::initialize_platform(platform);
+//! V8::initialize();
+//!
+//! let mut create_params = Isolate::create_params();
+//! create_params.set_array_buffer_allocator(Allocator::new_default_allocator());
+//! let isolate = Isolate::new(create_params);
+//! let mut locker = Locker::new(&isolate);
+//!
+//! HandleScope::enter(&mut locker, |scope| {
```
bartlomieju:
```
let mut hs = v8::HandleScope::new(&mut locker);
let scope = hs.enter();
```

manyuanrong: > ```
> let mut hs = v8::HandleScope::new(&mut locker);
> let scope = hs.enter();
> ```

 But I can't find `new` in`HandleScope`

manyuanrong:
I created a project of Hello world to ensure that the sample code can run normally
https://github.com/manyuanrong/rusty_v8_hello_world

src/lib.rs
bartlomieju: https://github.com/denoland/rusty_v8/blob/29fa5388f547e16596f71bfb2f0fd8e6b7021c2d/src/handle_scope.rs#L35-L54

manyuanrong: OK, I saw it in this project, but my example uses the version released on "create.io", which seems to be older.

src/lib.rs
```diff
@@ -1,5 +1,39 @@
...
+//! code.to_rust_string_lossy(scope);
+//! let mut script = Script::compile(scope, context, code, None).unwrap();
+//! let result = script.run(scope, context).unwrap();
+//! let result: Local<rusty_v8::String> = unsafe { std::mem::transmute_copy(&result) };
+//!
+//! let str = result.to_rust_string_lossy(scope);
+//!
+//! println!("{}", str);
+//!
+//! context.exit();
```
bartlomieju: Can you wrap this bit in curly braces (to showcase how long `HandleScope` lives). Also please add `drop(locker)` at the end.

ry:
@manyuanrong thanks!

PR #145, opened 2019-12-28
[WIP] try safer cast

bartlomieju:


src/string.rs
```diff
@@ -151,3 +151,21 @@ impl Deref for String {
...
+impl Into<String> for Value {
+  fn into(self) -> String {
+    // TODO: do this check in debug mode only
+    if !self.is_string() {
+      panic!("unable to convert")
+    } 
+    
+    unsafe { std::mem::transmute(self) }
+  }
+}
```
bartlomieju:
```c++
void v8::Function::CheckCast(Value* that) {
  i::Handle<i::Object> obj = Utils::OpenHandle(that);
  Utils::ApiCheck(obj->IsCallable(), "v8::Function::Cast",
                  "Could not convert to function");
}

void v8::Boolean::CheckCast(v8::Value* that) {
  i::Handle<i::Object> obj = Utils::OpenHandle(that);
  Utils::ApiCheck(obj->IsBoolean(), "v8::Boolean::Cast",
                  "Could not convert to boolean");
}

void v8::Name::CheckCast(v8::Value* that) {
  i::Handle<i::Object> obj = Utils::OpenHandle(that);
  Utils::ApiCheck(obj->IsName(), "v8::Name::Cast", "Could not convert to name");
}
```

Lot of `CheckCast` methods use types from `v8::internal` namespace which I'm pretty sure we don't want to port...

src/string.rs
```diff
@@ -151,3 +151,21 @@ impl Deref for String {
...
+
+impl<'sc> Into<Local<'sc, String>> for Local<'sc, Value>
+{
+  fn into(self) -> Local<'sc, String> {
+    unsafe { std::mem::transmute(self) }
+  }
+}
```
bartlomieju: This is interesting; in `local.rs` we have:
```rust
impl<'sc, T> From<Local<'sc, T>> for Local<'sc, Value>
where
  T: Deref<Target = Value>,
{
  fn from(v: Local<'sc, T>) -> Local<'sc, Value> {
    unsafe { std::mem::transmute(v) }
  }
}
```
If type implements `From` then blanket implementation for `Into` is provided, but in the case above it's not working without explicit implementation.

manyuanrong: I also tried something similar, the compiler prompts conflict

PR #146, opened 2019-12-29
automated conversion

manyuanrong:


src/local.rs
```diff
@@ -86,6 +94,29 @@ where
...
+into_subtype!(
+  String,
+  Boolean,
+  Integer,
+  Number,
+  Object,
+  ArrayBufferView,
+  Function,
+  Uint8Array
+);
```
ry: Interesting. But what about casting from, say, Local<Boolean> into Local<Primitive>? Or Local<String> into Local<Name>? This solution doesn't seem to cover those.

bartlomieju: I think we're focusing on bad part. We should figure out conversion between concrete types, `Local` is just a wrapper to underlying data type. The other problem is Rust's blanket implementation for `Into` trait if you implement `From` trait, eg. `Promise` can be cast to `Value` (because `Value` is superclass of `Promise`) but `Value` **might** be cast to `Promise`, but this cast might fail (because not every `Value` is a `Promise`)

manyuanrong: @ry I don't quite understand what you mean *(It seems that github has eaten some of your characters?)* . This pr implements the conversion from `Local<Value>` to `Local<OtherType>`.
```rust
let some_value: Local<Value> = xxx;
let string:Local<String> = some_value.into();
let number:Local<Number> = some_value.into();
...
```
The conversion from `Local<OtherType>` to `Local <Value>` is
```rust
impl<'sc, T> From<Local<'sc, T>> for Local<'sc, Value>
where
  T: Deref<Target = Value>,
{
  fn from(v: Local<'sc, T>) -> Local<'sc, Value> {
    unsafe { std::mem::transmute(v) }
  }
}
```
As for @bartlomieju 's point of view, I may agree with it. In C ++, we may pay more attention to the conversion of types. Similarly, in Rust, we can also determine the type by `Value.isXXX` before conversion, but the current check in `Value` The method is only partially implemented and is missing a lot. I am going to add other missing methods in another pr.

bartlomieju: > As for @bartlomieju 's point of view, I may agree with it. In C ++, we may pay more attention to the conversion of types. Similarly, in Rust, we can also determine the type by Value.isXXX before conversion, but the current check in Value The method is only partially implemented and is missing a lot. I am going to add other missing methods in another pr.

This is not 100% sure, a lot of checks during casts are using methods and types from `v8::internal` namespace.

ry: Oops - indeed my comment didn't come across properly - but you answered my question. I misunderstood the direction of the cast happening.

manyuanrong:
Better in #148

PR #154, opened 2019-12-30
add Send impl

bartlomieju:
Needed for the server example, I'm not sure this is correct, please review


src/array_buffer.rs
```diff
@@ -200,3 +202,5 @@ impl ArrayBuffer {
     }
   }
 }
+
+unsafe impl Send for ArrayBuffer {}
```
ry: hmm is this really needed?

bartlomieju: Oooops, I thought I pushed.

PR #163, opened 2020-01-01
add all value checkers

manyuanrong:


src/binding.cc
```diff
@@ -279,24 +279,178 @@ v8::Module* v8__ScriptCompiler__CompileModule(
 }
 
 bool v8__Value__IsUndefined(const v8::Value& self) {
-  return self.IsUndefined();
+    return self.IsUndefined();
```
ry: Thanks - can you run clang-format on this file? Indent should be 2 spaces.

PR #168, opened 2020-01-02
Create CallbackScope from Promise and PromiseRejectMessage

piscisaureus:


ry:
blocking @bartlomieju in porting deno, so landing now. needs test.

PR #181, opened 2020-01-04
Experiment: use `impl Into<Local<'a, T>>`

piscisaureus:


bartlomieju:
This is very ergonomic :+1:

piscisaureus:
> This is very ergonomic 👍

Thanks :D I like it too but there are some rough edges. I think I'll finish it when the more important things are done.

tests/test_api.rs
```diff
@@ -1692,9 +1678,9 @@ fn dynamic_import() {
     let mut cbs = v8::CallbackScope::new(context);
     let mut hs = v8::HandleScope::new(cbs.enter());
     let scope = hs.enter();
-    assert!(specifier.strict_equals(v8_str(scope, "bar.js").into()));
+    assert!(specifier.strict_equals(v8_str(scope, "bar.js")));
     let e = v8_str(scope, "boom");
-    scope.isolate().throw_exception(e.into());
+    scope.isolate().throw_exception(e);
```
ry: very nice!

piscisaureus:
Closing for now.

I still want to do this some day, but we'll add it piecemeal, and probably after defining a specific trait for it (IntoLocal) rather than using the generic Into trait.

PR #183, opened 2020-01-04
Refactor scopes and callbacks so users don't need to transmute

piscisaureus:


tests/test_api.rs
```diff
@@ -1036,58 +1035,39 @@ fn promise_rejected() {
...
-
-    let s = v8::String::new(scope, "Hello callback!").unwrap();
-    let value: v8::Local<v8::Value> = s.into();
-    let rv_value = rv.get(scope);
-    assert!(rv_value.is_undefined());
-    rv.set(value);
-    context.exit();
-  }
+fn fn_callback2(
+  scope: v8::FunctionScope,
```
piscisaureus: @bartlomieju, @ry 

You guys ok if I break up `FunctionCallbackInfo` into three pieces as demonstrated here. That makes it a lot easier to play nice with the rust borrow checker.

ry: Does this mean overhead for functions that wouldn’t otherwise use the return value ?

ry: Does this mean overhead for functions that wouldn’t otherwise use the return value ? EG Deno.core.print

ry: Also worth noting that this brings the API further away from the C++ API

piscisaureus: > Does this mean overhead for functions that wouldn’t otherwise use the return value ? EG Deno.core.print

No this is a zero cost abstraction.

piscisaureus: > Also worth noting that this brings the API further away from the C++ API

I am well aware. I try to avoid it as much as possible.

ry: GetReturnValue isn’t called on every call?

piscisaureus: > GetReturnValue isn’t called on every call?

It does. However GetReturnValue() does basically nothing, it just returns a pointer from the FunctionCallbackInfo struct. The status quo where we have to create an EscapableHandleScope in all these function callbacks probably wastes more cycles (although, in the grand scheme of things, just as insignificant).

The ReturnValue argument can also be made optional if so desired.

ry: Ok so it’s not really zero cost - we’re forcing the user to make more V8 API calls than necessary. But maybe the small overhead is worth it.

The EscapableHandleScope was also overhead induced by the type system. So it’s good you’re reducing the overhead.

src/callback_scope.rs
```diff
@@ -1,70 +1,76 @@
...
+/// A CallbackScope can be used to obtain a mutable Isolate reference within
+/// a callback that is called by V8 on the thread that already has a Locker
+/// on the stack.
+///
+/// Using a CallbackScope in any other situation is unsafe.
+/// Also note that CallbackScope should not be used for function and property
+/// accessor callbacks; use FunctionCallbackScope and PropertyCallbackScope
+/// instead.
+///
+/// A CallbackScope can be created from the following inputs:
```
piscisaureus: This comment is actually not so relevant any more.
I entend to pass a suitable Scope object to all callbacks that need one. The user won't have to create one manually.

src/object.rs
```diff
@@ -6,12 +6,12 @@ use crate::Context;
...
 use crate::Name;
 use crate::Object;
-use crate::PropertyCallbackInfo;
+use crate::PropertyCallbackScope;
 use crate::ToLocal;
 use crate::Value;
 
 pub type AccessorNameGetterCallback =
-  extern "C" fn(Local<Name>, &PropertyCallbackInfo);
+  extern "C" fn(Local<Name>, PropertyCallbackScope);
```
piscisaureus: I should probably undo this experimental change before landing.
This is not the final form it'll take and also not heading in that direction.

tests/test_api.rs
```diff
@@ -1036,58 +1037,39 @@ fn promise_rejected() {
...
-      let value: v8::Local<v8::Value> = s.into();
-      let rv_value = rv.get(scope);
-      assert!(rv_value.is_undefined());
-      rv.set(value);
-    }
-  }
+fn fn_callback(
+  scope: v8::FunctionCallbackScope,
+  args: v8::FunctionCallbackArguments,
+  mut rv: v8::ReturnValue,
```
piscisaureus: I'm torn on whether this requiring 'mut' is a good thing or not.

src/scope_traits.rs
```diff
@@ -0,0 +1,168 @@
...
+
+/// When scope implements this trait, this means that it Local handles can be
+/// created inside it.
+pub trait ToLocal<'s>: InIsolate {
+  unsafe fn to_local<T>(&mut self, ptr: *mut T) -> Option<Local<'s, T>> {
+    crate::Local::<'s, T>::from_raw(ptr)
+  }
+}
+
+impl<'s> ToLocal<'s> for Entered<'s, FunctionCallbackInfo> {}
```
piscisaureus: I intend to do away with the 'Entered' struct.

src/function.rs
```diff
@@ -91,52 +90,71 @@ impl<'cb> ReturnValue<'cb> {
...
-pub struct FunctionCallbackInfo(Opaque);
-
-impl InIsolate for FunctionCallbackInfo {
-  fn isolate(&mut self) -> &mut Isolate {
-    self.get_isolate()
-  }
+pub struct FunctionCallbackInfo {
+  implicit_args: *mut Opaque,
+  values: *mut Value,
+  length: int,
```
ry: This specifically matches the C++ class definition? What if say, the order of these changes? Seems fragile. This ideally would be solved by code generation, but could also be solved by using an opaque and adding bindings to retrieve the  members. Both solutions sound annoying - so I would suggest to just add a comment warning about this.

tests/test_api.rs
```diff
@@ -9,6 +9,7 @@ use std::sync::Mutex;
 
 use rusty_v8 as v8;
 use v8::InIsolate;
+use v8::MapFnTo;
```
piscisaureus: It's not ideal that the user has to import this "internal" trait. I try to avoid it as much as possible, and I believe I can eventually make it go away.

src/function.rs
```diff
@@ -91,52 +90,71 @@ impl<'cb> ReturnValue<'cb> {
...
-  pub fn get_isolate(&mut self) -> &mut Isolate {
-    unsafe { v8__FunctionCallbackInfo__GetIsolate(self) }
-  }
-
-  /// The number of available arguments.
   pub fn length(&self) -> int {
-    unsafe { v8__FunctionCallbackInfo__Length(self) }
+    unsafe {
+      let length = (*self.info).length;
+      debug_assert_eq!(length, v8__FunctionCallbackInfo__Length(self.info));
```
ry: This check probably will avoid the problem mentioned above. I would still suggest a warning comment on FunctionCallbackInfo's declaration.

src/function.rs
piscisaureus: I considered that too, and I'll do as you suggest and add a warning.
That said, the layout of this struct is effectively part of the v8 ABI as it is defined in v8.h and there are plenty of inline functions that (also defined in v8.h) that access its members. That makes it more-or-less fixed.
I also believe that the tests would catch a change in layout immediately.

src/scope_traits.rs
```diff
@@ -0,0 +1,168 @@
...
+
+  impl<'s> GetRawIsolate for Local<'s, Context> {
+    fn get_raw_isolate(&self) -> *mut Isolate {
+      (&**self).get_raw_isolate()
+    }
+  }
+
+  impl<'s> GetRawIsolate for Local<'s, Message> {
+    fn get_raw_isolate(&self) -> *mut Isolate {
+      (&**self).get_raw_isolate()
```
ry: I don't understand why there are two stars here...

piscisaureus: The first * calls `<Local<T> as Deref>::deref` to `&Message`.
The second * calls `<Message as Deref>::deref` to get to its base class (`Object`).

tests/test_api.rs
```diff
@@ -1036,58 +1037,39 @@ fn promise_rejected() {
...
+  let arg2_val = v8::Integer::new(scope, 2);
+  let arg2 = args.get(1);
+  assert!(arg2.is_number());
+  assert!(arg2.strict_equals(arg2_val.into()));
+
+  let s = v8::String::new(scope, "Hello callback!").unwrap();
+
+  let rv_value = rv.get(scope);
+  assert!(rv_value.is_undefined());
+  rv.set(s.into());
```
ry: 👍

tests/test_api.rs
ry: Indeed. Maybe add a TODO comment here reflecting that, so we have a record of the intention.

PR #184, opened 2020-01-04
add v8::Boolean::new()

bartlomieju:


src/primitives.rs
```diff
@@ -35,3 +35,14 @@ pub fn new_false<'sc>(scope: &mut impl ToLocal<'sc>) -> Local<'sc, Boolean> {
   let ptr = unsafe { v8__False(scope.isolate()) };
   unsafe { scope.to_local(ptr) }.unwrap()
 }
+
+pub fn new_boolean<'sc>(
```
piscisaureus: Not a big deal of course but can we match the V8 unless there's a good reason not to.
v8 has this function:

```
Local<Boolean> Boolean::New(Isolate* isolate, bool value) {
  return value ? True(isolate) : False(isolate);
}
```

bartlomieju: Sure, why not

piscisaureus: No I mean actually exactly the opposite. I don't care how it's implemented.
I meant that the rust api should be v8::Boolean::new.

Or is there a reason to do it differently?

tests/test_api.rs
```diff
@@ -737,6 +737,12 @@ fn test_primitives() {
     assert!(!false_.is_undefined());
     assert!(!false_.is_null());
     assert!(!false_.is_null_or_undefined());
+
+    let true_ = v8::new_boolean(scope, true);
+    assert!(true_.is_true());
+
+    let false_ = v8::new_boolean(scope, false);
+    assert!(false_.is_false());
```
ry: debug_assert ?

bartlomieju:
Err, so so `v8::new_boolean()` or `v8::Boolean::new()`?

ry:
v8::Boolean::new please

ry:
I'd be happy to remove v8::new_true and v8::new_false now that we have v8::Boolean::new

bartlomieju:
> I'd be happy to remove v8::new_true and v8::new_false now that we have v8::Boolean::new

Ok

PR #186, opened 2020-01-05
v0.0.22

ry:
allow v8/tools to be included in package


ry:
Unfortunately removing that exclude line from Cargo.toml makes the rusty_v8 crate 11mb
```
> ls -lh target/package/rusty_v8-0.0.22.crate
-rw-r--r--  1 rld  staff    11M Jan  4 20:01 target/package/rusty_v8-0.0.22.crate
```

ry:
ok got it down to 9.3 
```
> ls -lh target/package/rusty_v8-0.0.22.crate
-rw-r--r--  1 rld  staff   9.3M Jan  4 20:53 target/package/rusty_v8-0.0.22.crate
```

PR #195, opened 2020-01-06
Remove unnecessary derive(Debug)

ry:


piscisaureus:
I don't want to remove the c++ base class support. We will need it soon.

ry:
sorry - undone

piscisaureus:
OK diff looks better now.

PR #198, opened 2020-01-08
v0.0.24

piscisaureus:


ry:
Can you make this 0.1.0 instead?

piscisaureus:
> Can you make this 0.1.0 instead?

No not yet. After we get rid of the FunctionCallbackInfo transmute I'm +1 on that.

PR #202, opened 2020-01-10
ArrayBuffer::new_backing_store_from_boxed_slice

bartlomieju:


src/array_buffer.rs
```diff
@@ -175,6 +192,17 @@ impl ArrayBuffer {
     unsafe { scope.to_local(ptr) }.unwrap()
   }
 
+  pub fn new_with_unique_backing_store<'sc>(
+    scope: &mut impl ToLocal<'sc>,
+    backing_store: &mut UniqueRef<BackingStore>,
```
piscisaureus: This should not be an &mut UniqueRef. 
When you call this function that means that you're going to share the BackingStore with v8 so the user can't hold on to a unique reference.
Instead they can call the `.get_backing_store()` method on the returned `Local<ArrayBuffer>` and get a `SharedRef<BackingStore>` from it.

bartlomieju: I failed to obtain `SharedRef` from it... I'll give it another go tomorrow

tests/test_api.rs
```diff
@@ -259,6 +269,19 @@ fn array_buffer() {
...
+    let capacity = 100;
+    let mut data: Box<Vec<u8>> = Box::new(Vec::with_capacity(capacity));
+    let data_ptr = Box::into_raw(data);
+    let mut bs = v8::ArrayBuffer::new_backing_store_from_raw(
+      data_ptr as *mut std::ffi::c_void,
+      capacity,
+      deleter_callback,
+    );
+    assert_eq!(100, bs.byte_length());
+    assert_eq!(false, bs.is_shared());
```
bartlomieju: @piscisaureus I tried with just `Vec::with_capacity(capacity)` but then in deleter when I tried to `Vec::from_raw_parts(...)` I got segfault:
```
test_api-d423bd81ca175a6c(15580,0x70000d7f8000) malloc: *** error for object 0x7fe13da20fd0: pointer being freed was not allocated
test_api-d423bd81ca175a6c(15580,0x70000d7f8000) malloc: *** set a breakpoint in malloc_error_break to debug
```
Using `Box` works though

piscisaureus: A boxed vec is far from ideal though. It's a heap allocation to hold a pointer to another heap allocation. Maybe Vec::into_boxed_slice could help here.
Let's discuss Monday as I'm not working today.

bartlomieju:
@ry removed `SharedArrayBuffer::new_DEPRECATED`

PR #228, opened 2020-01-20
Add Object::define_own_property() 

bnoordhuis:
Also `Template::set_with_attr()` and the NONE, READ_ONLY, DONT_ENUM and DONT_DELETE
property attributes.


src/object.rs
```diff
@@ -120,6 +128,24 @@ impl Object {
     unsafe { v8__Object__CreateDataProperty(self, &*context, &*key, &*value) }
   }
 
+  /// Implements DefineOwnProperty.
+  ///
+  /// In general, CreateDataProperty will be faster, however, does not allow
```
bnoordhuis: Should I s/CreateDataProperty/create_data_property/ here?

ry: We have just been copying docstrings directly, so I wouldn't worry about it. 

In the future we hope to automate this somehow - maybe with some sort of smart switch/replace? But I think just having some docs is useful.

src/attr.rs
```diff
@@ -0,0 +1,94 @@
+#[repr(C)]
```
ry: Maybe rename this to src/property_attribute ?

bnoordhuis: Done!

src/object.rs
```diff
@@ -44,6 +45,13 @@ extern "C" {
     key: *const Name,
     value: *const Value,
   ) -> MaybeBool;
+  fn v8__Object__DefineOwnProperty(
+    object: &Object,
+    context: *const Context,
```
piscisaureus: I would swear I commented here, but my comment seems to be nowhere...

It's not necessary to unwrap Local<T> to a raw pointer before passing them as a function argument from Rust to C++.
If possible, try to be consistent between the `extern "C"` declarations on the C++ vs Rust side, which is lacking here:

`fn v8__Object__DefineOwnProperty(object: &Object, context: *const Context, ...) -> MaybeBool;`
`MaybeBool v8__Object__DefineOwnProperty(v8::Object& self, v8::Local<v8::Context> context, ...);`

I know it's probably copypasta but let's start somewhere. 
(cc @ry, @bartlomieju).

bnoordhuis: Yep, I copied it from create_data_property(). I'll see if I have time today to do a (general) clean up.

PR #236, opened 2020-01-21
Implement fmt::Display for StringView

### LINKED ISSUE #233: inspector: human-readable StringBuffer, StringView
For debugging the wire protocol. Just adding `#[derive(Debug)]` produces:

    UniquePtr(Some(StringBuffer { _cxx_vtable: CxxVTable(0x55d8a19e34b0) }))

Not super useful, I daresay.

bnoordhuis:
Make StringView objects inspectable to make it easier to debug the
V8 inspector's wire protocol.

StringBuffer objects can't be inspected yet because they need a mutable
reference if they want to call `StringBuffer::string()`, which makes
sense because that method calls out to a C++ virtual method that can
modify the object.

For now, instances can be inspected like this:

    let mut sb = StringBuffer::create(...);
    println!("StringBuffer({})", sb.string().unwrap());

Closes #233

### DIFF
--- a/src/inspector/string_view.rs
+++ b/src/inspector/string_view.rs
@@ -40,7 +40,6 @@ use std::string;
 //    TODO: find/open upstream issue to allow #[repr(bool)] support.
 
 #[repr(u8)]
-#[derive(Debug)]
 pub enum StringView<'a> {
   // Do not reorder!
   U16(CharacterArray<'a, u16>),
@@ -196,7 +195,7 @@ impl<'a, T> Deref for CharacterArray<'a, T> {
   }
 }
 
-#[derive(Copy, Clone, Debug)]
+#[derive(Copy, Clone)]
 pub struct StringViewIterator<'a: 'b, 'b> {
   view: &'a StringView<'b>,
   pos: usize,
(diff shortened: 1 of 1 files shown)


ry:
Looks good. How about removing the derive Debug too?

bnoordhuis:
I pushed a fix-up that removes it but let me know if you think it's a good idea to implement a custom `fmt::Debug`.

src/inspector/string_view.rs
```diff
@@ -146,6 +156,18 @@ where
 unsafe impl<'a, T> Send for CharacterArray<'a, T> where T: Copy {}
 unsafe impl<'a, T> Sync for CharacterArray<'a, T> where T: Sync {}
 
+impl fmt::Display for CharacterArray<'_, u8> {
+  fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+    f.write_str(&string::String::from_utf8_lossy(&*self))
```
piscisaureus: For debugging it's not such a big deal, but keep in mind that 8-bit V8 strings are Latin-8 and not UTF-8.
I would suggest:
```
write!(f, "{}", self.iter().map(|c| *c as char).collect::<String>()));
```

PR #238, opened 2020-01-22
Add String::new_empty()

bnoordhuis:


src/string.rs
```diff
@@ -61,11 +63,21 @@ bitflags! {
 }
 
 impl String {
+  pub fn new_empty<'sc>(scope: &mut impl ToLocal<'sc>) -> Local<'sc, String> {
+    let ptr = unsafe { v8__String__Empty(scope.isolate()) };
+    // FIXME(bnoordhuis) v8__String__Empty() is infallible so there
+    // is no need to box up the result, only to unwrap it again.
```
bnoordhuis: I initially wanted to fix this but Chesterton's fence: I didn't understand why Local is a NonNull instead of a raw pointer.

piscisaureus:
LGTM

src/string.rs
piscisaureus: > I didn't understand why Local is a NonNull instead of a raw pointer.

So the in-memory representation of `Option<Local<T>>` is the same as `*mut T`. It's not strictly necessary but it should avoid some wrapping/unwrapping overhead.

piscisaureus: > // FIXME(bnoordhuis) v8__String__Empty() is infallible so there
>     // is no need to box up the result, only to unwrap it again.

Eliminating null checks on values returned by V8 is really the least of our concern.
In fact, I think we generally should keep them, just to validate the assumption/assertions V8 makes about whether a returned Local may be empty or not.

Only if we find that we use some API a lot and the null checks do introduce significant overhead I would consider eliminating them. But I suspect that would happen with `v8::undefined()` much sooner than with `String::empty()`.

Issue #253, opened 2020-01-24
test: ScriptOrigin makes context_from_object_template crash 


bnoordhuis:
At commit 683aa2b2a981c7991932b793a8203ec245649d7a.

```diff
diff --git a/tests/test_api.rs b/tests/test_api.rs
index aabbac7..05adf01 100644
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -379,7 +379,9 @@ fn eval<'sc>(
   let mut hs = v8::EscapableHandleScope::new(scope);
   let scope = hs.enter();
   let source = v8_str(scope, code);
-  let mut script = v8::Script::compile(scope, context, source, None).unwrap();
+  let origin = mock_script_origin(scope, "test.js");
+  let mut script =
+    v8::Script::compile(scope, context, source, Some(&origin)).unwrap();
   let r = script.run(scope, context);
   r.map(|v| scope.escape(v))
 }
```

```
$ cargo test context_from_object_template
    Finished test [unoptimized + debuginfo] target(s) in 0.03s
     Running target/debug/deps/rusty_v8-98486895b0660b61

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 6 filtered out

     Running target/debug/deps/test_api-e710bec475054917

running 1 test


#
# Fatal error in ../../../v8/src/execution/isolate.h, line 546
# Debug check failed: (isolate) != nullptr.
#
#
#
#FailureMessage Object: 0x7f9635787b50
==== C stack trace ===============================

    /home/bnoordhuis/src/rusty_v8/target/debug/deps/test_api-e710bec475054917(+0x13642a3) [0x55a7c3bef2a3]
    /home/bnoordhuis/src/rusty_v8/target/debug/deps/test_api-e710bec475054917(+0xfa486d) [0x55a7c382f86d]
    /home/bnoordhuis/src/rusty_v8/target/debug/deps/test_api-e710bec475054917(+0x839473) [0x55a7c30c4473]
    /home/bnoordhuis/src/rusty_v8/target/debug/deps/test_api-e710bec475054917(+0x839175) [0x55a7c30c4175]
    /home/bnoordhuis/src/rusty_v8/target/debug/deps/test_api-e710bec475054917(+0x198fc4) [0x55a7c2a23fc4]
    /home/bnoordhuis/src/rusty_v8/target/debug/deps/test_api-e710bec475054917(+0x1a7422) [0x55a7c2a32422]
[...]

PR #254, opened 2020-01-25
Add IsolateScope

bnoordhuis:
Stack-allocated class which sets the isolate for all operations
executed within a local scope.

I didn't convert everything yet because I'd like feedback/consensus first on whether this is the right approach.

Refs: https://github.com/denoland/rusty_v8/issues/253#issuecomment-578399124

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -154,6 +154,7 @@ pub use property_attribute::*;
 pub use scope::CallbackScope;
 pub use scope::ContextScope;
 pub use scope::FunctionCallbackScope;
+pub use scope::IsolateScope;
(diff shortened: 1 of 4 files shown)


piscisaureus:
The code looks great but I don't think we need this. We can get safe access to a &mut Isolate through a `Locker` already. 

The type 'above' Locker doesn't need to be so carefully guarded; we actually need more something like an ArcIsolate which is reference counted and thread-safe and provides access to the few thread-safe methods in the V8 API.

bnoordhuis:
Closing per discussion in #253 (tl;dr V8 bug.)

PR #256, opened 2020-01-26
Reset slot when dropping v8::Global

bnoordhuis:
Harmonize the behavior of the Drop trait for v8::Global with the
behavior of the C++ destructor ~Global(). Before this commit it
would panic, now it clears the slot.

(Please tell me if there's a reason it currently works this way.)

### DIFF
--- a/src/global.rs
+++ b/src/global.rs
@@ -92,11 +92,7 @@ impl<T> Global<T> {
     let other_value = other.read(isolate);
     match (&mut self.value, &other_value) {
       (None, None) => {}
-      (target, None) => unsafe {
-        v8__Global__Reset__0(
-          &mut *(target as *mut Option<NonNull<T>> as *mut *mut Value),
-        )
-      },
+      (_, None) => self.reset0(),
       (target, source) => unsafe {
         v8__Global__Reset__2(
           &mut *(target as *mut Option<NonNull<T>> as *mut *mut Value),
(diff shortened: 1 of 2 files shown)


piscisaureus:
I'm 100% in favor of supporting this.
The reason it currently works the way it does is that resetting the Global panics if the Isolate is dropped before the Global, which does happen (in particular when the program exits).

piscisaureus:
So my long-term plan is more or less as follows:

* Change what is currently `OwnedIsolate` into a heap-allocated struct that is Send+Sync and contains a reference counter.
* Make v8::Isolate::new return an `OwnedIsolate`.
* Store a weak reference to the `OwnedIsolate` in one of the `v8::Isolate`'s user data slots.
* a non-empty `v8::Global` increases the `OwnedIsolate` reference count.

That way a Global can never be dropped after the Isolate is dropped, which would make the desirable reset-in-destructor behavior possible and safe.

bnoordhuis:
Right, that makes sense. I suppose that tying them to the lifetime of the isolate is impossible / un-ergonomic?

I'll take a look at #257 but that probably won't be until tonight or tomorrow; I have a private cello performance to attend. :-)

piscisaureus:
> Right, that makes sense. I suppose that tying them to the lifetime of the isolate is impossible / un-ergonomic?

Well, you'd end up with the semantic equivalent of a `mut Option<Local<'a, T>>`, so, although possible, it's not very useful.

> I'll take a look at #257 but that probably won't be until tonight or tomorrow; I have a private cello performance to attend. :-)

No worries ... I'll be travelling tomorrow and therefore AFK.

bnoordhuis:
I'll close this.

PR #257, opened 2020-01-27
WIP: `OwnedIsolate` -> reference-counting `IsolateHandle`

piscisaureus:


src/isolate.rs
```diff
@@ -344,38 +349,107 @@ impl Isolate {
...
 
-impl Deref for OwnedIsolate {
-  type Target = Isolate;
-  fn deref(&self) -> &Self::Target {
-    unsafe { self.0.as_ref() }
+impl Clone for IsolateHandle {
+  fn clone(&self) -> Self {
+    let prev_ref_count = unsafe { IsolateAnnex::get(&self.0) }
+      .get_ref_count()
+      .fetch_add(1, Ordering::Release);
```
bnoordhuis: This can be `Ordering::Relaxed` and the fetch_sub() on line 446 `Ordering::AcqRel` (IOW, this one's too strong and the other one too weak.)

For comparison, [that's how std::shared_ptr is implemented in libc++](https://bugs.llvm.org/show_bug.cgi?id=22803).

ry:
Does this PR help us towards reducing Lockers? If not what situation is it helping? I'm worried we're going to need to do some large refactors to resolve the Locker/scope design oversight, and that additional machinery (e.g. IsolateAnnex) will just get in the way of implementing the corrections.

src/isolate.rs
piscisaureus: Good catch!

piscisaureus:
> Random observation: an isolate currently only has 4 embedder slots, so that means there are only 3 left after this PR. 🤷‍♂

Luckily we're only using 2 right now in Deno. And if we actually need more we can easily add it to IsolateAnnex.

ry:
Bert and I had a long conversation about this patch. I took some rather unintelligible notes here https://gist.github.com/ry/4ffe794fbf248909b21259877ca47d84

Issue #265, opened 2020-02-04
support arm platforms


afinch7:
I've done a little research on this. It looks like cross compiling to arm64 works fine, but it requires some  setup and small changes:

* Sysroot usage is required (need to remove `use_sysroot = false` from gn options when compiling to arm)
* `target_cpu` needs to be set correctly (and maybe `v8_target_cpu` not sure yet)

Other than these and installing the correct toolchain it's not to hard to compile for arm64 successfully. I would like to get a build with tests running in CI for at least arm64 and maybe other arm platforms if possible.

ry:
I would also like to get arm64 build settled. Looks like Github Action has support for arm https://github.blog/changelog/2019-12-03-github-actions-self-hosted-runners-on-arm-architectures/

joshgillies:
Any chance of getting armv7 builds off the back of the work done in #266?

piscisaureus:
Arm64 has been supported for a while now.

I'm not aware of anyone working on armv7 support, and we (as in "the maintainers") are not going to in the foreseeable future. 

However if someone submits PR I'd be willing to work with them to get it landed.

PR #266, opened 2020-02-04
aarch64/arm64 builds and tests

afinch7:
ref #265

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   build:
-    name: ${{ matrix.config.os }} ${{ matrix.config.TARGET }}
+    name: ${{ matrix.config.os }} ${{ matrix.config.target }}
     if: |
       github.event_name == 'push' ||
       !startsWith(github.event.pull_request.head.label, 'denoland:')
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -39,6 +44,19 @@ jobs:
         with:
           python-version: "2.7.x"
           architecture: x64
+        
+      - name: Install cross compile deps
```
ry: No luck with compiling directly on arm?

afinch7: Haven't really tried since I added arm64 builds last time. Last time it was problem of missing build  tool binaries on these platforms. It would also be very slow. Without setting up our own github actions runners the only way to build on arm is qemu emulation.

piscisaureus: Would it be possible to compile on a Raspberry Pi 4 (I'm mostly concerned about it having only 4GB of RAM).

piscisaureus: Actually thinking about it -- cross build is probably a lot faster. But it'd be nice to run the tests on real HW.

afinch7: > Would it be possible to compile on a Raspberry Pi 4 (I'm mostly concerned about it having only 4GB of RAM).

It might be I got some versions of v8 to compile fine on a Pi 3, but it's really slow(2-3 hours). Not sure about the Pi 4 though.

.github/workflows/ci.yml
```diff
@@ -67,7 +85,7 @@ jobs:
           echo "::set-env name=RUSTC_WRAPPER::sccache"
 
       - name: Test
-        run: cargo test -vv --locked --all-targets
+        run: cargo test -vv --locked --all-targets --target ${{ matrix.config.TARGET }}
```
ry: Does cargo automatically run the tests in qemu when the target != host?

afinch7: No. This relies on `binfmt_misc`(kernel module) which reads file headers and executes with the correct interpreter. Works sorta like a shebang.

PR #272, opened 2020-02-11
Remove v8::Locker

ry:
This patch clarifies that v8::Isolate is a single threaded creature
which can only be accessed from other threads in special circumstances.
To ensure optimal operation in Deno, we remove v8::Locker, which ought
to be unnecessary when a thread is dedicated to each Isolate and the
Isolates never move between threads.

There are valid use-cases for v8::Locker, and we hope to address them in
future revisions.

### DIFF
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -163,6 +163,9 @@ impl Isolate {
     CreateParams::new()
   }
 
+  // TODO(ry) The return value here is pretty sloppy:
+  // Arc<Mutex<Option<*mut Isolate>>>. Only the V8 thread ever modifies it,
+  // mutex shouldn't be necessary, nor does it need to be clone-able.
   pub fn thread_safe_handle(&mut self) -> Arc<IsolateHandle> {
     let slot_ptr = self.get_data(0) as *mut IsolateHandle;
     if slot_ptr.is_null() {
@@ -173,7 +176,10 @@ impl Isolate {
       }
       return handle;
     } else {
-      todo!()
+      let handle = unsafe { Arc::from_raw(slot_ptr) };
+      // Call into_raw so again to avoid double free.
+      let _ptr = Arc::into_raw(handle.clone());
+      handle
     }
   }
 
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -2,13 +2,13 @@
 use crate::array_buffer::Allocator;
 use crate::external_references::ExternalReferences;
 use crate::promise::PromiseRejectMessage;
-use crate::scope_traits::InIsolate;
 use crate::support::intptr_t;
 use crate::support::Delete;
 use crate::support::Opaque;
 use crate::support::UniqueRef;
 use crate::Context;
 use crate::Function;
+use crate::InIsolate;
 use crate::Local;
 use crate::Message;
 use crate::Module;
@@ -22,8 +22,6 @@ use std::ffi::c_void;
 use std::ops::Deref;
 use std::ops::DerefMut;
 use std::ptr::NonNull;
-use std::sync::Arc;
-use std::sync::Mutex;
 
 pub type MessageCallback = extern "C" fn(Local<Message>, Local<Value>);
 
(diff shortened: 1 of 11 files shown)


src/isolate.rs
```diff
@@ -282,23 +289,77 @@ impl Isolate {
...
+      let handle = Arc::new(Mutex::new(x));
+      let handle_ptr = Arc::into_raw(handle.clone());
+      unsafe {
+        isolate.set_data(0, handle_ptr as *mut c_void);
+      }
+      IsolateHandle(handle)
+    } else {
+      let handle = unsafe { Arc::from_raw(slot_ptr) };
+      // Call into_raw so again to avoid double free.
+      let _ptr = Arc::into_raw(handle.clone());
```
piscisaureus: How about `IsolateHandle(handle.clone())`. 
That explains more clearly what is going on here than the code and comment above.

src/isolate.rs
```diff
@@ -282,23 +289,77 @@ impl Isolate {
...
+      let handle = unsafe { Arc::from_raw(slot_ptr) };
+      let some_ptr = handle.lock().unwrap().take();
+      assert!(some_ptr.is_some());
+    }
+  }
+
+  fn new(isolate: &mut Isolate) -> Self {
+    let slot_ptr = isolate.get_data(0) as *mut Mutex<Option<*mut Isolate>>;
+    if slot_ptr.is_null() {
+      let x: Option<*mut Isolate> = Some(&mut *isolate);
```
piscisaureus: It would be good to see this through and implement Global::drop() using the reference count you get with this.

src/isolate.rs
```diff
@@ -282,23 +289,77 @@ impl Isolate {
...
+  unsafe fn dispose(&mut self) {
+    IsolateHandle::dispose(self);
+    v8__Isolate__Dispose(self)
+  }
+}
+
+// TODO(ry) Use AtomicPtr? Use Cell? Only one thread ever modifies it.
+/// Thread safe reference to an Isolate. This handle will out-live the Isolate
+/// itself, when that happens all methods on it will return false.
+pub struct IsolateHandle(Arc<Mutex<Option<*mut Isolate>>>);
```
piscisaureus: Why is a Mutex needed? And why care to remove Locker if you sneak a Mutex in behind the curtains?

src/isolate.rs
```diff
@@ -282,23 +289,77 @@ impl Isolate {
...
+
+  /// Disposes the isolate.  The isolate must not be entered by any
+  /// thread to be disposable.
+  unsafe fn dispose(&mut self) {
+    IsolateHandle::dispose(self);
+    v8__Isolate__Dispose(self)
+  }
+}
+
+// TODO(ry) Use AtomicPtr? Use Cell? Only one thread ever modifies it.
```
piscisaureus: > What's the "it" that only one thread ever modifies?

src/isolate.rs
piscisaureus: What's the advantage of creating this thing lazily? It makes hard to reason about with virtually no benefits. 
Note that creating an Isolate is a quite expensive operation. Allocating an Arc certainly is not going to register on any benchmark.

tests/test_api.rs
```diff
@@ -1720,11 +1685,12 @@ fn external_references() {
   let startup_data = {
     let mut snapshot_creator =
       v8::SnapshotCreator::new(Some(&EXTERNAL_REFERENCES));
-    let isolate = snapshot_creator.get_isolate();
-    let mut locker = v8::Locker::new(&isolate);
-    let scope = locker.enter();
     {
-      let mut hs = v8::HandleScope::new(scope);
+      // TODO(ry) this shouldn't be necessary. workaround unfinished business in
```
piscisaureus: That would be a great area to pay some attention to.

src/isolate.rs
ry: I want to make sure that in the simple case (no threads) there's zero overhead

src/isolate.rs
ry: Not used or accessed by the the hot loops. ArcMutex is just my default too for these jobs. What would you suggest?

src/isolate.rs
ry: it = IsolateHandle

src/isolate.rs
piscisaureus: Versus 0.00000.1% overhead? 
Premature optimization much?

tests/test_api.rs
ry: Agree. But I want to limit my job here to removing Lockers and adding IsolateHandle (it's already quite a significant patch). This situation with the SnapshotCreator here seems to derive from some bad mutability assumptions from before - namely that v8::Locker was constructed with a `&Isolate` instead of a `&mut Isolate`.

src/isolate.rs
piscisaureus: * `pub struct IsolateHandle(NonNull<Isolate>);`
* create a `Box<AtomicUsize>`, cast it to `*mut AtomicUsize`, stuff that pointer in user data slot 0.

piscisaureus:
Can this patch be split in two parts, one that removes locker, one that does everything else?

src/isolate.rs
piscisaureus: The IsolateHandle never gets modified by any thread. What gets modified?

ry: take() modifies it in IsolateHandle::dispose

ry:
@piscisaureus As requested, I've split it into Locker removal only. There is a bit of entanglement, so a test was disabled.

src/isolate.rs
ry: Ideally I'll be able to strace deno and see no mutex locks being made. I think it's worth the one line it takes.

PR #273, opened 2020-02-11
WIP experimenting replacing the scopes with closures

ry:
Seems to work out nicely so far, allows us to remove lifetime parameter from Local.


bnoordhuis:
Is replacing ContextScope with manual enter/exit calls a transitional thing or ...? Because that doesn't seem like an improvement.

ry:
@bnoordhuis just transitional - it's a relatively small feature (relative to the design of handle scopes and lifetimes)

PR #274, opened 2020-02-11
Add IsolateHandle

ry:
part 2 of https://github.com/denoland/rusty_v8/pull/272

### DIFF
--- a/src/global.rs
+++ b/src/global.rs
@@ -141,9 +141,13 @@ impl<T> Drop for Global<T> {
       }
       Some(_)
         if unsafe {
-          self.isolate_handle.as_ref().unwrap().get_isolate_ptr()
-        }
-        .is_null() =>
+          self
+            .isolate_handle
+            .as_ref()
+            .unwrap()
+            .get_isolate_ptr()
+            .is_null()
+        } =>
       {
         // This global handle is associated with an Isolate that has already
         // been disposed.
--- a/src/global.rs
+++ b/src/global.rs
@@ -115,13 +115,12 @@ impl<T> Global<T> {
     self.set(scope, None);
   }
 
-  fn check_isolate(&self, isolate: &Isolate) {
+  fn check_isolate(&self, isolate: &mut Isolate) {
     match self.value {
       None => assert!(self.isolate_handle.is_none()),
-      Some(_) => assert!(
-        unsafe { self.isolate_handle.as_ref().unwrap().get_isolate_ptr() }
-          as *const _
-          == isolate
+      Some(_) => assert_eq!(
+        unsafe { self.isolate_handle.as_ref().unwrap().get_isolate_ptr() },
+        isolate as *mut _
       ),
     }
   }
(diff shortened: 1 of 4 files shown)


piscisaureus:
Ok so before we land this, add support for auto-dropping Global handles.

Tip: they will need to retain a reference-counted Isolate* -- whether these are weak or strong references is up to you. 

Note that Globals are Send and therefore their destructor can run in any thread. Preferably creating/dropping a Global would not require a mutex lock.

src/isolate.rs
```diff
@@ -283,23 +289,77 @@ impl Isolate {
...
+      let handle = Arc::new(Mutex::new(x));
+      let handle_ptr = Arc::into_raw(handle.clone());
+      unsafe {
+        isolate.set_data(0, handle_ptr as *mut c_void);
+      }
+      IsolateHandle(handle)
+    } else {
+      let handle = unsafe { Arc::from_raw(slot_ptr) };
+      // Call into_raw so again to avoid double free.
+      let _ptr = Arc::into_raw(handle.clone());
```
piscisaureus: This is a memory leak, as dropping the IsolateHandle this clone is created for does not decrease the reference count.

piscisaureus: Nevermind, I think I was mistaken.

src/global.rs
```diff
@@ -34,17 +35,15 @@ extern "C" {
 #[repr(C)]
 pub struct Global<T> {
   value: Option<NonNull<T>>,
-  isolate: Option<NonNull<Isolate>>,
+  isolate_handle: Option<IsolateHandle>,
```
ry: This implies mutex locks for accessing globals - which in the single threaded case is unnecessary overhead.

src/global.rs
```diff
@@ -130,30 +132,52 @@ impl<T> Default for Global<T> {
...
+        } =>
+      {
+        // This global handle is associated with an Isolate that has already
+        // been disposed.
+      }
+      addr @ Some(_) => unsafe {
+        // Destroy the storage cell that contains the contents of this Global.
+        v8__Global__Reset__0(
+          &mut *(addr as *mut Option<NonNull<T>> as *mut *mut Value),
+        )
```
ry: If the Isolate was destroyed already, but this one wasn't, wouldn't we have gotten a segfault? I think this is unreachable.

tests/test_api.rs
```diff
@@ -126,6 +126,28 @@ fn global_handles() {
...
+  let mut hs = v8::HandleScope::new(&mut isolate);
+  let scope = hs.enter();
+
+  let l1 = v8::String::new(scope, "foo").unwrap();
+  g1.set(scope, l1);
+
+  // Global 'g2' will be dropped _before_ the Isolate has been disposed.
+  let l2 = v8::String::new(scope, "bar").unwrap();
+  let _g2 = v8::Global::new_from(scope, l2);
+}
```
ry: This test ought to use drop(isolate) to really test the V8 segfault we're attempting to avoid.

ry: No test cases here show globals being accessed after isolate is destroyed.

piscisaureus: > This test ought to use drop(isolate) to really test the V8 segfault we're attempting to avoid.

No. Rust drops variables opposite of declaration order. This is guaranteed.

src/global.rs
piscisaureus: No mutexes are touched at any time while accessing globals.

tests/test_api.rs
piscisaureus: > No test cases here show globals being accessed after isolate is destroyed.

The type system doesn't allow that, so you're gonna have to live with it.

src/global.rs
piscisaureus: ??

src/global.rs
ry: I see - well almost - `Mutex::new(())` is called once, like when you create your first global variable. Will that show up in dtrace? I would guess not.

tests/test_api.rs
piscisaureus: https://doc.rust-lang.org/book/ch15-03-drop.html

"Rust automatically called drop for us when our instances went out of scope, calling the code we specified. Variables are dropped in the reverse order of their creation, so d was dropped before c. This example gives you a visual guide to how the drop method works; usually you would specify the cleanup code that your type needs to run rather than a print message."

ry: If you put a global in a box, store it into a static pointer, can you not get it to outlive the isolate long enough to hit the V8 segfault?

ry:
@piscisaureus looks fine despite my complaints - feel free to land this. I do think we eventually need some tests with Globals allocated on the heap.

src/global.rs
piscisaureus: * Check out https://github.com/denoland/rusty_v8/tree/proof
* Run `cargo test global -- --test-threads=1 --nocapture`

PR #275, opened 2020-02-12
Add compile_fail test for boxed lifetimes

ry:
just trying to debug something but ended up with a test


piscisaureus:
LGTM except do you mean "boxed local" instead of "boxed lifetime"?

Issue #248, opened 2020-01-23
add test for Isolate::get_entered_or_microtask_context


ry:
https://github.com/denoland/rusty_v8/commit/5e55cc5a5cf3fcc0e8d1e6777e5d8f8251fd8fe5

PR #279, opened 2020-02-13
Move get_*_context() methods to scope::Entered, remove InContext trait

piscisaureus:
The `get_current_context()` and `get_entered_or_microtask_context()`
methods now return `Option<Local<Context>>` to reflect that an isolate
may not have entered any context.

They're also moved from `Isolate` to `struct Entered` because it turns
out that the underlying V8 API calls actually create new local handles,
hence they should only be used inside an active HandleScope.

The `InContext` trait has been removed.

A test exercising `ContextScope` and the `get_*_context()` methods
mentioned above was added.

Closes: #248.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -478,57 +478,68 @@ bool v8__Value__SameValue(const v8::Value& self, v8::Value* that) {
   return self.SameValue(ptr_to_local(that));
 }
 
-v8::Uint32* v8__Value__ToUint32(const v8::Value& self, v8::Context* context) {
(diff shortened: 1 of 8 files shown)


piscisaureus:
@ry I'm going to land this now, feel free to revert if it causes problems while I'm away.

There does seem to be some flakiness going on with occasional segfaults happening in test 'request_interrupt_small_scripts' on linux (e.g. https://github.com/denoland/rusty_v8/runs/442488628), however that is unrelated to this patch.

PR #282, opened 2020-02-14
Isolate with state

ry:
Trying to come up with a usable pattern we can use for deno_core::Isolate ... not really a satisfactory solution

### DIFF
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -19,14 +19,23 @@ use crate::StartupData;
 use crate::String;
 use crate::Value;
 
+use std::any::Any;
+use std::any::TypeId;
+use std::cell::RefCell;
+use std::collections::HashMap;
 use std::ffi::c_void;
 use std::mem::replace;
 use std::ops::Deref;
 use std::ops::DerefMut;
 use std::ptr::null_mut;
 use std::ptr::NonNull;
+use std::rc::Rc;
 use std::sync::Arc;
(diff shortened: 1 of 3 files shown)


src/isolate.rs
```diff
@@ -247,12 +247,11 @@ impl Isolate {
 
     let type_id = TypeId::of::<S>();
     let state = states.get(&type_id).unwrap();
-    // TODO how to change Rc<RefCell<(dyn Any + 'static)>> into Rc<RefCell<S>>
-    // without transmute?
+    // TODO(ry) how to change Rc<RefCell<(dyn Any + 'static)>> into
+    // Rc<RefCell<S>> without transmute?
+    #[allow(clippy::transmute_ptr_to_ptr)]
     let state = unsafe { std::mem::transmute::<_, &Rc<RefCell<S>>>(state) };
```
bartlomieju: `if let Some(s) = state.downcast::<S> {}` ?

ry:
```
252 |     let state = state.downcast::<S>().unwrap();
    |                       ^^^^^^^^ method not found in `&std::rc::Rc<std::cell::RefCell<(dyn std::any::Any + 'static)>>`
```

ry:
Although this is unreviewed and the API is probably not the simplest, I'm going to land this in the interest of trying to upgrade Deno to the latest rusty_v8 API.

PR #283, opened 2020-02-14
Add Array::new_with_elements()

bnoordhuis:
A little faster than Array::new() followed by one or more set() calls.

Please keep as ~~two~~ three commits, thank you. :-)


ry:
> Please keep as two three commits, thank you. :-)

Oops 😬 sorry

PR #290, opened 2020-02-20
Add HandleScope::new2 hack to construct from const ref

ry:
as Locker used to...

### DIFF
--- a/src/handle_scope.rs
+++ b/src/handle_scope.rs
@@ -46,8 +46,8 @@ impl<'s> HandleScope {
     Scope::new(isolate, parent)
   }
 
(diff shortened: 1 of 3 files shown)


src/handle_scope.rs
```diff
@@ -45,6 +45,11 @@ impl<'s> HandleScope {
     let isolate: *mut Isolate = parent.isolate();
     Scope::new(isolate, parent)
   }
+
+  // TODO(ry) This is a hack so we can upgrade Deno.
+  pub fn new2(isolate: &Isolate) -> Scope<'s, Self> {
```
piscisaureus: This is OK but the function should be marked unsafe.

PR #301, opened 2020-03-06
Implement 'Eq' and 'PartialEq' traits for local handles (#301)

piscisaureus:


tests/test_api.rs
```diff
@@ -1995,41 +1996,88 @@ fn value_checker() {
     let value = eval(scope, context, "undefined").unwrap();
     assert!(value.is_undefined());
     assert!(value.is_null_or_undefined());
+    assert!(value == value);
+    assert!(value == v8::Local::<v8::Primitive>::try_from(value).unwrap());
+    assert!(value == v8::undefined(scope));
+    assert!(value != v8::null(scope));
+    assert!(value != v8::Boolean::new(scope, false));
+    assert!(value != v8::Integer::new(scope, 0));
```
ry: Does this pass clippy? Should these be `assert_eq!`?

piscisaureus:
> LGTM - except maybe use assert_eq

I would have, but it requires `Local<T>` to also implement `Debug` and I'm not sure how to do that meaningfully.

Issue #298, opened 2020-03-05
support taking HeapSnapshots


bartlomieju:
To figure out performance problems with [`deno_tcp.ts`](https://github.com/denoland/deno/blob/master/tools/deno_tcp.ts) (which are GC problems) we need to inspect what's going on on V8 heap.

This is partially related to #205 but can be implemented independently and used without full inspector support.

There's [`node-heapdump`](https://github.com/bnoordhuis/node-heapdump) which implements that functionality fo node. 

Grepping through [`node-heapdump/heapdump.cc`](https://github.com/bnoordhuis/node-heapdump/blob/master/src/heapdump.cc) it seems that APIs that we should port are:
- `Isolate::GetHeapProfiler`
- `HeapProfiler::TakeHeapSnapshot `
- `HeapSnapshot::Serialize`

CC @bnoordhuis for expert advice on the subject

bnoordhuis:
I can write bindings, that shouldn't be too complicated, but the question is where to send the data chunks to.

node-heapdump writes them out to disk to avoid having to round-trip through JS land because it's easy to hit out-of-memory conditions that way.

bartlomieju:
> I can write bindings, that shouldn't be too complicated, but the question is where to send the data chunks to.
> 
> node-heapdump writes them out to disk to avoid having to round-trip through JS land because it's easy to hit out-of-memory conditions that way.

That'd be awesome thanks! My direct need for this feature is to take some dumps of `deno_tcp.ts` heap; I haven't put much thought how to call that API, but I guess I could just directly bind it to global object (`globalThis.takeHeapSnapshot()`) and write out to disk - but that would be done in `deno_core`

bnoordhuis:
I'll open a PR here tomorrow.

bartlomieju:
Done in #302

PR #302, opened 2020-03-08
Add Isolate::take_heap_snapshot()

bnoordhuis:
This doesn't really follow the current V8 API (it's pretty close to how
V8 used to be back in 2012 though.) However:

1. The C++ API is very C++-y and doesn't carry over well to Rust, and
2. It addresses the immediate need of being able to take heap snapshots.

I can expand it into something that mimics the V8 API more closely later
but right now it's Sunday afternoon and I want to go bouldering. :-)

Refs #298

cc @bartlomieju - even if this doesn't get merged it should be enough to help you track down that memory issue locally.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -6,6 +6,7 @@
 #include "v8/include/libplatform/libplatform.h"
 #include "v8/include/v8-inspector.h"
 #include "v8/include/v8-platform.h"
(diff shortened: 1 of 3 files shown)


bartlomieju:
Thanks @bnoordhuis! I did manage to take heap snapshot for code of interest!

PR #306, opened 2020-03-14
Upgrade V8 to 8.2.308

piscisaureus:


piscisaureus:
https://s3.amazonaws.com/deno.land/gn_ninja_binaries.tar.gz needs to be updated to contain the latest version of gn. The current version doesn't support the `{{frameworks}}` and `{{framework_dirs}}` substitutions.

PR #315, opened 2020-03-16
Simplify .gn to reduce size of static lib

ry:
We noticed that the [static lib binaries](https://github.com/denoland/rusty_v8/releases/tag/v0.3.6) were quite large. 176mb for mac/debug.

We aren't sure what exactly is causing this, but I thought by simplifying the `.gn` file we might discover something. With these changes I get:
```
> ls -lh target/debug/gn_out/obj/librusty_v8.a
-rw-r--r--  1 rld  staff   940K Mar 16 18:54 target/debug/gn_out/obj/librusty_v8.a
```

### DIFF
--- a/.gn
+++ b/.gn
@@ -29,6 +29,7 @@ default_args = {
   clang_use_chrome_plugins = false
   v8_enable_i18n_support = false
   v8_monolithic = false
+  v8_use_external_startup_data = false
   v8_use_snapshot = true
   is_component_build = false
 }
--- a/.gn
+++ b/.gn
@@ -19,33 +19,14 @@ secondary_source = "//v8/"
 default_args = {
   linux_use_bundled_binutils = false
 
-  # TODO(ry) We may want to turn on CFI at some point. Disabling for simplicity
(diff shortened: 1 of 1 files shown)


ry:
Link errors when running `cargo test` tho. Many undefined symbols in the `.a`.

piscisaureus:
Can you check what effect this has on the size of deno release executables on different platforms (linux in particular)?

ry:
I only have mac-debug to report for version 8040a7cbc4c7d559ec1b4168893a735f881c8ca7
```
> ls -lh target/debug/gn_out/obj/librusty_v8.a
-rw-r--r--  1 rld  staff   604M Mar 16 23:42 target/debug/gn_out/obj/librusty_v8.a
```
-_-

piscisaureus:
I think it might still be worthwhile to land this because it makes the .gn file simpler. It's just that symbol_level=1 needs to stay.

ry:
Using `symbol_level = 0` brings it down to 63M.

piscisaureus:
> Using `symbol_level = 0` brings it down to 63M.

Is that desirable? It means that when v8 crashes you get no stack trace at all.

PR #316, opened 2020-03-17
Support binary downloads

ry:


tools/download_file.py
```diff
@@ -0,0 +1,73 @@
...
+    """Download url into output_file."""
+    CHUNK_SIZE = 4096
+    TOTAL_DOTS = 10
+    num_retries = 3
+    retry_wait_s = 5  # Doubled at each retry.
+
+    while True:
+        try:
+            response = urlopen(url)
+            content_length = int(response.headers['content-length'])
```
piscisaureus: Is this reliable? When I switched the gn/ninja download to come from github I found that the content-length header was not always present.

build.rs
```diff
@@ -23,9 +28,12 @@ fn main() {
     .map(|s| s.starts_with("rls"))
     .unwrap_or(false);
 
-  if !(is_trybuild || is_cargo_doc | is_rls) {
+  if BINARY_BUILD {
```
piscisaureus: `if cfg!(feature = "binary")` not working here?

build.rs
```diff
@@ -166,6 +174,70 @@ fn download_ninja_gn_binaries() {
...
 }
 
+fn static_lib_url() -> (String, String) {
+  let target = std::env::var("TARGET").unwrap();
+  let profile = std::env::var("PROFILE").unwrap();
+  assert!(profile == "release" || profile == "debug");
+  let base = "https://github.com/denoland/rusty_v8/releases/download/v0.3.6/";
+  #[cfg(windows)]
+  {
+    let url = format!("{}/rusty_v8_{}_{}.lib", base, profile, target);
```
piscisaureus: There are no windows 'debug' builds.

aniketbiprojit:
How do I consume the binary build in deno_core Cargo.toml? Even using `V8_FROM_SOURCE=1` how does that happen?

PR #321, opened 2020-03-23
Implement GetPrototype and SetPrototype on objects

Protryon:


tests/test_api.rs
```diff
@@ -2805,3 +2805,32 @@ fn take_heap_snapshot() {
...
+      v8::String::new(scope, "test_proto_key").unwrap().into();
+    let value_local: v8::Local<v8::Value> =
+      v8::String::new(scope, "test_proto_value").unwrap().into();
+    proto_obj.set(context, key_local, value_local);
+    obj.set_prototype(context, proto_obj.into());
+
+    let sub_gotten = obj.get(scope, context, key_local).unwrap();
+    assert!(sub_gotten.is_string());
+    let sub_gotten = sub_gotten.to_string(scope).unwrap();
+    assert_eq!(sub_gotten.to_rust_string_lossy(scope), "test_proto_value");
```
bnoordhuis: It would be good to test the behavior of:

1. setting the prototype to a non-object value (null, undefined), and
2. retrieving the prototype of a JS-created object `var obj = { __proto__: null }`:
```rust
let val = eval(scope, context, "({ __proto__: null })").unwrap();
let obj = val.to_object(scope).unwrap();
assert!(obj.get_prototype(scope).is_null());
```

Issue #304, opened 2020-03-13
Support for Externals


JohnDoneth:
Currently it doesn't appear that we can use `v8::External`'s from Rust.

I would like to bind a Rust struct to the VM and the way I would do that in C++ would be to store the pointer using `obj->SetInternalField(0, External::New(myobjectpointer));` and then back again.

It would be awesome if we could `Box` up a struct and put it in an `External` by it's pointer. I'm not sure if there's a way to do that in the reverse without using unsafe but I would be happy with that as I know what types I am using.

The last thing to keep in mind is that a way to provide a finalizer to drop the box and free the memory should be possible in order to avoid holding onto memory forever. From a quick search it looks like `PersistentBase::SetWeak` is used to cleanup memory for bound C++ classes in Node.js. I'm not sure how this would translate to Rust.

piscisaureus:
Support for 'External' was added in https://github.com/denoland/rusty_v8/commit/82d02870286ebb9ae75639d775d0af6b97eca00.

Support for weak handles / finalizers is yet TBD, which I consider a separate issue.

PR #327, opened 2020-03-28
Weakable Globals & FFI Object Wrapping

Protryon:
This PR:
* Fixes an apparent non-deterministic test that depended on a script id
* Adds `Isolate::request_garbage_collection_for_testing` for GC-related tests
* Adds internal field support to `Object`s through:
    * `Object::internal_field_count`
    * `Object::get_internal_field` -- unsafe
    * `Object::set_internal_field` -- unsafe
    * `Object::get_internal_field_ptr` -- unsafe
    * `Object::set_internal_field_ptr` -- unsafe
    * `ObjectTemplate::internal_field_count`
    * `ObjectTemplate::set_internal_field_count`
    * The non-ptr variants could be made safe in the future by fully managing internal fields, but inside V8 the non-ptr and ptr variants overlap, so these have been marked unsafe.
* Adds a subset of weak support to `Global`s through:
    * `Global::set_weak`
    * `Global::is_weak`
    * `Global::clear_weak`

This PR partially addresses #304.


ry:
@protryon Thanks - the patch generally looks good - except for ObjectWrap. It certainly looks useful but in rusty_v8 we’re trying to stick as close to the C++ API as possible. Could you remove it?

Protryon:
@ry Done. I have made `Global::Weakable` a public unsafe trait in order to allow foreign implementation of analogs to `ObjectWrap`. I figure I'll publish a crate for `ObjectWrap` once this lands.

piscisaureus:
@Protryon 

I'm generally in favor of adding these bindings; thank you so much for all the hard work you put into it.

However I have doubts about safely of auto-resetting "phantom" handles as it's done in the DefaultWeakable trait implementation. In Rust all types, including Globals, are freely movable; if the user moves the global to a different memory address V8 will not find out, risking memory corruption. I think that generally it is better to just provide weak callbacks and let the user decide what to do with them.

Second, it interferes a bit with an item on my roadmap which is to make Globals behave much more like Local handles. That is, they would not be allowed to be empty (use `Option<Global<T>>` instead for nullable handles) and it would not be mandatory to always convert them to Locals before they can be passed into various APIs. I believe it is possible to do this safely because the borrow checker can verify that a Global doesn't "disappear" while in use. If we let weak handles auto-reset that is no longer a possibility.

I believe a better solution would be to define a separate type for weak handles, much like the Arc/Weak split in Rust.

Now I won't be working on this very soon, and your patch generally looks solid, so I can land it in the meantime. However there is a risk that I might have to undo parts of it later on as we move forward with these plans.

Protryon:
Replaced by #335 .

PR #330, opened 2020-04-01
add v8::proxy

mhvsa:
Adds v8::Proxy

I need this for the https://github.com/denoland/deno/pull/4535


ry:
@mhvsa Object::get is an example of how MaybeLocal is translated into Rust https://github.com/denoland/rusty_v8/blob/v0.3.9/src/object.rs#L180-L190
https://github.com/denoland/rusty_v8/blob/v0.3.9/tests/test_api.rs#L1162

mhvsa:
@ry Thank you. 

I don't understand why I am getting `undefined reference to `v8__Proxy__New'`. I have a super simple test (create proxy, and check if there is something) and it fails.

I have `v8__Proxy__New` definition in `bindings.cc` and `extern "C" { fn v8__Proxy__New ...` in `proxy.rs`.

What more is needed to reference it?

bartlomieju:
> @ry Thank you.
> 
> I don't understand why I am getting `undefined reference to `v8__Proxy__New'`. I have a super simple test (create proxy, and check if there is something) and it fails.
> 
> I have `v8__Proxy__New` definition in `bindings.cc` and `extern "C" { fn v8__Proxy__New ...` in `proxy.rs`.
> 
> What more is needed to reference it?

That sometimes happens during development that linker gets confused, I just run `cargo clean` to remove all files from `target/` directory (note that it triggers full rebuild including V8)

ry:
You need to build from source when making changes. Make sure env var `V8_FROM_SOURCE=1` is set.

src/proxy.rs
```diff
@@ -0,0 +1,46 @@
...
+
+extern "C" {
+  fn v8__Proxy__New(
+    context: *mut Context,
+    target: *mut Object,
+    handler: *mut Object,
+  ) -> *mut Proxy;
+  fn v8__Proxy__GetHandler(proxy: *mut Proxy) -> *mut Value;
+  fn v8__Proxy__GetTarget(proxy: *mut Proxy) -> *mut Value;
+}
```
ry, line 18: This looks correct; same for the additions in binding.cc

mhvsa:
Thank you @ry @bartlomieju 

I run tests locally and they pass

mhvsa:
review ready @ry

PR #333, opened 2020-04-02
Map implement size and as_array

FrankBlue:


src/object.rs
```diff
@@ -278,3 +281,16 @@ impl Array {
     unsafe { v8__Array__Length(self) }
   }
 }
+
+impl Map {
+  pub fn size(&self) -> usize {
+    unsafe { v8__Map__Size(self) }
+  }
+  pub fn as_array<'sc>(
```
ry, line 289: Please add the docstring from v8.h
```
/// Returns an array of length Size() * 2, where index N is the Nth key and
/// index N + 1 is the Nth value.
```

tests/test_api.rs
```diff
@@ -2209,11 +2210,32 @@ fn value_checker() {
...
+        == v8::String::new(scope, "s").unwrap()
+    );
+    assert!(
+      map_array.get_index(scope, context, 2).unwrap()
+        == v8::String::new(scope, "v").unwrap()
+    );
+    assert!(
+      map_array.get_index(scope, context, 3).unwrap()
+        == v8::Number::new(scope, 8f64)
+    );
```
ry, line 2238: Please start a new test function for map rather than putting this in the proxy test.

FrankBlue: Thanks, I add test code and comment

FrankBlue: I changed it

ry:
Please run `cargo clippy --all-targets` and fix the errors.

PR #337, opened 2020-04-04
Add bindings for 'Object::get_(own)_property_names()'

Protryon:


src/object.rs
```diff
@@ -246,6 +255,49 @@ impl Object {
...
+  /// but the returned array doesn't contain the names of
+  /// properties from prototype objects.
+  pub fn get_own_property_names<'sc>(
+    &self,
+    scope: &mut impl ToLocal<'sc>,
+    context: Local<Context>,
+  ) -> Option<Vec<String>> {
+    let raw_name_ptr =
+      unsafe { v8__Object__GetOwnPropertyNames(self, context) };
+    self.resolve_js_array(scope, context, raw_name_ptr)
```
piscisaureus, line 286: Why does the returned value need to be processed by `resolved_js_array()`
I'm in favor of adding this binding, but we should just return the `v8::Local<v8::Array>` that v8 gives us.
(Cf. goal #1 as stated in the readme: "the API should match the original API as closely as possible")

src/object.rs
```diff
@@ -71,6 +72,14 @@ extern "C" {
...
   fn v8__Object__GetIdentityHash(object: &Object) -> int;
   fn v8__Object__CreationContext(object: &Object) -> *mut Context;
+  fn v8__Object__GetOwnPropertyNames(
+    object: &Object,
+    context: Local<Context>,
+  ) -> *mut Value;
+  fn v8__Object__GetPropertyNames(
+    object: &Object,
+    context: Local<Context>,
+  ) -> *mut Value;
```
piscisaureus, line 82: s/Value/Array/

src/object.rs
```diff
@@ -71,6 +72,14 @@ extern "C" {
   ) -> MaybeBool;
   fn v8__Object__GetIdentityHash(object: &Object) -> int;
   fn v8__Object__CreationContext(object: &Object) -> *mut Context;
+  fn v8__Object__GetOwnPropertyNames(
+    object: &Object,
+    context: Local<Context>,
+  ) -> *mut Value;
```
piscisaureus, line 78: s/Value/Array/

src/binding.cc
```diff
@@ -781,6 +781,14 @@ v8::Context* v8__Object__CreationContext(v8::Object& self) {
   return local_to_ptr(self.CreationContext());
 }
 
+v8::Value* v8__Object__GetOwnPropertyNames(v8::Object& self, v8::Local<v8::Context> context) {
+  return maybe_local_to_ptr(self.GetOwnPropertyNames(context));
+}
+
+v8::Value* v8__Object__GetPropertyNames(v8::Object& self, v8::Local<v8::Context> context) {
```
piscisaureus, line 788: `v8::Value*` -> `v8::Array`

src/binding.cc
```diff
@@ -781,6 +781,14 @@ v8::Context* v8__Object__CreationContext(v8::Object& self) {
   return local_to_ptr(self.CreationContext());
 }
 
+v8::Value* v8__Object__GetOwnPropertyNames(v8::Object& self, v8::Local<v8::Context> context) {
```
piscisaureus, line 784: `v8::Value*` -> `v8::Array`

src/object.rs
Protryon: Gotcha, just fit my needs at that exact moment. Will fixup in a bit.

PR #357, opened 2020-04-20
Make Isolate take ownership of CreateParams

piscisaureus:


piscisaureus:
Not perfect, but landing it now b/c I need to get on with other things.

PR #360, opened 2020-04-22
WIP: get/set_data_2

piscisaureus:


src/isolate.rs
```diff
@@ -181,6 +201,57 @@ impl Isolate {
...
+
+  /// Safe alternative to Isolate::set_data
+  ///
+  /// Use with Isolate::get_slot and Isolate::get_slot_mut to associate state
+  /// with an Isolate.
+  ///
+  /// This method gives ownership of value to the Isolate. Exactly one object of
+  /// each type can be associated with an Isolate. If called more than once with
+  /// an object of the same type, the earlier version will be dropped and
+  /// replaced.
```
ry, line 239: This mechanic is pretty wacky and difficult to describe. I wish we could come up with something that was simpler.

ry:
This is good enough to land and iterate on.

PR #364, opened 2020-04-23
Explicitly drop slots when disposing an isolate (#364)

piscisaureus:


tests/slots.rs
```diff
@@ -138,6 +141,7 @@ fn slots_layer1() {
   core_isolate.run_microtasks();
   drop(core_isolate);
   assert_eq!(drop_count.load(Ordering::SeqCst), 1);
+  drop(isolate_handle);
```
ry, line 144: 👍

PR #367, opened 2020-05-03
Turn on enable_stripping and exclude_unwind_tables

ry:
To reduce binary size


piscisaureus:
What's the size difference? Also, maybe do this only for release builds?

ry:
I didn't notice a size difference on osx debug build. It needs more analysis to figure out if there's any benefit.

piscisaureus:
Closing as there are no plans to land this and the analysis is stalled.

PR #368, opened 2020-05-06
Upgrade V8 to 8.4.300

piscisaureus:


piscisaureus:
I've already seen a green build. Running it again because the cache had to be fixed.

src/binding.cc
```diff
@@ -1600,13 +1601,13 @@ void v8_inspector__StringBuffer__DELETE(v8_inspector::StringBuffer* self) {
   delete self;
 }
 
-const v8_inspector::StringView* v8_inspector__StringBuffer__string(
-    v8_inspector::StringBuffer* self) {
-  return &self->string();
+three_pointers_t v8_inspector__StringBuffer__string(
+    const v8_inspector::StringBuffer& self) {
+  return make_pod<three_pointers_t>(self.string());
 }
```
ry, line 1607: Related to v8 upgrade?

.github/workflows/ci.yml
```diff
@@ -99,7 +99,7 @@ jobs:
         run: echo "::set-env name=CARGO_HOME::$(pwd)/target/cargo"
 
       - name: Cache
-        uses: actions/cache@master
+        uses: denoland/github-actions-cache@stable-prerelease
```
ry, line 102: Why do you need to fork this?

src/binding.cc
piscisaureus: https://github.com/v8/v8/commit/844fe8f7d965fcb305bd1b71fe6b34628fc4d1fd

.github/workflows/ci.yml
piscisaureus: Because actions/cache@master is broken and the most recent tagged version doesn't support caching multiple directories.

PR #371, opened 2020-05-06
Fix: SnapshotCreator's internal Isolate does not get initialized

piscisaureus:


tests/test_api.rs
```diff
@@ -1761,6 +1761,9 @@ fn snapshot_creator() {
       // the scope type system.
       let mut isolate = unsafe { snapshot_creator.get_owned_isolate() };
 
+      // Check that the SnapshotCreator isolate has been set up correctly.
+      let _ = isolate.thread_safe_handle();
+
```
ry, line 1766: I checked that it fails on master 👍

PR #373, opened 2020-05-14
Remove python as a required dependency for non-source builds

GeorgeHahn:
This change allows non-`V8_FROM_SOURCE` builds to succeed without python available. This is done by falling back to curl if the download with python fails. Curl is installed by default across most Linux, Mac, and Windows installations, so this fallback should typically succeed.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -186,6 +186,39 @@ fn static_lib_url() -> (String, String) {
   }
 }
 
+fn download_file(url: String, filename: PathBuf) {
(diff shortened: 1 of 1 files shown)


piscisaureus:
I'm in favor of this. In fact, I'd prefer to try and use `curl` only for not-from-source builds.
Otherwise the likelihood of this breaking is pretty high, given that all the maintainers are going to have python2 installed anyway.

Issue #381, opened 2020-05-26
Add support for GetHeapStatistics


dallen66:
Can you add support for GetHeapStatistics. I need to monitor the isolate memory usage. I try to implement it my self but failed.  My c++ and binding is not great.

binding.cc
```c
void v8__Isolate__GetHeapStatistics(v8::Isolate* isolate) {
  v8::HeapStatistics stats;
  return isolate->GetHeapStatistics(&stats);
}
```

isolate.rs
```c
extern "C" {
  fn v8__Isolate__GetHeapStatistics(isolate: *mut Isolate);
}
/// Get statistics about the heap memory usage. 
pub fn get_heap_statistics(&mut self) {
  unsafe { v8__Isolate__GetHeapStatistics(self) }
}
```
I get this when I try to run the function.
```
undefined reference to `v8__Isolate__GetHeapStatistics'
          collect2: error: ld returned 1 exit status
```
I build from source with ``V8_FROM_SOURCE=1 cargo build -vv`` Is there some else I need to do.

ry:
@dallen66 The binding to GetHeapStatistics is welcome. If you start a PR with your patch, we can help get it done.

> undefined reference to `v8__Isolate__GetHeapStatistics'
>          collect2: error: ld returned 1 exit status

I believe this is a bug on our side. You need to run "cargo clean" and then "V8_FROM_SOURCE=1 cargo build -vv".

dallen66:
Thank will try that. I will also work on the unit test.

dallen66:
I made the required patches.
https://github.com/denoland/rusty_v8/pull/383

dallen66:
These functions should be moved to the IsolateHandle struct to make them work across threads.
I'll wait until you check what I did before I make any changes.

bnoordhuis:
Sorry, I forgot add a `Fixes` tag but this was fixed by #440.

PR #383, opened 2020-05-26
Add GetHeapStatistics and LowMemoryNotification

dallen66:
It compile and return values, not sure if they are correct. Hopefully I did everything correctly.
I also add LowMemoryNotification but not sure how to test for it. Test will fail

### DIFF
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -121,6 +121,8 @@ extern "C" {
     callback: extern "C" fn(*mut c_void, *const u8, usize) -> bool,
     arg: *mut c_void,
   );
+  fn v8__Isolate__LowMemoryNotification(isolate: *mut Isolate);
+  fn v8__Isolate__GetHeapStatistics(isolate: *mut Isolate, stats:*mut HeapStatistics);
 }
 
 #[repr(C)]
@@ -345,6 +347,17 @@ impl Isolate {
   /// Any exceptions thrown by microtask callbacks are swallowed.
   pub fn run_microtasks(&mut self) {
     unsafe { v8__Isolate__RunMicrotasks(self) }
+  }
+
+  /// Optional notification that the system is running low on memory. 
+  /// V8 uses these notifications to attempt to free memory.
+  pub fn low_memory_notification(&mut self) {
+    unsafe { v8__Isolate__LowMemoryNotification(self) }
+  }
+
+  /// Get statistics about the heap memory usage. 
+  pub fn get_heap_statistics(&mut self,  stats:&mut HeapStatistics)  {
+    unsafe { v8__Isolate__GetHeapStatistics(self, stats) }
   }
 
   /// Enqueues the callback to the default MicrotaskQueue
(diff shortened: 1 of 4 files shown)


src/isolate.rs
```diff
@@ -579,3 +592,112 @@ impl DerefMut for OwnedIsolate {
...
+  }
+
+  // Is a 0/1 boolean, which signifies whether the --zap_code_space option is enabled or not.
+  pub fn does_zap_garbage(&self) -> usize {
+    self.does_zap_garbage_
+  }
+
+}
+
+impl Default for HeapStatistics {
```
piscisaureus, line 684: Why don't you just use `#[derive(Default)]` ?

src/isolate.rs
```diff
@@ -347,6 +349,17 @@ impl Isolate {
...
   }
 
+  /// Optional notification that the system is running low on memory. 
+  /// V8 uses these notifications to attempt to free memory.
+  pub fn low_memory_notification(&mut self) {
+    unsafe { v8__Isolate__LowMemoryNotification(self) }
+  }
+
+  /// Get statistics about the heap memory usage. 
+  pub fn get_heap_statistics(&mut self,  stats:&mut HeapStatistics)  {
```
piscisaureus, line 359: 1. I would suggest to change the function signature to `pub fn get_heap_statistics(&self) -> HeapStatistics`. 
(Note that the `self` type also changed from mutable to immutable.)

2. Since this struct is likely to change over time, can you add an assertion that `HeapStatistics` has the same size in Rust and C++. You can find an example how this is done elsewhere [here](https://github.com/denoland/rusty_v8/blob/9ad0d35718673e7afd1ec8d4d6b71ae6c50da60c/src/isolate_create_params.rs#L171-L172).

tests/test_api.rs
```diff
@@ -2938,3 +2938,37 @@ fn test_map_api() {
...
+
+#[test]
+fn get_heap_statistics() {
+  let _setup_guard = setup();
+  let mut isolate = v8::Isolate::new(Default::default());
+    
+  let mut stats = v8::HeapStatistics::default();
+  isolate.get_heap_statistics(&mut stats);
+
+  assert!((stats.total_heap_size() >= std::usize::MIN));
```
piscisaureus, line 2950: This is not a useful test because a `usize` is unsigned, which means that `foo >= 0` is always true...
Maybe you can make it more useful by checking that it has a reasonable value. It's going to be at least a few pages.

tests/test_api.rs
```diff
@@ -2938,3 +2938,37 @@ fn test_map_api() {
...
+}
+
+#[test]
+fn low_memory_notification() {
+  let _setup_guard = setup();
+  let mut isolate = v8::Isolate::new(Default::default());
+  
+  isolate.low_memory_notification();
+
+  assert!(false);
```
piscisaureus, line 2973: I'm fine with just a smoke test for this. I'd suggest to remove the `assert(false)`.

src/isolate.rs
```diff
@@ -579,3 +592,112 @@ impl DerefMut for OwnedIsolate {
...
+  external_memory_:usize,
+  peak_malloced_memory_:usize,
+  number_of_native_contexts_:usize,
+  number_of_detached_contexts_:usize,
+  does_zap_garbage_:usize,
+}
+
+impl HeapStatistics {
+  
+  // Number of bytes V8 has allocated for the heap. This can grow if usedHeap needs more.
```
piscisaureus, line 617: Can you use triple slash comments `///` so cargo doc picks these doc comments up.

src/isolate.rs
```diff
@@ -579,3 +592,112 @@ impl DerefMut for OwnedIsolate {
...
+  pub fn number_of_native_contexts(&self) -> usize {
+    self.number_of_native_contexts_
+  }
+
+  pub fn number_of_detached_contexts(&self) -> usize {
+    self.number_of_detached_contexts_
+  }
+
+  // Is a 0/1 boolean, which signifies whether the --zap_code_space option is enabled or not.
+  pub fn does_zap_garbage(&self) -> usize {
```
piscisaureus, line 678: Since this is either 0 or 1, the function should return a `bool` not a `usize`.

src/isolate.rs
```diff
@@ -579,3 +592,112 @@ impl DerefMut for OwnedIsolate {
...
+  pub fn total_heap_size_executable(&self) -> usize {
+    self.total_heap_size_executable_
+  }
+
+  // Committed size
+  pub fn total_physical_size(&self) -> usize {
+    self.total_physical_size_
+  }
+
+  //Available heap size
```
piscisaureus, line 632: If you don't mind, please include the full description as found in v8.h, and pay attention to formatting (e.g. add a space between `///` and text, add dots to the end of sentences, etc.)

piscisaureus:
When you're done, please run `cargo fmt`, then I don't have to nitpick about code style :)

src/isolate.rs
```diff
@@ -579,3 +592,112 @@ impl DerefMut for OwnedIsolate {
...
+  used_global_handles_size_:usize,
+  used_heap_size_:usize,
+  heap_size_limit_:usize,
+  malloced_memory_:usize,
+  external_memory_:usize,
+  peak_malloced_memory_:usize,
+  number_of_native_contexts_:usize,
+  number_of_detached_contexts_:usize,
+  does_zap_garbage_:usize,
+}
```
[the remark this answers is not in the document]
dallen66: Sorry I completely forget about this pull request. I got busy at work. @bnoordhuis I am not that good at C++ #381. This was my attempt to implement it, so if you have better solution go ahead and fix it.

dallen66:
> I went ahead and opened #440 for `Isolate::get_heap_statistics()`. It takes an approach that makes it a little more resilient to V8 ABI changes.

dallen66:
I closed this pull request because a new solution was uploaded #440

PR #385, opened 2020-05-31
Various scope fixes and a failing test for an open bug

piscisaureus:


tests/test_api.rs
```diff
@@ -72,6 +72,27 @@ fn handle_scope_numbers() {
   }
 }
 
+// TODO: the type checker is kumbaya with this but in reality the
+// `Local<Integer>` created at the end of the test is created in HandleScope
+// `hs2` and not in `hs1` as specified. When this local is accessed, which is
+// after `hs2` is destroyed, a crash happens.
+// #[test]
+// fn handle_scope_early_drop() {
```
ry, line 80: In the future use #[ignore] so we continue checking that it compiles until it can be fixed

piscisaureus: Makes sense, thanks.
I had no idea that existed.

piscisaureus:
It seems that I forgot to commit a bunch of files that should have been included in https://github.com/denoland/rusty_v8/commit/8213c0e4289effedabe84f22ec177de5ef1b4cb9

PR #386, opened 2020-05-31
Make to_local() take a closure that returns a raw pointer

piscisaureus:
This makes it possible to add a run-time check that verifies that the
specified closure is actually the one that contains the local handle.


piscisaureus:
Closing in favor of #406

PR #387, opened 2020-05-31
improve error message.

utam0k:
This is likely to be more user-friendly than the unwrap method.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -132,7 +132,7 @@ fn platform() -> &'static str {
 fn download_ninja_gn_binaries() {
   let root = env::current_dir().unwrap();
   let out_dir = env::var_os("OUT_DIR").expect(
-    "OUT_DIR expected target/debug//build/rusty_v8-xxxxxxxxxxxxxxxx/out/.",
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -131,8 +131,9 @@ fn platform() -> &'static str {
 
 fn download_ninja_gn_binaries() {
   let root = env::current_dir().unwrap();
-  // target/debug//build/rusty_v8-d9e5a424d4f96994/out/
-  let out_dir = env::var_os("OUT_DIR").unwrap();
+  let out_dir = env::var_os("OUT_DIR").expect(
+    "OUT_DIR expected target/debug//build/rusty_v8-xxxxxxxxxxxxxxxx/out/.",
```
piscisaureus, line 135: I don't want to be nitpicky but this is just as cryptic as the original error message.
Maybe...
`The 'OUT_DIR' environment is not set (it should be something like 'target/debug/rusty_v8-{hash}')`.

utam0k: Thank you so much! 
Corrected.

piscisaureus:
Thanks @utam0k

PR #406, opened 2020-06-14
Scopes refactor

piscisaureus:


tests/test_api.rs
```diff
@@ -55,13 +53,11 @@ fn handle_scope_numbers() {
   let _setup_guard = setup();
   let mut isolate = v8::Isolate::new(Default::default());
   {
-    let mut hs = v8::HandleScope::new(&mut isolate);
-    let scope1 = hs.enter();
+    let scope1 = &mut v8::HandleScope::new(&mut isolate);
```
ry, line 56: 👍

tests/test_api.rs
```diff
@@ -72,38 +68,33 @@ fn handle_scope_numbers() {
...
+  // Despite `local` living slightly longer than `scope2`, this test should
+  // not crash.
   let local = {
-    let mut hs2 = v8::HandleScope::new(hs1);
-    let _hs2 = hs2.enter();
-    v8::Integer::new(hs1, 123)
+    let scope2 = &mut v8::HandleScope::new(scope1);
+    v8::Integer::new(scope2, 123)
   };
   assert_eq!(local.value(), 123);
```
ry, line 83: 👍

src/scope.rs
```diff
@@ -1,292 +1,950 @@
...
+  impl<'s, C> Drop for HandleScope<'s, C> {
+    fn drop(&mut self) {
+      data::ScopeData::get_mut(self).notify_scope_dropped();
+    }
+  }
+
+  pub unsafe trait NewHandleScopeParam<'s> {
+    type NewScope: Scope;
+    fn get_isolate_mut(&mut self) -> &mut Isolate;
+  }
```
ry, line 292: Where is `NewScope` used?

src/scope.rs
```diff
@@ -1,292 +1,950 @@
...
+  #[derive(Debug)]
+  pub struct EscapableHandleScope<'s, 'e: 's, C = Context> {
+    data: NonNull<data::ScopeData>,
+    _phantom:
+      PhantomData<(&'s mut raw::HandleScope, &'e raw::EscapeSlot, &'s C)>,
+  }
+
+  impl<'s, 'e: 's> EscapableHandleScope<'s, 'e> {
+    #[allow(clippy::new_ret_no_self)]
+    pub fn new<P>(param: &'s mut P) -> P::NewScope
```
ry, line 345: > Where is `NewScope` used?

I guess here... TBH I'm a little confused about what's happening here. Maybe you can update[ the "Design of Scopes" text](https://github.com/denoland/rusty_v8/blob/acdcede4d67ad4fab6d969f71eb3f99b5b4a13cc/src/lib.rs#L30-L65)?

tests/test_ui.rs
```diff
@@ -7,19 +7,5 @@ fn ui() {
...
-  t.compile_fail("tests/compile_fail/try_catch_exception_lifetime.rs");
-  t.compile_fail("tests/compile_fail/try_catch_message_lifetime.rs");
-
-  // For unclear reasons rustc on Windows in Github Actions omits some
-  // diagnostic information, causing this test to fail. It might have something
-  // to do with this Rust issue: https://github.com/rust-lang/rust/issues/53081.
-  if cfg!(not(windows)) || env::var("GITHUB_ACTION").is_err() {
-    t.compile_fail("tests/compile_fail/handle_scope_escape_to_nowhere.rs");
-  }
+  t.compile_fail("tests/compile_fail/*.rs");
```
ry, line 10: I'm sure you checked this, but just a reminder to check that this is actually running the tests.

src/scope.rs
piscisaureus: https://github.com/denoland/rusty_v8/pull/406/files#diff-7a8f2fd4af9cbf971ac52735898b4eacR186

piscisaureus: https://github.com/denoland/rusty_v8/blob/af0738c595ef0d3bd5a1a1d333b81bc5138d5d57/src/scope.rs#L186-L193

tests/test_ui.rs
piscisaureus: It definitely does, and it is also the way we used to have it before one test had to be disabled on GHA-windows. 

Also notice that all the stderr files were updated - not something you'd expect when tests aren't running. https://github.com/denoland/rusty_v8/pull/406/files?file-filters%5B%5D=.stderr

src/scope.rs
```diff
@@ -1,292 +1,961 @@
...
-  unsafe fn enter_scope(buf: *mut Self, args: Self::Args);
-}
+use crate::TryCatch;
+use crate::Value;
+
+#[doc(inline)]
+pub use api::{CallbackScope, ContextScope, EscapableHandleScope, HandleScope};
+pub(crate) use data::ScopeData;
+
+mod api {
```
ry, line 32: The `api` and `data` modules presumably are some sort of organizational structure - but it's not obvious to me what they mean exactly - it would help readers to have a small triple-slash description at the top of each.

src/lib.rs
```diff
@@ -145,12 +101,8 @@ pub use primitives::*;
...
 pub use property_attribute::*;
 pub use proxy::*;
-pub use scope::CallbackScope;
-pub use scope::ContextScope;
-pub use scope::FunctionCallbackScope;
-pub use scope::PropertyCallbackScope;
-pub use scope::Scope;
-pub use scope_traits::*;
+#[doc(inline)]
+pub use scope::*;
```
ry, line 105: I would prefer if you explicitly listed exactly what is made public. [Elsewhere](https://github.com/denoland/deno/blob/0ffc99a61ddfa958a436beef0d003ecead630d0f/core/lib.rs#L26-L50) we've found it helpful to not use `pub use` wildcard statements.

```rust
pub use scope::api::{CallbackScope, ContextScope, EscapableHandleScope, HandleScope};
```

src/scope.rs
```diff
@@ -1,292 +1,961 @@
...
+    fn init(&mut self, isolate: *mut Isolate) {
+      unsafe {
+        let undefined = v8__Undefined(isolate) as *const Data;
+        let local = v8__Local__New(isolate, undefined);
+        let address_ptr =
+          &*local as *const Data as *mut Data as *mut raw::Address;
+        let slot = NonNull::new_unchecked(address_ptr);
+        let uninit_slot = self.0.replace(slot);
+        debug_assert!(uninit_slot.is_none());
+      }
```
ry, line 849: O_O

We really need valgrind.

src/scope.rs
```diff
@@ -1,292 +1,961 @@
...
-  Self: Sized,
-{
-  type Args;
-  unsafe fn enter_scope(buf: *mut Self, args: Self::Args);
-}
+use crate::TryCatch;
+use crate::Value;
+
+#[doc(inline)]
+pub use api::{CallbackScope, ContextScope, EscapableHandleScope, HandleScope};
```
ry, line 29: Remove. Unnecessary redirection.

src/scope.rs
```diff
@@ -1,292 +1,961 @@
...
+          .get_isolate()
+          .get_current_scope();
+        match current_scope {
+          Some(current_scope_nn) if current_scope_nn == self_nn => break,
+          Some(mut current_scope_nn)
+            if unsafe { current_scope_nn.as_ref().deferred_drop } =>
+          unsafe { current_scope_nn.as_mut().drop() }
+          _ => panic!("can't use scope that has been shadowed by another"),
+        }
+      }
```
ry, line 740: I guess we cannot claim this is a zero-overhead binding anymore...

src/isolate.rs
```diff
@@ -187,21 +183,43 @@ impl Isolate {
...
+  }
+
+  /// Updates the most recently entered scope.
+  pub(crate) fn set_current_scope(
+    &mut self,
+    current_scope: Option<NonNull<ScopeData>>,
+  ) {
+    let data =
+      current_scope.map(NonNull::as_ptr).unwrap_or_else(null_mut) as *mut _;
+    unsafe { v8__Isolate__SetData(self, 1, data) }
```
ry, line 215: Maybe define 
```rust
const CURRENT_SCOPE_SLOT: u32 = 1;
```

src/scope.rs
piscisaureus: The loop iterates 0 or 1 times

tests/test_api.rs
ry, line 83: The test here has been changed in a rather fundamental way. I agree that the previous version ought to pass (where the local is created with scope1), but I don't agree that the new version should - the local is outliving scope2.

piscisaureus:
Optimized the run-time is-scope-active check further. Now produces the following assembly for `v8::null()` and `v8::Value::to_integer()`:

#### v8::null()

```asm
 # pub fn null<'s>(scope: &mut HandleScope<'s, ()>) -> Local<'s, Primitive> {
 push    rsi
 sub     rsp, 32
 # unsafe { scope.cast_local(|sd| v8__Null(sd.get_isolate_ptr())) }.unwrap()
 mov     rsi, qword, ptr, [rcx]
 # match self.status.get() { (src\scope.rs:762)
 mov     al, byte, ptr, [rsi + 80]
 # ScopeStatus::ActiveReferenced => { (src\scope.rs:763)
 cmp     al, 1
 je      .LBB227_4
.LBB227_2:
 # « cold: drop weak scope »
.LBB227_4:
 # self.isolate.as_ptr() (src\scope.rs:842)
 mov     rcx, qword, ptr, [rsi + 16]
 # unsafe { scope.cast_local(|sd| v8__Null(sd.get_isolate_ptr())) }.unwrap()
 call    v8__Null
 # #[track_caller] (libcore\option.rs:385)
 test    rax, rax
 je      .LBB227_7
 # }
 add     rsp, 32
 pop     rsi
 ret
.LBB227_6:
 # « cold: unreachable!() »
.LBB227_7:
 # « cold: panic!() »
```

#### v8::Value::to_integer()

```asm
 # pub fn to_integer<'s>(
 #   &self,
 #   scope: &mut HandleScope<'s>,
 # ) -> Option<Local<'s, Integer>> {
 push    rsi
 push    rdi
 sub     rsp, 40
 mov     rsi, rcx
 # scope
 mov     rdi, qword, ptr, [rdx]
 # match self.status.get() { (src\scope.rs:762)
 mov     al, byte, ptr, [rdi + 80]
 # ScopeStatus::ActiveReferenced => { (src\scope.rs:763)
 cmp     al, 1
 je      .LBB392_4
.LBB392_2:
 # « cold: drop weak scope »
.LBB392_4:
 # match self.context.get().map(|nn| nn.as_ptr() as *const _) { (src\scope.rs:853)
 mov     rdx, qword, ptr, [rdi + 24]
 # #[inline] (libcore\option.rs:456)
 test    rdx, rdx
 # Some(context) => { (src\scope.rs:854)
 jne     .LBB392_6
 # « cold: v8__Isolate__GetCurrentContext() »
.LBB392_6:
 # .cast_local(|sd| v8__Value__ToInteger(self, &*sd.get_current_context()))
 mov     rcx, rsi
 add     rsp, 40
 pop     rdi
 pop     rsi
 jmp     v8__Value__ToInteger
.LBB392_7:
 # « cold: unreachable!() »
```

tests/test_api.rs
piscisaureus: Unfortunately Rust lifetimes cannot express the exact semantics of V8 HandlesScopes, so there's always a trade-off to make. The previous version of this test (which used to be disabled because it'd make v8 crash) is now rejected by the borrow checker: https://github.com/denoland/rusty_v8/blob/f6bd4d4a273913443378256683cf94704b66c02b/tests/compile_fail/handle_scope_lifetime_3.rs.

piscisaureus:
TODO: implement allow creating a HandleScope from either:
* `T where T: AsMut<Isolate>` or
* any `T` that *indirectly* implements `Deref<Target = Isolate> + DerefMut`  (not sure that is possible)

src/scope.rs
piscisaureus: The answer to your question is probably somewhere here: https://github.com/denoland/rusty_v8/blob/375895dd55b5da4acda7a7caa9d1ffa710f615b7/src/scope.rs#L432-L444

piscisaureus:
current Deno delta: https://github.com/denoland/deno/compare/master...piscisaureus:scopes_refactor_update

ry:
@piscisaureus Very nice to see the deno delta - thanks. One thing that sticks out is that there are several new instances of unsafe. 
Example:
```diff
-  let mut cbs = v8::CallbackScope::new_escapable(context);
-  let mut hs = v8::EscapableHandleScope::new(cbs.enter());
-  let scope = hs.enter();
+  let scope = &mut unsafe { v8::CallbackScope::new(context) };
 ```
It would be nice to have some sort of utility function in rusty_v8 to do that safely  (maybe impl From?)

**Comment by piscisaureus** [comment]
> It would be nice to have some sort of utility function in rusty_v8 to do that safely (maybe impl From?)

I think might be hard to get right -- everytime we're entering the vm we'd have to set some flag that says "callbacks expect - CallbackScope ok!", and then we have to clear that flag as we exit the VM. But who knows which callback could be triggered when...

I think the long term solution is what we are already doing for FunctionCallbackInfo/PropertyCallbackInfo, that is, rusty_v8 takes care of creating a scope and passes a safe one to the user callback. [here[(https://github.com/piscisaureus/deno/blob/e557c041577a4dd5421a53e5f666f5f414d22c58/core/bindings.rs#L365) for instance - there's no unsafe and not even a CallbackScope.

File: build.rs [hunk 12]
```diff
@@ -61,6 +61,10 @@ fn build_v8() {
     vec!["is_debug=false".to_string()]
   };
 
+  if !cargo_gn::is_debug() {
+    gn_args.push("v8_enable_handle_zapping=false".to_string());
+  }
```
ry, line 66: Can you add a comment here? What's this for?

src/scope.rs
```diff
@@ -591,42 +585,55 @@ pub(crate) mod data {
...
     escape_slot: Option<NonNull<raw::EscapeSlot>>,
     scope_type_specific_data: ScopeTypeSpecificData,
   }
 
   impl ScopeData {
+    /// Returns a mutable reference to the data associated with topmost scope
+    /// on the scope stack. This function does not automatically exit zombie
+    /// scopes, so it might return a zombie ScopeData reference.
+    pub(crate) fn get_current_mut(isolate: &mut Isolate) -> &mut Self {
+      loop {
```
piscisaureus, line 599: Remove nonsense loop

piscisaureus:
https://github.com/denoland/deno/compare/master...piscisaureus:scopes_refactor_update?expand=1

PR #427, opened 2020-07-28
Allow managing V8 heap limits

mraerino:
*Applies to https://github.com/denoland/deno/issues/6916*

V8 allows setting the `max_old_generation_size` and `max_young_generation_size` create parameters through a handy helper: https://chromium.googlesource.com/v8/v8/+/e423f0040333dd3ab410b3c6cab5b71a4f75fa6a/include/v8.h#6409

This PR also exposes the methods to register a callback when the limit is reached, so a crash of V8 can be avoided.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -178,6 +178,18 @@ bool v8__Isolate__AddMessageListener(v8::Isolate* isolate,
   return isolate->AddMessageListener(callback);
 }
 
+void v8__Isolate__AddNearHeapLimitCallback(v8::Isolate* isolate,
+                                           v8::NearHeapLimitCallback callback,
+                                           void* data) {
+  isolate.AddNearHeapLimitCallback(callback, data);
+}
+
+void v8__Isolate__RemoveNearHeapLimitCallback(
+    v8::Isolate* isolate, v8::NearHeapLimitCallback callback,
+    size_t heap_limit) {
+  isolate.RemoveNearHeapLimitCallback(callback, heap_limit);
+}
+
 const v8::Value* v8__Isolate__ThrowException(v8::Isolate* isolate,
                                              const v8::Value& exception) {
   return local_to_ptr(isolate->ThrowException(ptr_to_local(&exception)));
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -220,7 +220,7 @@ void v8__ResourceConstraints__ConfigureDefaultsFromHeapSize(
     v8::ResourceConstraints* constraints, size_t initial_heap_size_in_bytes,
     size_t maximum_heap_size_in_bytes) {
   constraints->ConfigureDefaultsFromHeapSize(initial_heap_size_in_bytes,
-                                             maximum_heap_size_in_bytes)
+                                             maximum_heap_size_in_bytes);
 }
 
 void v8__HandleScope__CONSTRUCT(uninit_t<v8::HandleScope>* buf,
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -181,13 +181,13 @@ bool v8__Isolate__AddMessageListener(v8::Isolate* isolate,
 void v8__Isolate__AddNearHeapLimitCallback(v8::Isolate* isolate,
                                            v8::NearHeapLimitCallback callback,
                                            void* data) {
-  isolate.AddNearHeapLimitCallback(callback, data);
+  isolate->AddNearHeapLimitCallback(callback, data);
 }
 
 void v8__Isolate__RemoveNearHeapLimitCallback(
     v8::Isolate* isolate, v8::NearHeapLimitCallback callback,
     size_t heap_limit) {
-  isolate.RemoveNearHeapLimitCallback(callback, heap_limit);
+  isolate->RemoveNearHeapLimitCallback(callback, heap_limit);
 }
 
 const v8::Value* v8__Isolate__ThrowException(v8::Isolate* isolate,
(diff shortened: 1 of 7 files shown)


src/binding.cc
```diff
@@ -178,6 +178,18 @@ bool v8__Isolate__AddMessageListener(v8::Isolate* isolate,
...
+void v8__Isolate__AddNearHeapLimitCallback(v8::Isolate* isolate,
+                                           v8::NearHeapLimitCallback callback,
+                                           void* data) {
+  isolate.AddNearHeapLimitCallback(callback, data);
+}
+
+void v8__Isolate__RemoveNearHeapLimitCallback(
+    v8::Isolate* isolate, v8::NearHeapLimitCallback callback,
+    size_t heap_limit) {
+  isolate.RemoveNearHeapLimitCallback(callback, heap_limit);
```
piscisaureus, line 190: Does this work? I must be very behind on my c++ skills...

mraerino: nope 😄 
CI caught it pretty fast

mraerino:
ok, when trying this out in `deno_core` i'm getting a linker error...

bartlomieju:
> ok, when trying this out in `deno_core` i'm getting a linker error...
> 
> ```
>   = note: Undefined symbols for architecture x86_64:
>             "_v8__ResourceConstraints__ConfigureDefaultsFromHeapSize", referenced from:
>                 rusty_v8::isolate_create_params::raw::ResourceConstraints::configure_defaults_from_heap_size::h0bff31ae2991bf2e in librusty_v8-22814f38384a97cd.rlib(rusty_v8-22814f38384a97cd.3vpkah4o8p3w6byd.rcgu.o)
>           ld: symbol(s) not found for architecture x86_64
>           clang: error: linker command failed with exit code 1 (use -v to see invocation)
> ```

@mraerino `cargo clean && cargo build` fixes usually fixes those errors

piscisaureus:
> ok, when trying this out in `deno_core` i'm getting a linker error...

Yeah that's a known issue. Once the build script has either downloaded or built libdeno.{a|lib} it will not automatically rebuild/re-download it when there are updates. Patches that fix this are always welcome :)

src/isolate.rs
```diff
@@ -345,6 +361,40 @@ impl Isolate {
...
+  /// Add a callback to invoke in case the heap size is close to the heap limit.
+  /// If multiple callbacks are added, only the most recently added callback is
+  /// invoked.
+  ///
+  /// # Safety
+  ///
+  /// Don't pass a null-pointer for the callback!
+  /// Make sure the callback only casts `data` to the type the `data` argument
+  /// here.
+  pub unsafe fn add_near_heap_limit_callback(
```
piscisaureus, line 373: I don't think this needs to be unsafe.

src/isolate.rs
```diff
@@ -345,6 +361,40 @@ impl Isolate {
     }
   }
 
+  /// Add a callback to invoke in case the heap size is close to the heap limit.
+  /// If multiple callbacks are added, only the most recently added callback is
+  /// invoked.
+  ///
+  /// # Safety
+  ///
+  /// Don't pass a null-pointer for the callback!
```
piscisaureus, line 370: Does it really accept a null ptr?

src/isolate_create_params.rs
```diff
@@ -114,6 +114,14 @@ impl CreateParams {
     self
   }
 
+  pub fn heap_limits(mut self, min: usize, max: usize) -> Self {
```
piscisaureus, line 117: We need some docs here. The prose by ConfigureDefaultsFromHeapSize in v8.h would be ok if the parameter names are made to match.

src/isolate.rs
mraerino: yeah, the caller might have to do some casting but i'm thinking of making a generics-based safe impl in Deno for this

mraerino:
i'm still getting a linker error when running `cargo test -p deno_core` in Deno.

I have this change in my `core/Cargo.toml`:
```diff
-rusty_v8 = "0.7.0"
+rusty_v8 = { path = "../../rusty_v8" }
```
This is how i'm using it: https://github.com/denoland/deno/pull/6914

I did `cargo clean` in both projects and deno builds fine, but tests fail with the linker problem. Any clues?

piscisaureus:
The implementation of this "feature" looks good to me now. The only thing that's missing is a test.

src/isolate.rs
piscisaureus: Generics based safe is a bit tricky. You'd have to pass an owned pointer like `Box<T>` or `Rc<T>`. This thing would need to be dropped when the isolate is disposed or the heap limit callback is removed.

mraerino: that's the idea

piscisaureus:
Are you sure you had V8_FROM_SOURCE set when you built it?

mraerino:
> Are you sure you had V8_FROM_SOURCE set when you built it?

somehow missed that concept. using it now, thanks!

tests/test_api.rs
```diff
@@ -10,6 +10,7 @@ use std::sync::Mutex;
 
 use rusty_v8 as v8;
 // TODO(piscisaureus): Ideally there would be no need to import this trait.
+use std::ffi::c_void;
```
bnoordhuis, line 13:
```suggestion
use std::ffi::c_void;
// TODO(piscisaureus): Ideally there would be no need to import this trait.
```
Otherwise the comment no longer makes sense.

tests/test_api.rs
```diff
@@ -3313,3 +3314,47 @@ fn module_snapshot() {
     }
   }
 }
+
+struct TestHeapLimitState {
+  near_limit_callback_called: bool,
```
bnoordhuis, line 3319: Consider making this an integral type so you can count the calls.

mraerino:
tests added. could you review again?

PR #429, opened 2020-08-04
Allow creating snapshot based on other snapshot

mraerino:
Introduces the ability to initialize the isolate that the `SnapshotCreator` creates with an optional snapshot blob, just as it's possible on a regular isolate.

This is a small breaking change. If you prefer it to be not breaking, I can introduce a second constructor instead.

### DIFF
--- a/src/snapshot.rs
+++ b/src/snapshot.rs
@@ -86,9 +86,9 @@ impl SnapshotCreator {
   ) -> Self {
     let mut snapshot_creator: MaybeUninit<Self> = MaybeUninit::uninit();
     let external_references_ptr =
-      external_references.map_or_else(|| std::ptr::null(), |er| er.as_ptr());
+      external_references.map_or_else(std::ptr::null, |er| er.as_ptr());
     let startup_data_ptr = startup_data
-      .map_or_else(|| std::ptr::null_mut(), |sd| Box::into_raw(Box::new(sd)));
+      .map_or_else(std::ptr::null_mut, |sd| Box::into_raw(Box::new(sd)));
     unsafe {
       v8__SnapshotCreator__CONSTRUCT(
         &mut snapshot_creator,
--- a/src/snapshot.rs
+++ b/src/snapshot.rs
@@ -17,6 +17,7 @@ extern "C" {
   fn v8__SnapshotCreator__CONSTRUCT(
     buf: *mut MaybeUninit<SnapshotCreator>,
     external_references: *const intptr_t,
+    startup_data: *mut StartupData,
   );
   fn v8__SnapshotCreator__DESTRUCT(this: *mut SnapshotCreator);
   fn v8__SnapshotCreator__GetIsolate(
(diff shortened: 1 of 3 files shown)


piscisaureus:
Breaking the API is okay. It'll happen many times in the foreseeable future.

piscisaureus:
I agree - we need this.

I think in the long term we should use CreateParams for both Isolate and SnapshotCreator. But it's good to follow the C++ api for now.

src/snapshot.rs
```diff
@@ -79,17 +80,20 @@ pub struct SnapshotCreator([usize; 1]);
 impl SnapshotCreator {
   /// Create and enter an isolate, and set it up for serialization.
   /// The isolate is created from scratch.
-  pub fn new(external_references: Option<&'static ExternalReferences>) -> Self {
+  pub fn new(
+    external_references: Option<&'static ExternalReferences>,
+    startup_data: Option<StartupData>,
```
piscisaureus, line 85: Downside of this setup is that the caller must own the snapshot (it can't be `&'static StartupData` or `Rc<StartupData>`), and therefore sharing the snapshot blob with another Isolate or SnapshotCreator. It is the direct opposite of `external_references` which takes *only* a reference and not an owned value.

mraerino:
so, i totally missed that a reference would be enough here.

still it needs to be `mut` according to the C headers.
do you think there would be a way of working around that fact?

mraerino:
> I think in the long term we should use CreateParams for both Isolate and SnapshotCreator. But it's good to follow the C++ api for now.

the bad part is that a snapshot creator will not allow you to set anything except those two params exposed now.
since the snapshot creator will initialize the isolate, you cannot set heap limits, for example 😞

src/snapshot.rs
```diff
@@ -79,17 +80,20 @@ pub struct SnapshotCreator([usize; 1]);
...
+      external_references.map_or_else(std::ptr::null, |er| er.as_ptr());
+    let startup_data_ptr = startup_data
+      .map_or_else(std::ptr::null_mut, |sd| Box::into_raw(Box::new(sd)));
     unsafe {
       v8__SnapshotCreator__CONSTRUCT(
         &mut snapshot_creator,
         external_references_ptr,
+        startup_data_ptr,
       );
       snapshot_creator.assume_init()
```
piscisaureus, line 98: It is not safe to return a stack-allocated SnapshotCreator. It causes the SnapshotCreator to potentially move -- which is incompatible with the C++ class definition. I think it would be best to bind `new` and `delete` in C++ rather than the constructor/destructor. Alternatively you could construct it into a box.

mraerino: wait, how does this have anything to do with my changes? I haven't touched the way the SnapshotCreator is constructed?

piscisaureus: Actually I take that back - SnapshotCreator is just a wrapper around a pointer to `struct SnapshotCreatorData`.

piscisaureus:
@mraerino I started working on it (#431) but now I got a little stuck and I have other things to do first.
I'll try if I can get this to work later today.

piscisaureus:
@mraerino 

> so, i totally missed that a reference would be enough here.

A reference would need to be valid for the 'static lifetime tho. 
 
> still it needs to be `mut` according to the C headers.
> do you think there would be a way of working around that fact?

I agree that this looks like a mistake on V8's end. I would be okay with using `const_cast<>` in binding.cc.

I was planning on landing this before the release today but unfortunately this patch is not ready for prime time.
The issue is that the only way to obtain a `StartupData`  value is by creating a snapshot. You can't load one from disk or bake it into the binary - the same issue we had to work out for `CreateParams` in the past. 

So the type of `snapshot` will have to be generic, but this brings about some other problems because it is wrapped in an option here - we could change the constructor signature to:

```rust
pub fn new(
    external_references: Option<&'static ExternalReferences>,
    snapshot_blob: Option<impl Allocated<[u8]>>,
  ) -> Self {
  // ...
}
```

That works, but it's quite inconvenient when you don't want to use a snapshot because you'll have to specify the type of the snapshot that you are *not* loading:

```rust
let snapshot_creator = SnapshotCreator::new(None, Option::<StartupData>::None);
```

mraerino:
cool, i see. thanks for expanding on it 🙌 

i'm ok with this not landing right now. good to get the ball rolling though!

ry:
@mraerino We'd be happy to have this feature, but we need a signed CLA to consider changes. Closing because stale.

Feel free to contact me on discord!

PR #434, opened 2020-08-08
Downgrade V8 to 8.6.334

ry:
to exclude https://chromium-review.googlesource.com/c/v8/v8/+/2280100 (0c837e8342dff02294d34d1866c9fcdbd89ffc9f)

experiencing hangs due to this CL in Deno when trying to upgrade to rusty_v8 0.8.0
https://github.com/denoland/deno/pull/6980

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Rusty V8 Binding
 
-V8 Version: 8.6.334, 2020-08-08
+V8 Version: 8.6.334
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 # Rusty V8 Binding
 
-V8 Version: 8.6.337, 2020-08-05
+V8 Version: 8.6.334, 2020-08-08
```
piscisaureus, line 3: Let's remove this date. It's unclear what it means. I always put the date of the v8 release there vs whatever today's date is.

ry: fine - can we do that next time?

PR #417, opened 2020-07-05
WIP: make all callbacks safe

piscisaureus:

PR #435, opened 2020-08-12
Allow configuring ModifyCodeGenerationFromStrings callback

NeoLegends:
This allows modifying whether JS is allowed to dynamically execute JS via `eval` or the `Function` constructor. This is useful in sandboxed environments when executing untrusted code.

I hope the way I defined the callback is correct. Please see
- https://v8docs.nodesource.com/node-14.1/d5/dda/classv8_1_1_isolate.html#a01ef69d6fd2d0eb43edf7245a0f31199
- https://v8docs.nodesource.com/node-14.1/d2/dc3/namespacev8.html#a4502d34e1886b964761899c245138d66

for more information on the v8 definition of the types.

### DIFF
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -75,6 +75,18 @@ pub type NearHeapLimitCallback = extern "C" fn(
   initial_heap_limit: usize,
 ) -> usize;
 
+#[repr(C)]
+pub struct ModifyCodeGenerationFromStringsResult<'s> {
+  pub codegen_allowed: bool,
+  pub modified_source: Option<Local<'s, String>>,
+}
+
+pub type ModifyCodeGenerationFromStringsCallback<'s> =
+  extern "C" fn(
+    ctx: Local<'s, Context>,
+    code: Local<'s, Value>,
+  ) -> ModifyCodeGenerationFromStringsResult<'s>;
+
 extern "C" {
   fn v8__Isolate__New(params: *const raw::CreateParams) -> *mut Isolate;
   fn v8__Isolate__Dispose(this: *mut Isolate);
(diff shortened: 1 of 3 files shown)


src/isolate.rs
```diff
@@ -75,6 +75,18 @@ pub type NearHeapLimitCallback = extern "C" fn(
...
+  pub codegen_allowed: bool,
+  pub modified_source: Option<Local<'s, String>>,
+}
+
+pub type ModifyCodeGenerationFromStringsCallback<'s> =
+  extern "C" fn(
+    ctx: Local<'s, Context>,
+    code: Local<'s, Value>,
+  ) -> ModifyCodeGenerationFromStringsResult<'s>;
+
```
NeoLegends, line 89: I wonder if this is the correct definition of the callback. When exposing the callback from Deno (which is the next step up from this PR) I get linker errors.

image

Any hints on how to do this properly? v8 passes two Locals and in the struct expects a MaybeLocal back (which I assume is equivalent to `Option<Local<'_, T>>`.

bartlomieju: @NeoLegends if this errors pops up when you tried to build Deno you'll need to run `cargo clean` before `cargo build`

NeoLegends: Added a test that tries to pass a callback, and: it fails in rusty_v8, too (and not only when building Deno). `cargo clean` doesn't help there.

piscisaureus:
@NeoLegends You can use the callback definition from this patch: https://github.com/denoland/rusty_v8/pull/417/files#diff-ac9fae4f88db0c64421fbaab6136222cR5777-R5893

NeoLegends:
Oh, wow. Kudos to your Rust-foo.

Is this supposed to wait until #417 lands then?

piscisaureus:
> Oh, wow. Kudos to your Rust-foo.
> 
> Do you want this to wait until #417 lands then?

Sorry for the lack of communication. It might take a while before #417 lands. When it eventually does I'll land this PR too.

If you need it done sooner, you could copy the relevant type definitions from #417 and add them to this patch.

NeoLegends:
Sounds good!

bnoordhuis:
@NeoLegends Do you want to move forward with this PR? It seems useful to have and I wouldn't wait for #417 to land.

NeoLegends:
Eventually I want to, yes. We currently have things to focus on where having the callback isn't required, but I hope to be able to come back to this rather soon.

ry:
@NeoLegends I'm going to close this because it's old and stale. Happy to have this feature - if you're still willing please open a new PR.

PR #440, opened 2020-08-23
Add Isolate::get_heap_statistics()

bnoordhuis:


dallen66:
One problem I had with my old pull request  #383 it did work across threads. I think IsolateHandle struct  should have a get_heap_statistics() to make them work across threads.

bnoordhuis:
@dallen66 Do you mean call `isolate.get_heap_statistics()` on thread A for an isolate that's running on thread B? I think you can make that work with `isolate.request_interrupt()`:
```rust
isolate.request_interrupt(callback, std::ptr::null_mut());

extern "C" fn callback(isolate: &mut v8::Isolate, _: *mut std::ffi::c_void) {
  let mut stats = v8::HeapStatistics::default();
  isolate.get_heap_statistics(&mut stats);
  handle(&stats);
}

dallen66:
> @dallen66 Do you mean call `isolate.get_heap_statistics()` on thread A for an isolate that's running on thread B? I think you can make that work with `isolate.request_interrupt()`:
> 
> ```rust
> isolate.request_interrupt(callback, std::ptr::null_mut());
> 
> extern "C" fn callback(isolate: &mut v8::Isolate, _: *mut std::ffi::c_void) {
>   let mut stats = v8::HeapStatistics::default();
>   isolate.get_heap_statistics(&mut stats);
>   handle(&stats);
> }
> ```

Thank I will try that.

PR #442, opened 2020-08-26
Add ValueSerializer, ValueDeserializer bindings for structured clone in deno

inteon:
Hey!

I started working on this issue, meanwhile I'm trying to familiarize myself with the project.
I'm not an experienced Rust developer; so please review my code!
The code I uploaded compiles, but maybe there are still some non-best-practices in the code.

Could you give some feedback before I continue working on this issue?
The current progress:
- [x] ValueSerializer bindings
- [x] ValueSerializer::Delegate bindings
- [x] ValueDeserializer bindings
- [x] ValueDeserializer::Delegate bindings
- [x] Added Tests
- [x] Added default `Not-Implemented` exceptions for `write_host_object`, `get_shared_array_buffer_id`, ...
- [x] Added `throw_data_clone_error` custom callback

### DIFF
--- a/src/support.h
+++ b/src/support.h
@@ -111,6 +111,16 @@ inline static MaybeBool maybe_to_maybe_bool(v8::Maybe<bool> maybe) {
   }
 }
 
+inline static v8::Maybe<bool> maybe_bool_to_maybe(MaybeBool maybe) {
+  switch(maybe) {
+    case MaybeBool::JustTrue:
+    case MaybeBool::JustFalse:
+      return v8::Just<bool>(maybe == MaybeBool::JustTrue);
+    default:
+      return v8::Nothing<bool>();
+  }
+}
+
 template <class T>
 inline static T* local_to_ptr(v8::Local<T> local) {
   return *local;
--- a/src/value_deserializer.rs
+++ b/src/value_deserializer.rs
@@ -1,0 +1,422 @@
+use crate::ArrayBuffer;
+use crate::Context;
+use crate::Exception;
+use crate::HandleScope;
+use crate::Isolate;
+use crate::Local;
+use crate::Object;
+use crate::SharedArrayBuffer;
+use crate::String;
+use crate::Value;
+use crate::WasmModuleObject;
+
+use crate::support::CxxVTable;
+use crate::support::FieldOffset;
+use crate::support::MaybeBool;
+
+use std::ffi::c_void;
+use std::mem::MaybeUninit;
+use std::pin::Pin;
+
+// Must be == sizeof(v8::ValueDeserializer::Delegate),
+// see v8__ValueDeserializer__Delegate__CONSTRUCT().
+#[repr(C)]
+pub struct CxxValueDeserializerDelegate {
+  _cxx_vtable: CxxVTable,
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn v8__ValueDeserializer__Delegate__ReadHostObject(
+  this: &mut CxxValueDeserializerDelegate,
+  _isolate: *mut Isolate,
+) -> *const Object {
+  let value_deserializer_heap = ValueDeserializerHeap::dispatch_mut(this);
+  let scope =
+    &mut crate::scope::CallbackScope::new(value_deserializer_heap.context);
+  let value_deserializer_impl =
+    value_deserializer_heap.value_deserializer_impl.as_mut();
+  match value_deserializer_impl
+    .read_host_object(scope, &value_deserializer_heap.cxx_value_deserializer)
+  {
+    None => std::ptr::null(),
+    Some(x) => x.as_non_null().as_ptr(),
+  }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn v8__ValueDeserializer__Delegate__GetSharedArrayBufferFromId(
+  this: &mut CxxValueDeserializerDelegate,
+  _isolate: *mut Isolate,
+  transfer_id: u32,
+) -> *const SharedArrayBuffer {
+  let value_deserializer_heap = ValueDeserializerHeap::dispatch_mut(this);
+  let scope =
+    &mut crate::scope::CallbackScope::new(value_deserializer_heap.context);
+  let value_deserializer_impl =
+    value_deserializer_heap.value_deserializer_impl.as_mut();
+  match value_deserializer_impl
+    .get_shared_array_buffer_from_id(scope, transfer_id)
+  {
+    None => std::ptr::null(),
+    Some(x) => x.as_non_null().as_ptr(),
+  }
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn v8__ValueDeserializer__Delegate__GetWasmModuleFromId(
+  this: &mut CxxValueDeserializerDelegate,
+  _isolate: *mut Isolate,
+  clone_id: u32,
+) -> *const WasmModuleObject {
+  let value_deserializer_heap = ValueDeserializerHeap::dispatch_mut(this);
+  let scope =
+    &mut crate::scope::CallbackScope::new(value_deserializer_heap.context);
+  let value_deserializer_impl =
+    value_deserializer_heap.value_deserializer_impl.as_mut();
+  match value_deserializer_impl.get_wasm_module_from_id(scope, clone_id) {
+    None => std::ptr::null(),
+    Some(x) => x.as_non_null().as_ptr(),
+  }
+}
+
+extern "C" {
+  fn v8__ValueDeserializer__Delegate__CONSTRUCT(
+    buf: *mut MaybeUninit<CxxValueDeserializerDelegate>,
+  );
+}
+
+// Must be == sizeof(v8::ValueDeserializer),
+// see v8__ValueDeserializer__CONSTRUCT().
+#[repr(C)]
+pub struct CxxValueDeserializer {
+  _cxx_vtable: CxxVTable,
+}
+
+extern "C" {
+  fn v8__ValueDeserializer__CONSTRUCT(
+    buf: *mut MaybeUninit<CxxValueDeserializer>,
+    isolate: *mut Isolate,
+    data: *const u8,
+    size: usize,
+    delegate: *mut CxxValueDeserializerDelegate,
+  );
+
+  fn v8__ValueDeserializer__DESTRUCT(this: *mut CxxValueDeserializer);
+
+  fn v8__ValueDeserializer__TransferArrayBuffer(
+    this: *mut CxxValueDeserializer,
+    transfer_id: u32,
+    array_buffer: Local<ArrayBuffer>,
+  );
+
+  fn v8__ValueDeserializer__SetSupportsLegacyWireFormat(
+    this: *mut CxxValueDeserializer,
+    supports_legacy_wire_format: bool,
+  );
+
+  fn v8__ValueDeserializer__ReadHeader(
+    this: *mut CxxValueDeserializer,
+    context: Local<Context>,
+  ) -> MaybeBool;
+
+  fn v8__ValueDeserializer__ReadValue(
+    this: *mut CxxValueDeserializer,
+    context: Local<Context>,
+  ) -> *const Value;
+
+  fn v8__ValueDeserializer__ReadUint32(
+    this: *mut CxxValueDeserializer,
+    value: *mut u32,
+  ) -> bool;
+
+  fn v8__ValueDeserializer__ReadUint64(
+    this: *mut CxxValueDeserializer,
+    value: *mut u64,
+  ) -> bool;
+
+  fn v8__ValueDeserializer__ReadDouble(
+    this: *mut CxxValueDeserializer,
+    value: *mut f64,
+  ) -> bool;
+
+  fn v8__ValueDeserializer__ReadRawBytes(
+    this: *mut CxxValueDeserializer,
+    length: usize,
+    data: *mut *const c_void,
+  ) -> bool;
+}
+
+/// The ValueDeserializerImpl trait allows for
+/// custom callback functions used by v8.
+pub trait ValueDeserializerImpl {
+  #[allow(unused_variables)]
+  fn read_host_object<'s>(
+    &mut self,
+    scope: &mut HandleScope<'s>,
+    value_deserializer: &dyn ValueDeserializerHelper,
+  ) -> Option<Local<'s, Object>> {
+    let msg =
+      String::new(scope, "Deno deserializer: read_host_object not implemented")
+        .unwrap();
+    let exc = Exception::error(scope, msg);
+    scope.throw_exception(exc);
+    None
+  }
+
+  #[allow(unused_variables)]
+  fn get_shared_array_buffer_from_id<'s>(
+    &mut self,
+    scope: &mut HandleScope<'s>,
+    transfer_id: u32,
+  ) -> Option<Local<'s, SharedArrayBuffer>> {
+    let msg = String::new(
+      scope,
+      "Deno deserializer: get_shared_array_buffer_from_id not implemented",
+    )
+    .unwrap();
+    let exc = Exception::error(scope, msg);
+    scope.throw_exception(exc);
+    None
+  }
+
+  #[allow(unused_variables)]
+  fn get_wasm_module_from_id<'s>(
+    &mut self,
+    scope: &mut HandleScope<'s>,
+    clone_id: u32,
+  ) -> Option<Local<'s, WasmModuleObject>> {
+    let msg = String::new(
+      scope,
+      "Deno deserializer: get_wasm_module_from_id not implemented",
+    )
+    .unwrap();
+    let exc = Exception::error(scope, msg);
+    scope.throw_exception(exc);
+    None
+  }
+}
+
+/// The ValueDeserializerHeap object contains all objects related to a
+/// deserializer. This object has to be pinned to the heap because of the Cpp
+/// pointers that have to remain valid. Moving this object would result in the
+/// Cpp pointer to the delegate to become invalid and thus causing the delegate
+/// callback to fail. Additionally the deserializer and implementation are also
+/// pinned in memory because these have to be accessable from within the
+/// delegate callback methods.
+pub struct ValueDeserializerHeap<'a, 's> {
+  value_deserializer_impl: Box<dyn ValueDeserializerImpl + 'a>,
+  cxx_value_deserializer: CxxValueDeserializer,
+  cxx_value_deserializer_delegate: CxxValueDeserializerDelegate,
+  context: Local<'s, Context>,
+}
+
+impl<'a, 's> ValueDeserializerHeap<'a, 's> {
+  fn get_cxx_value_deserializer_delegate_offset(
+  ) -> FieldOffset<CxxValueDeserializerDelegate> {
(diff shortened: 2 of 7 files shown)


piscisaureus:
First of all - structure clone support is high on our wish list, so your effort is very much appreciated.

Your rust code so far looks great. I'll comment in-line if I have any suggestions.

One thing though - it looks like we will also need bindings for `ValueSerializer::Delegate` and `ValueDeserializer::Delegate`. I expect that adding bindings for these Delegate classes will turn
out to be a lot more difficult than what you've done so far - I hope you're up for that challenge too!

src/value_serializer.rs
```diff
@@ -0,0 +1,109 @@
...
+
+  pub fn write_uint64(&mut self, value: u64) {
+    unsafe { v8__ValueSerializer__WriteUint64(self, value) };
+  }
+
+  pub fn write_double(&mut self, value: f64) {
+    unsafe { v8__ValueSerializer__WriteDouble(self, value) };
+  }
+
+  pub unsafe fn write_raw_bytes(
```
piscisaureus, line 102: We try to adapt function/method signatures in such way that they can be used without `unsafe` and do not use raw pointers.

I think you could make `write_raw_bytes()` safe by doing something along these lines:

```rust
pub fn write_raw_bytes(
    &mut self,
    source: &[u8]
  ) {
    unsafe { v8__ValueSerializer__WriteRawBytes(self, source.as_ptr() as *const _, source.len()); }
  }
}
```

src/value_serializer.rs
```diff
@@ -0,0 +1,109 @@
...
+
+use crate::support::MaybeBool;
+
+use crate::data::ArrayBuffer;
+use crate::data::Value;
+
+use std::ffi::c_void;
+use std::mem::MaybeUninit;
+
+/// Instances of this class can be passed
```
piscisaureus, line 14: 1. I'm not sure exactly what you mean here - maybe "instances of this class can be passed by value"?
2. Triple slash comments are used for doc comments. That means that if we were to publish the crate like this
    the words "Instances of this class..." would be visible [here](https://docs.rs/rusty_v8/0.8.1/rusty_v8/#functions).
    For code-only comments, use `//` instead, like you've done on the line below.

src/value_serializer.rs
```diff
@@ -0,0 +1,109 @@
...
+  fn v8__ValueSerializer__WriteDouble(this: *mut ValueSerializer, value: f64);
+  fn v8__ValueSerializer__WriteRawBytes(
+    this: *mut ValueSerializer,
+    source: *const c_void,
+    length: usize,
+  );
+}
+
+impl ValueSerializer {
+  pub fn new<'s>(scope: &mut HandleScope<'s>) -> Self {
```
piscisaureus, line 49: As this function does not return a Local or anything else that borrows from `scope`, I think you don't need a named lifetime here. Instead:

```rust
pub fn new(scope: &mut HandleScope<'_>) -> Self
```

...should work.

src/value_serializer.rs
```diff
@@ -0,0 +1,109 @@
...
+use crate::support::MaybeBool;
+
+use crate::data::ArrayBuffer;
+use crate::data::Value;
+
+use std::ffi::c_void;
+use std::mem::MaybeUninit;
+
+/// Instances of this class can be passed
+// Must be >= sizeof(v8::ValueSerializer), see v8__ValueSerializer__CONSTRUCT().
```
piscisaureus, line 15: Generally I maintain that the size must match exactly.

inteon:
Thanks for the feedback, I fixed the issues and continued programming the missing parts.
For the `Value(Des|S)erializer::Delegate` code I based the solution on the `task.rs` solution.
If you have any new feedback; feel free to share. I think adding some tests will also help fixing any remaining bugs/ problems.

src/value_deserializer.rs
```diff
@@ -0,0 +1,302 @@
...
+    unsafe { v8__ValueDeserializer__ReadUint64(self, value) }
+  }
+
+  pub fn read_double(&mut self, value: &mut f64) -> bool {
+    unsafe { v8__ValueDeserializer__ReadDouble(self, value) }
+  }
+
+  pub fn read_raw_bytes(&mut self, length: usize, data: *mut &[u8]) -> bool {
+    unsafe {
+      v8__ValueDeserializer__ReadRawBytes(self, length, data as *mut *const _)
```
piscisaureus, line 299: * `&[u8]` is a fat pointer a.k.a. a (start, length) tuple. The V8 API will write only update the start address of that pointer and leave the length unchanged.
* V8 updates `data` to point at a buffer that is owned by the ValueDeserializer; however this is not properly expressed in the method signature.

My suggestion would be to try this (preserving zero-copy behavior):
 
```rust
  pub fn read_raw_bytes<'a>(&'a mut self, length: usize) -> Option<&'a [u8]> {
    let mut data: *const c_void = std::ptr::null_mut();
    let ok =
      unsafe { v8__ValueDeserializer__ReadRawBytes(self, length, &mut data) };
    if ok {
      assert!(!data.is_null());
      unsafe { Some(std::slice::from_raw_parts(data as *const u8, length)) }
    } else {
      None
    }
  }
```

piscisaureus:
> For the `Value(Des|S)erializer::Delegate` code I based the solution on the `task.rs` solution.

That is indeed what I would have suggested; I'm impressed you were able to figure it out since that solution is not so straightforward or elegant.

src/support.h
```diff
@@ -124,6 +134,23 @@ inline static const v8::MaybeLocal<T> ptr_to_maybe_local(const T* ptr) {
...
+inline static const T* maybe_to_ptr(v8::Maybe<T> maybe) {
+  if (maybe.IsNothing()) {
+    return nullptr;
+  }
+  return maybe.FromJust();
+}
+
+template <class T>
+inline static const v8::Maybe<T> ptr_to_maybe(const T* ptr) {
+  static_assert(sizeof(v8::Maybe<T>) == sizeof(T*), "");
```
piscisaureus, line 147: Does this pass? The non-specialized definition of `Maybe` is:

```c++
template <class T>
class Maybe {
  bool has_value_;
  T value_; 
}
```

... and I could not find a specialization for pointers.

I think you could achieve what you need by using the `Maybe<T> Maybe::Just(const T& t);` method. 
Like `return Maybe::Just(*ptr);`.

piscisaureus: Additionally, the assertion is incorrect. It should be:

```
static_assert(sizeof(v8::Maybe<T*>) == sizeof(T*), "");
```

inteon: Actually, shouldn't Maybe be mapped to a Rust Option instead of a simple pointer?

src/value_deserializer.rs
inteon: seems to be working, I removed the explicit lifetime due to clippy complaining

inteon:
Thanks for the support; I think we are making some great progress! I was wondering with regards to cleaning up the rust_v8 interface:
- should I always use "scope: &mut HandleScope<'_>" instead of "*mut Isolate" as function argument?
- Is it ok to return "*const Object", or should this be reconverted to "Local<Object>"?
-

piscisaureus:
> * should I always use `scope: &mut HandleScope<'_>` instead of `*mut Isolate` as function argument?

`*mut Isolate` cannot be used safely, so indeed you'd need to use another type. Which type exactly depends on the situation. In this situation, since you need to return Local handles, `&mut HandleScope<'_>` is indeed the type of choice. 

> * Is it ok to return `*const Object`, or should this be reconverted to `Local<Object>`?

It should be converted to either `Local<Object>` if it's guaranteed to be non-null, or `Option<Local<Object>>` if it might be null. You can use `HandleScope::to_local()` for this; there are many examples in the code base, e.g. [here](https://github.com/denoland/rusty_v8/blob/fa8f636397822ba7dbb823d25dc637bf77f8b1ce/src/json.rs#L22-L30).

inteon:
> > * should I always use `scope: &mut HandleScope<'_>` instead of `*mut Isolate` as function argument?
> 
> `*mut Isolate` cannot be used safely, so indeed you'd need to use another type. Which type exactly depends on the situation. In this situation, since you need to return Local handles, `&mut HandleScope<'_>` is indeed the type of choice.
> 
> > * Is it ok to return `*const Object`, or should this be reconverted to `Local<Object>`?
> 
> It should be converted to either `Local<Object>` if it's guaranteed to be non-null, or `Option<Local<Object>>` if it might be null. You can use `HandleScope::to_local()` for this; there are many examples in the code base, e.g. [here](https://github.com/denoland/rusty_v8/blob/fa8f636397822ba7dbb823d25dc637bf77f8b1ce/src/json.rs#L22-L30).

:thinking: It seems like the `read_host_object` function in the `ValueDeserializerDelegate` has the `Isolate` pointer as one of its arguments. How can we now convert this pointer to a `HandleScope`? Additionally we need this `HandleScope` so the return value can be converted from `*const Object` to `Local<Object>` (like you just described in your comment).

I would guess the following code would not work as this creates a new `HandleScope`?
```rust
```
Instead, is there a way to retrieve the instantiated `HandleScope` objects or should an alternative solution be used?

src/support.h
piscisaureus: > I think `MaybeLocal` can be directly mapped to `Option<Box>` because both can be represented as a nullable pointer. But for Maybe to be mapped to Option; an additional boolean is required I think.

What matters here is that the actual layout of the C++ and Rust types need match, and they also need to be *guaranteed* to match (otherwise clang / rust will complain that the type is not FFI-safe).

Indeed, `Option<NonNull<T>>` and `Option<Box<T>>` have the same layout as a raw pointer (`*mut T`) and rust *does* explicitly guarantee this -- but this is the exception, most non-primitive types are not considered FFI-safe.

`Maybe<T>`  probably needs this rust definition:
```
[repr(C)]
struct Maybe<T> {
  has_value: bool;
  value: T
}
```

The above applies to the FFI interface. If we can avoid it we would not expose `Maybe<T>` to the end user, but rather use the applicable idiomatic rust type, e.g.  `Option<&T>` in the end-user interface.

piscisaureus:
> Instead, is there a way to retrieve the instantiated `HandleScope` objects or should an alternative solution be used?

It is possible, although there are some safety constraints to it. You'd use something like `let scope = unsafe { &mut crate::scope::CallbackScope::new(&mut *isolate);`
You can see it in action [here](https://github.com/denoland/rusty_v8/blob/37b6f8fa0c5bcac9b3051e658cd5ac8cf9cbbbb9/src/callback.rs#L6574-L6581).

inteon:
Ok thx, I cleaned up the interface. Currently, the `free_buffer_memory` is the only function that still uses a pointer.

src/value_serializer.rs
```diff
@@ -0,0 +1,374 @@
...
+    &mut self,
+    transfer_id: u32,
+    array_buffer: Local<ArrayBuffer>,
+  ) {
+    unsafe {
+      v8__ValueSerializer__TransferArrayBuffer(self, transfer_id, array_buffer)
+    };
+  }
+
+  pub fn release(&mut self) -> &[u8] {
```
piscisaureus, line 331: This function unfortunately leaks memory. You can go in two directions with this:

1. The ValueSerializer remains the owner of the buffer even after `release()` is called. It frees the buffer in it's drop handler.
2. The ValueSerializer is consumed by this function, and returns a buffer that is owned by the caller. What return type is appropriate depends on how the buffer is allocated.

inteon: I think we have to go for option 2; as the v8 ValueSerializer won't be usable after calling this function. (https://github.com/v8/v8/blob/master/include/v8.h#L2373)

piscisaureus: > I think we have to go for option 2; as the v8 ValueSerializer won't be usable after calling this function. (https://github.com/v8/v8/blob/master/include/v8.h#L2373)

I think option 2 is indeed the more elegant choice. The unsolved question however is which "owned" type we should return. It would depend on the concrete implementation that the user provides for ReallocateBufferMemory/FreeBufferMemory. I see another complication too - the `ReallocateBufferMemory` method does not receive the "old" length of the buffer, so it might be difficult to provide an implementation that uses an appropriate rust type (`Vec<u8>` or `Box<[u8]>` would be the most likely candidates). 

I'll need to think about this a bit longer.

inteon: I thought the `size` parameter given to the `ReallocateBufferMemory` function indicates the length of the old buffer?  
In the current implementation I mapped the `ReallocateBufferMemory` function to a rust function with the following interface:

```rust
fn reallocate_buffer_memory(
    &mut self,
    old_buffer: Option<&[u8]>,
  ) -> Option<&mut [u8]>;
```

So I'm currently using (optional) slices. I think `Vec<u8>` won't work because a vector also keeps track of the "used" memory space additional to the allocated space.
Maybe `Box<[u8]>` should be used instead of `&mut [u8]`? From what I can find in the rust docs, passing a box actually means transferring ownership, whereas a reference means your lending the memory.
I created a simple rust playground to better explain how we can convert pointer + size to a boxed slice: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=5dfa5af617124fe545a45af3d46c8770

piscisaureus: > Maybe `Box<[u8]>` should be used instead of `&mut [u8]`? From what I can find in the rust docs, passing a box actually means transferring ownership,

Correct! `Box<[u8]>` would actually be great.
Unfortunately that, too, is a fat pointer.

src/value_serializer.rs
```diff
@@ -116,6 +119,8 @@ pub unsafe extern "C" fn v8__ValueSerializer__Delegate__BASE__FreeBufferMemory(
   this: &mut ValueSerializerDelegate,
   buffer: *mut c_void,
 ) {
+  let buffer =
+    Box::from_raw(std::ptr::slice_from_raw_parts_mut(buffer as *mut u8, 1));
```
inteon, line 123: I think this works; but it is definitely not ideal. V8 does not give us the old buffer size, because normally calling free(ptr) in c does not require this. Although for Rust this seems to be required.

src/value_serializer.rs
inteon: I changed the array slices to boxed slices.

inteon:
I added an extra test that serializes a javascript object and deserializes it in another "Isolate".
It seems like `SharedArrayBuffer` does not work without creating a `SerializerDelegate` with working `GetSharedArrayBufferId` function. I'll try to add this to the tests too.
Feel free to add extra tests that might be needed for completeness.

It looks like we're getting close to finishing this WIP 🎉!

src/value_serializer.rs
piscisaureus: Actually, I just realized that if every ValueSerializerDelegate will only be allocating one buffer (and I'm pretty sure of that, but will verify), it's really not such a big deal. The implementer just store the previously allocated length -- we could even decide to make rusty_v8 store it on `struct ValueSerializerDelegateBase`.

src/value_serializer.rs
```diff
@@ -0,0 +1,382 @@
...
+    source: *const c_void,
+    length: usize,
+  );
+}
+
+impl ValueSerializer {
+  pub fn new(scope: &mut HandleScope<'_>) -> Self {
+    let isolate: *mut Isolate = scope.get_isolate_ptr();
+    let mut value_serializer: MaybeUninit<Self> = MaybeUninit::uninit();
+    unsafe {
```
piscisaureus, line 296: With the current implementation where release() returns a `Box<[u8]>` we will have to provide a default `ValueSerializedDelegate`, because otherwise the default behavior will be to allocate memory with an unknown allocator (malloc()/free()) which does not necessarily produce a valid `Box<[u8]>`.

src/value_serializer.rs
```diff
@@ -0,0 +1,382 @@
...
+    &mut self,
+    transfer_id: u32,
+    array_buffer: Local<ArrayBuffer>,
+  ) {
+    unsafe {
+      v8__ValueSerializer__TransferArrayBuffer(self, transfer_id, array_buffer)
+    };
+  }
+
+  pub fn release(&mut self) -> Box<[u8]> {
```
piscisaureus, line 336: For 'consume' semantics, the first parameter should be `self` not `&mut self`.

src/value_serializer.rs
```diff
@@ -0,0 +1,382 @@
...
+  fn v8__ValueSerializer__Delegate__BASE__CONSTRUCT(
+    buf: *mut MaybeUninit<ValueSerializerDelegate>,
+  );
+}
+
+pub trait ValueSerializerDelegateImpl {
+  fn base(&self) -> &ValueSerializerDelegateBase;
+  fn base_mut(&mut self) -> &mut ValueSerializerDelegateBase;
+
+  fn throw_data_clone_error(&mut self, message: Local<String>);
```
piscisaureus, line 137: Unfortunately this function will be a bit useless because there's literally nothing the user can do with the message parameter (because we are not getting a `HandleScope` to work with).

piscisaureus:
> It seems like `SharedArrayBuffer` does not work without creating a `SerializerDelegate` with working `GetSharedArrayBufferId` function. I'll try to add this to the tests too.

I think that is the most important TODO item at this point. 

As you might have noticed I also have some concerns w.r.t the current "strategy" for allocating and resizing serializer buffers - I have some ideas but I need to sit down and work on the code myself for a bit (otherwise I'll end up proposing "solutions" that turn out not to work after all). 

Other than that this looks pretty good and almost done.

src/value_serializer.rs
inteon: Yeah, this would be a good solution. I now also understand what you told me earlier [the ReallocateBufferMemory method does not receive the "old" length of the buffer](https://github.com/denoland/rusty_v8/pull/442#discussion_r479556457). I thought the `size` parameter did supply that information, but this parameter actually indicates the **wanted** buffer size instead of the current one. So the current implementation contains a bug.

inteon: > Actually, I just realized that if every ValueSerializerDelegate will only be allocating one buffer (and I'm pretty sure of that, but will verify), it's really not such a big deal. The implementer just store the previously allocated length -- we could even decide to make rusty_v8 store it on `struct ValueSerializerDelegateBase`.

The only problem I see is reusing delegates for multiple ValueSerializers.

inteon: Ideally the buffer-length should be added to the `ValueSerializer` struct instead, a serializer always has only 1 buffer.
Although this requires the `ValueSerializerDelegate` to be able to access the `ValueSerializer` struct.

piscisaureus:
Other TODOs:

* `throw_data_clone_error()` needs `HandleScope`.
* `write_host_object` needs a reference to `Serializer`.

inteon:
UPDATE:
TODOs:
[] `throw_data_clone_error()` needs `HandleScope`.
[x] `write_host_object` needs a reference to `Serializer`.
[] default implementations for trait methods (??) -> `HandleScope` needed

piscisaureus:
Apologies for the delay, I’ll land this soon.

bnoordhuis:
@inteon I rebased your branch to fix the merge conflicts and I took the liberty of addressing some style issues. It passes for me locally and I'll merge it when the CI is green. Nice work, thanks!

inteon:
@bnoordhuis ok thx

bnoordhuis (approved):

PR #450, opened 2020-08-31
Fix flaky tests caused by relaxed load in C++ 'shared_ptr::use_count()'

piscisaureus:


src/support.rs
```diff
@@ -205,16 +211,23 @@ impl<T: Shared> Drop for SharedPtrBase<T> {
...
-  pub fn is_null(&self) -> bool {
-    <T as Shared>::get(&self.0).is_null()
+  /// Asserts that the number of references to the shared inner value is equal
+  /// to the `expected` count.
+  ///
+  /// This function relies on the C++ method `std::shared_ptr::use_count()`,
+  /// which usually performs a relaxed load. This function will repeatedly call
+  /// `use_count()` until it returns the expected value, for up to one second.
+  /// Therefore it should probably not be used in performance critical code.
+  #[track_caller]
```
ry, line 224: What's the purpose of `#[track_caller]`?

src/support.rs
```diff
@@ -297,6 +324,40 @@ impl<T: Shared> Borrow<T> for SharedRef<T> {
...
+      let start = Instant::now();
+      loop {
+        yield_now();
+        actual = T::use_count(shared_ptr);
+        if actual == expected {
+          break true;
+        } else if start.elapsed() > RETRY_TIMEOUT {
+          break false;
+        }
+      }
```
ry, line 347: I don't follow what's happening here? Is this a one second hot loop? Why? (Maybe add a comment?)

tests/test_api.rs
```diff
@@ -459,44 +459,44 @@ fn backing_store_segfault() {
...
   let shared_bs = {
-    assert_eq!(1, v8::SharedRef::use_count(&array_buffer_allocator));
+    array_buffer_allocator.assert_use_count_eq(1);
     let params = v8::Isolate::create_params()
       .array_buffer_allocator(array_buffer_allocator.clone());
-    assert_eq!(2, v8::SharedRef::use_count(&array_buffer_allocator));
+    array_buffer_allocator.assert_use_count_eq(2);
     let isolate = &mut v8::Isolate::new(params);
-    assert_eq!(2, v8::SharedRef::use_count(&array_buffer_allocator));
+    array_buffer_allocator.assert_use_count_eq(2);
```
ry, line 467: Seems odd this is exposed to the public API.

ry:
Are we even using shared_ptr in a multithreaded way? I don't really understand what is happening here.

tests/test_api.rs
piscisaureus: But the alternative is to have the unreliable `use_count()` method in the public API. 
In that light this seems preferable.

piscisaureus:
> Are we even using shared_ptr in a multithreaded way? I don't really understand what is happening here.

Theoretically the thread could be preempted, or a page fault could happen, in between dropping the `shared_ptr` reference and the call to `use_count()`. However when reproducing this locally it happens often enough (about 1 in 100 times) that I don't believe this is happening.

My alternative theory is that this is what SPECTRE mitigation looks like.

@bnoordhuis I wonder if you have any theories...

src/support.rs
piscisaureus: If the assert fails (generally: when a panic happens), it won't show the panic originating here (in support.rs) but rather at the call site (test_api.rs in this case).

PS: the top hit on google would have provided the answer.

PR #457, opened 2020-09-05
Add V8::set_flags_from_string()

bnoordhuis:


src/V8.rs
```diff
@@ -79,6 +80,13 @@ pub fn set_flags_from_command_line(args: Vec<String>) -> Vec<String> {
     .collect()
 }
 
+/// Sets V8 flags from a string.
+pub fn set_flags_from_string(flags: &str) {
+  unsafe {
+    v8__V8__SetFlagsFromString(flags.as_ptr(), flags.len());
```
bnoordhuis, line 86: Comment from `set_flags_from_command_line()` on line 45-46:

> /// TODO: Check whether this is safe to do after globally initializing v8.

The answer is 'no' - it works but it's a race condition (and unsafe for flags that take a string argument.)

ry (approved):
LGTM - thanks

PR #458, opened 2020-09-06
CI: "on" events trigger refactoring

ipriver:
It's a minor refactoring of the ci.yml. There're no specific branch options for "on" events trigger so that it can be combined into a list for simplicity. (e.g. [denoland/deno ci](https://github.com/denoland/deno/blob/master/.github/workflows/ci.yml#L3))


ry (approved):
thx

PR #459, opened 2020-09-07
Add support for LowMemoryNotification

NeoLegends:
This PR adds the bindings for LowMemoryNotification which has been left out of #440.


piscisaureus:
Can you add a smoke tests?

NeoLegends:
Like this?

piscisaureus:
Thanks!
I'll land it but without d840f7b (we stick to v8 names when possible, and here it's possible).

NeoLegends:
Alright! Reverted that commit.

piscisaureus (approved):

NeoLegends:
🎉 thanks for the quick reaction

NeoLegends:
When do you expect to make the next release?

PR #460, opened 2020-09-07
Add V8::set_entropy_source()

bnoordhuis:


src/V8.rs
```diff
@@ -12,13 +12,18 @@ use crate::support::UniqueRef;
...
   fn v8__V8__GetVersion() -> *const c_char;
   fn v8__V8__InitializePlatform(platform: *mut Platform);
   fn v8__V8__Initialize();
   fn v8__V8__Dispose() -> bool;
   fn v8__V8__ShutdownPlatform();
 }
 
+/// EntropySource is used as a callback function when v8 needs a source
+/// of entropy.
+pub type EntropySource = fn(&mut [u8]) -> bool;
```
bnoordhuis, line 25: The V8 docs don't mention it but returning `false` makes V8 fall back to system entropy (`/dev/urandom`, etc.)

piscisaureus:
@bnoordhuis 

Can I convince you to copy+paste the callback wrapper from https://github.com/denoland/rusty_v8/blob/37b6f8fa0c5bcac9b3051e658cd5ac8cf9cbbbb9/src/callback.rs#L7919-L8002 ?

By using function item types rather than function pointers we can statically wrap our function in a C ABI wrapper.
This avoids the need to add this global `Mutex<Option< yada yada >>`

bnoordhuis:
@piscisaureus I did what you requested but it makes it a little less ergonomic. Before:
```rust
V8::set_entropy_source(|buf| frob(buf));
```
Now:
```rust
let cb = |buf: &mut [u8]| frob(buf);
V8::set_entropy_source(cb.into());
```

piscisaureus:
But you could also:

```rust
fn set_entropy_source(cb: impl_into_entropy_source!()) -> {
  unsafe { set_it_now(cb.into()) }
}
```

piscisaureus:
^-- there you go

piscisaureus (approved):

bnoordhuis:
@piscisaureus I came to something similar as your commit 6bcf5af but without the macro and with `FnMut` instead of `FnOnce` (because it's invoked repeatedly, so... that's appropriate, I think?)

piscisaureus:
> so... that's appropriate, I think?)

Not really though... Fn/FnMut borrow their closure context so they can't outlive the context where they're created in.

In this case it doesn't make a difference because F is zero sized and even implements Copy. 
So I picked FnOnce because it's the more generic trait (a supertrait of Fn and FnMul).

piscisaureus:
... but let me be concrete:
Fn is fine, but FnMut is really out of place.

piscisaureus:
Also since the macro is useless without the fix that you pushed out, let me suggest to remove the macro entirely.

Those macros are supposed to relieve you from the need to write out complicated Fn trait types over and over again. But EntropySource is so simple that there's really no merit to that.

bnoordhuis:
>  So I picked FnOnce because it's the more generic trait (a supertrait of Fn and FnMut).

Just so we're on the same page, your argument is that because it ultimately decays to a function pointer instead a closure, it doesn't really matter if it's `FnOnce` or `Fn`? The documentation for [`FnOnce`](https://doc.rust-lang.org/std/ops/trait.FnOnce.html) say this:

> Instances of FnOnce can be called, but might not be callable multiple times. Because of this, if the only thing known about a type is that it implements FnOnce, it can only be called once.

Can I suggest we change it to `Fn` or `FnMut` to indicate it's called multiple times? Happy to open a PR! `FnMut` still seems appropriate to me because of this:

> If you need to call the parameter repeatedly, use FnMut as a bound; if you also need it to not mutate state, use Fn.

"Mutate state" is something anything but the most trivial entropy generator will do.

piscisaureus:
@bnoordhuis 

There's a lot to say about this, but essentially the Fn/FnMut/FnOnce collection sit on a "flexibility scale" with regard to what the function may do with it's own closure-captured state (which is like an implicit parameter).

FnOnce affords most flexibility to the implementer, but the caller is most restricted by its API contract - it has to give up exclusive ownership of the closure state when it calls the function, therefore it can do that only once.

FnMut is somewhere in between.

`Fn` is the most restrictive to the implementer of these three, but the caller can duplicate & share a reference to the closure freely, the only limitation coming from the duration (lifetime) of borrowed access to it's captured state. 

But you could imagine even more restrictive (from the implementer's perspective) contracts than `Fn`.  One would be `FnOnce + Copy`. The implementer would only be able to capture Copy types (which includes immutable &references since they are Copy). On the caller's end, not just **a reference** to the closure state can be shared/copied/called without restriction, the same is now true for the closure state **itself**. Thus the caller can never be restricted by any lifetime associated with a borrowed reference. 

And then there's the ultimately-restrictive kind: A function that can't use any captured closure state.

Concretely, `EntropyCallback` is of the latter kind: it can't capture any state.

So you could argue, since it is even more restrictive than `Fn`, using `Fn` is the more appropriate choice.

But there's a more mechanical perspective too: 

I generally consider it better style that, when you design an API, you give as much flexibility away as you can afford. E.g. in c++ you would mark methods and parameters as `const` when possible - something that "benefits" the caller of the function/method. In Rust, methods operate on `&self` (rather than `&mut self` or `self`) when they can. And you'd take `FnOnce` for a callback parameter when you can.  

Since this `set_entropy_source` "works" if you give it an `FnOnce` so why not put that in the signature?
(The sting of course is the other trait the callback is required to implement for `set_entropy_source()` to accept it, which is `UnitType`, implying also `Copy`.)

Ergo, `FnOnce` is a defensible choice, and `Fn` is probably best, and `FnMut` would (incorrectly) communicate a specific situation in the no-man's land between `Fn` and `FnOnce`.

> "Mutate state" is something anything but the most trivial entropy generator will do.

Nice try but that "mutate state" applies only to captured closure state - not any other function parameters nor anything appropriately wrapped in cells, nor thread_local/global variables. See for example this example of an `Fn` that mutates it's parameter:

```rust
fn increment(val: &mut i32) {
    *val += 1;
}
fn indirect_increment(val: &mut i32, incrementor: impl Fn(&mut i32) -> ()) {
    incrementor(val)
}
fn main() {
    let mut a = 1;
    increment(&mut a);
    println!("{}", a); // 2
    indirect_increment(&mut a, increment);
    println!("{}", a); // 3
    indirect_increment(&mut a, |a| *a += 1);
    println!("{}", a); // 4
}
```

piscisaureus:
Some other things:

* I want to rename `UnitType` to `NoCapture` or something. Originally written to assert that a type is zero-sized but it's not used for anything else other the new name indicates.
* Let's write a RFC for `FnClone` :D. I think it'd be really nice to have, working with closures usually means cloning a lot.

ijroth:
You two are really cute. ;)

PR #461, opened 2020-09-07
Simplify 'Isolate::get/set_slot()' and use optimized hasher

piscisaureus:


src/support.rs
```diff
@@ -506,6 +509,46 @@ impl<F> FieldOffset<F> {
   }
 }
 
+/// A special hasher that is optimized for hashing `std::any::TypeId` values.
+/// It can't be used for anything else.
+#[derive(Clone, Default)]
+pub(crate) struct TypeIdHasher {
+  state: Option<u64>,
+}
```
ry, line 517: Seems unnecessary but sure...

ry (approved):
LGTM

src/support.rs
piscisaureus: Not super necessary I suppose but more like a nice thing.

PR #464, opened 2020-09-09
ci: switch to Github Actions XL

piscisaureus:


somombo:
@piscisaureus what is "Github Actions XL"? 
I can neither find any useful information about this on [Google](https://www.google.com/search?q=Github+Actions+XL) nor on Github's [documentation about hosted runners](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources). Is this a feature of github actions that maybe was removed some time after this PR was merged?

lucacasonato:
No, they are just beefier hosted runners that GitHub gave us access to. They are not available for the general public.

somombo:
okay makes sense. However, I want to point out (if not known) that  `ubuntu-latest-xl` is  ubuntu v18.04, `ubuntu-latest` is now at v20.04.. so on forks some of the dependencies (like gcc-5) are broken because they are nolonger available on the public repo in that version of ubuntu. I was testing this because I have been trying to see if I can work on adding support for armv7 and was running the ci in the meantime on my fork, while i figure things out.

Thanks for the info!

CreepyGnome:
Out of curiosity did they give you the specs (CPUs, Mem, etc) for the `ubuntu-latest-xl` by chance?

lucacasonato:
8 cores, not sure about memory.

PR #475, opened 2020-09-19
Add examples

3c1u:
I am new to `rusty_v8`, trying to figure out how `rusty_v8` works, and re-wrote [examples of V8](https://github.com/v8/v8/tree/master/samples) in Rust.

- [x] hello-world.cc
- [x] shell.cc (no message loop pumping)
- [x] process.cc (some todos)

I hope this will help anyone who starts to learn how to use `rusty_v8`.


examples/process.rs
```diff
@@ -0,0 +1,373 @@
...
+    // Local scope for temporary handles.
+    let scope = &mut self.context_scope;
+
+    let result = v8::Object::new(scope);
+    let external = v8::External::new(
+      scope,
+      Box::leak(request) as *mut Box<dyn HttpRequest> as *mut c_void,
+    );
+
+    let key = v8::String::new(scope, "__private__").unwrap().into();
```
bnoordhuis, line 274: Oh, we don't have a binding for `v8::Object::SetInternalField()` yet, do we? And no way to instantiate a `v8::Private` either. I've opened #477 for the former.

examples/shell.rs
```diff
@@ -0,0 +1,236 @@
...
+      "--shell" => {
+        // Enables the shell.
+        *run_shell = true;
+      }
+      "-f" => {
+        // Ignore any -f flags for compatibility with the other stand-
+        // alone JavaScript engines.
+      }
+      "-e" => {
+        // Exectute script.
```
bnoordhuis, line 89:
```suggestion
        // Execute script.
```

examples/shell.rs
```diff
@@ -0,0 +1,236 @@
...
+    v8::Boolean::new(&mut handle_scope, false),
+    v8::Boolean::new(&mut handle_scope, false),
+  );
+
+  let script = if let Some(script) =
+    v8::Script::compile(&mut handle_scope, script, Some(&origin))
+  {
+    script
+  } else {
+    drop(handle_scope);
```
bnoordhuis, line 141: Just curious, is there a reason to explicitly drop the HandleScope?

bnoordhuis (commented):
Left some suggestions but in general looks really nice!

examples/process.rs
```diff
@@ -0,0 +1,373 @@
...
+  }
+}
+
+/// An http request processor that is scriptable using JavaScript.
+struct JsHttpRequestProcessor<'s, 'i> {
+  context: v8::Local<'s, v8::Context>,
+  context_scope: v8::ContextScope<'i, v8::HandleScope<'s>>,
+  process_fn: Option<v8::Local<'s, v8::Function>>,
+  _request_template: Option<v8::Local<'s, v8::ObjectTemplate>>,
+  _map_template: Option<v8::Local<'s, v8::ObjectTemplate>>,
```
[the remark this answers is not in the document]
3c1u: This is due to Rust restrictions, as we cannot borrow `isolate` as mutable twice.

examples/shell.rs
3c1u: This is to "trick" Rust borrow checker. `handle_scope` will not be dropped until the program reaches the end of the function.
```
cannot borrow `try_catch` as immutable because it is also borrowed as mutable
immutable borrow occurs hererustc(E0502)
shell.rs(121, 47): mutable borrow occurs here
shell.rs(166, 1): mutable borrow might be used here, when `handle_scope` is dropped and runs the `Drop` code for type `rusty_v8::HandleScope`
```

examples/process.rs
3c1u: Thanks!

examples/process.rs
```diff
@@ -0,0 +1,373 @@
...
+) {
+  let message = arg
+    .get(0)
+    .to_string(scope)
+    .unwrap()
+    .to_rust_string_lossy(scope);
+
+  println!("Logged: {}", message);
+
+  retval.set(v8::undefined(scope).into());
```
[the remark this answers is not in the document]
3c1u: Fixed in e96491b.

examples/hello_world.rs
```diff
@@ -0,0 +1,70 @@
+//! Rust version of `v8/samples/hello-world.cc`.
+
+#![deny(clippy::all)]
+#![deny(clippy::pedantic)]
```
ry, line 4: Please remove these

examples/process.rs
ry: #477 landed, if you want to update.

ry (commented):
Thanks for the patch - looks good.

examples/shell.rs
ry: Hm - I wonder if this can be reworked to avoid this drop? Can you use the `scope` argument of the `execute_string()` function instead of creating a new handle scope?

3c1u: Avoided `drop` by using `TryCatch` as a scope.

examples/hello_world.rs
3c1u: Removed!

examples/process.rs
3c1u: Now it uses `ObjectTemplate` and internal fields.

examples/count-hosts.js
```diff
@@ -0,0 +1,42 @@
+// Copyright 2008 the V8 project authors. All rights reserved.
```
ry, line 1: Remove this file - it isn't used anywhere?

Cargo.toml
```diff
@@ -60,3 +60,9 @@ which = "4.0.2"
 
 [dev-dependencies]
 trybuild = "1.0.33"
+
+[[example]]
+name = "hello_world"
+
+[[example]]
+name = "shell"
```
ry, line 68: The process.rs example isn't there.

examples/count-hosts.js
3c1u: This is included in `v8/samples`, and this intended to be used with `process.cc` (`process.rs`), which will not run without any script.

Cargo.toml
3c1u: Oh, I completely forgot! Thanks.

3c1u:
Rebased. This will pass the test.

ry (approved):
LGTM - thanks @3c1u ! Sorry it took so long to land this

PR #335, opened 2020-04-04
Weakable Globals & FFI Object Wrapping

Protryon:
This PR:
* Fixes an apparent non-deterministic test that depended on a script id
* Adds `Isolate::request_garbage_collection_for_testing` for GC-related tests
* Adds internal field support to `Object`s through:
    * `Object::internal_field_count`
    * `Object::get_internal_field` -- unsafe
    * `Object::set_internal_field` -- unsafe
    * `Object::get_internal_field_ptr` -- unsafe
    * `Object::set_internal_field_ptr` -- unsafe
    * `ObjectTemplate::internal_field_count`
    * `ObjectTemplate::set_internal_field_count`
    * The non-ptr variants could be made safe in the future by fully managing internal fields, but inside V8 the non-ptr and ptr variants overlap, so these have been marked unsafe.
* Adds a subset of weak support to `Global`s through:
    * `Global::set_weak`
    * `Global::is_weak`
    * `Global::clear_weak`

This PR partially addresses #304.

Move from #327 to change source branch.

PR #477, opened 2020-09-21
add Object + ObjectTemplate internal field support

bnoordhuis:
The rusty_v8 API deviates slightly from the V8 C++ API because the
latter is definitely unsound when you pass in out-of-range indexes.


3c1u:
Overlaps with https://github.com/denoland/rusty_v8/pull/335?

ry (approved):
LGTM - thanks Ben

PR #484, opened 2020-10-03
add some binding, and add some new feature to external

Icemic:
- Getting instance template object and prototype template object from its function template object
- Getting holder object from FunctionCallbackArguments
- Making external works with std::rc::Rc (which enhance set_internal_field)

All new methods have tested in my machine.

And btw, you may find I did some same thing with what @bnoordhuis has done in the same week 🤣 (about internal fields), so I've remove my implement after merging master.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1274,6 +1274,11 @@ const v8::Object* v8__FunctionCallbackInfo__This(
   return local_to_ptr(self.This());
 }
 
+const v8::Object* v8__FunctionCallbackInfo__Holder(
+    const v8::FunctionCallbackInfo<v8::Value>& self) {
+  return local_to_ptr(self.Holder());
+}
+
 int v8__FunctionCallbackInfo__Length(
     const v8::FunctionCallbackInfo<v8::Value>& self) {
   return self.Length();
@@ -1578,6 +1583,11 @@ v8::Value* v8__PropertyCallbackInfo__GetReturnValue(
 const v8::Object* v8__PropertyCallbackInfo__This(
     const v8::PropertyCallbackInfo<v8::Value>& self) {
   return local_to_ptr(self.This());
+}
+
+const v8::Object* v8__PropertyCallbackInfo__Holder(
+    const v8::PropertyCallbackInfo<v8::Value>& self) {
+  return local_to_ptr(self.Holder());
 }
 
 const v8::Proxy* v8__Proxy__New(const v8::Context& context,
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -547,6 +547,11 @@ const v8::Boolean* v8__Value__ToBoolean(const v8::Value& self,
   return local_to_ptr(self.ToBoolean(isolate));
 }
 
+const v8::External* v8__Value__ToExternal(v8::Value& self,
+                                        v8::Isolate* isolate) {
+  return v8::External::Cast(&self);
+}
+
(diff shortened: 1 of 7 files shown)


src/binding.cc
```diff
@@ -547,6 +547,11 @@ const v8::Boolean* v8__Value__ToBoolean(const v8::Value& self,
   return local_to_ptr(self.ToBoolean(isolate));
 }
 
+const v8::External* v8__Value__ToExternal(v8::Value& self,
+                                        v8::Isolate* isolate) {
```
bnoordhuis, line 551:
```suggestion
const v8::External* v8__Value__ToExternal(v8::Value& self,
                                          v8::Isolate* isolate) {
```

src/binding.cc
```diff
@@ -1255,6 +1260,14 @@ void v8__FunctionTemplate__SetClassName(const v8::FunctionTemplate& self,
   ptr_to_local(&self)->SetClassName(ptr_to_local(&name));
 }
 
+const v8::ObjectTemplate* v8__FunctionTemplate__PrototypeTemplate(const v8::FunctionTemplate& self) {
```
bnoordhuis, line 1263: Try to keep lines <= 80 columns.

src/external.rs
```diff
@@ -26,7 +27,25 @@ impl External {
...
+  pub fn new_rc<'s, T>(
+    scope: &mut HandleScope<'s, ()>,
+    value: Rc<T>,
+  ) -> Local<'s, Self> {
+    let value_ptr = Rc::into_raw(value) as *mut c_void;
+    unsafe {
+      scope.cast_local(|sd| v8__External__New(sd.get_isolate_ptr(), value_ptr))
+    }
+    .unwrap()
+  }
```
bnoordhuis, line 39: rusty_v8 is a fairly basic, low-level binding to the V8 C++ API so this kind of abstraction arguably has no place here.

src/function.rs
```diff
@@ -137,6 +143,13 @@ impl<'s> FunctionCallbackArguments<'s> {
     }
   }
 
+  /// Returns the instance of function template.
```
bnoordhuis, line 146: Can you copy the doc comment from v8/include/v8.h? Likewise around line 234.

src/value.rs
```diff
@@ -540,6 +545,16 @@ impl Value {
...
 
+  pub fn to_external<'s>(
+    &self,
+    scope: &mut HandleScope<'s, ()>,
+  ) -> Local<'s, External> {
+    unsafe {
+      scope.cast_local(|sd| v8__Value__ToExternal(self, sd.get_isolate_ptr()))
+    }
+    .unwrap()
+  }
```
bnoordhuis, line 556: I can see why you put this here but, contrary to the other methods, `v8::External::Cast()` is distinctly unsafe - it's basically a `reinterpret_cast` or `std::mem::transmute()`.

A better way to deal with converting from `Value` to `External` is through `.try_into()` (which might already work, not sure.)

bnoordhuis (commented):
Thanks for the PR! Can you please also add tests?

src/value.rs
Icemic: You're right. That's kind of overdesign and unnecessary. Removed.

ry:
Hi @Icemic  - I'm going to close this PR because it's stale and old. It's a shame we didn't land it. Glancing at it I think it generally looks good but there were some unaddressed review comments. If you'd like to create a new PR with these changes rebased, I'd be happy to review it.

PR #488, opened 2020-10-06
Make RUSTY_V8_MIRROR understand file:// URLs

### LINKED ISSUE #470: rusty_v8 long build times
Using _stable-x86_64-pc-windows-msvc_ toolchain, rusty_v8 v0.9.1.

Cargo commands which need to build the `rusty_v8` crate take very long for me (up to 15min). This happens for example at each `cargo publish --dry-run` invocation.

I am not sure what exactly is being done, but the progressbar is often stuck at
```
Building [=====================================================>   ] 74/78: rusty_v8(build)
```
for a very long time. I first thought it was re-downloading the .lib again and again, but it keeps building when I cut the internet. Maybe linking the library?

Since this doesn't occur on `cargo build` or `cargo test` (except for the initial call), there seems to be a mechanism to cache the built artifacts. I tried using system-wide `sccache`, which didn't change anything. I'm not specifying any flags to build from source.

Is there something obvious I'm missing? How could I speed up development with the rusty_v8 crate?

bnoordhuis:
Having rusty_v8 redownload libv8.a every time you blow away target/ is
a good way to blow through your monthly bandwith allotment quickly, to
say nothing about the slowdown of the build.

Fixes #470

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,11 +39,41 @@ Binaries builds are turned on by default: `cargo build` will initiate a download
 from github to get the static lib. To disable this build using the
 `V8_FROM_SOURCE` environmental variable.
 
-If you prefer another location for downloading binaries,
-use the`RUSTY_V8_MIRROR` environmental variable.
-
 When making changes to rusty_v8 itself, it should be tested by build from
 source. The CI always builds from source.
+
+## The `RUSTY_V8_MIRROR` environment variable
+
+Tells the build script where to get binary builds from. Understands
+`http://` and `https://` URLs, and file paths. The default is
+https://github.com/denoland/rusty_v8/releases/download.
+
+File-based mirrors are good for using cached downloads. First, point
+the environment variable to a suitable location:
+
+    # you might want to add this to your .bashrc
+    $ export RUSTY_V8_MIRROR=$HOME/.cache/rusty_v8
+
+Then populate the cache:
+
+```bash
+#!/bin/bash
+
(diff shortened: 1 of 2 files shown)


ry (approved):
If this is working for you LGTM

README.md
```diff
@@ -39,12 +39,42 @@ Binaries builds are turned on by default: `cargo build` will initiate a download
...
+
+Tells the build script where to get binary builds from. Understands
+`http://`, `https://` and `file://` URLs, and defaults to
+https://github.com/denoland/rusty_v8/releases/download.
+
+`file://` mirrors are good for using cached downloads. First, point
+the environmental variable to a suitable location:
+
+    # you might want to add this to your .bashrc
+    $ export RUSTY_V8_MIRROR=$HOME/.cache/rusty_v8
```
TTtie, line 55: The value of `$HOME` is not a `file://` URL - which would make the `RUSTY_V8_MIRROR` an invalid endpoint - might want to prepend `file://` manually and modify the script that populates the cache to get rid of the `file://` protocol respectively.
```suggestion
    $ export RUSTY_V8_MIRROR=file://$HOME/.cache/rusty_v8
```

TTtie (changes requested):

README.md
bnoordhuis: Good catch. Thinking about it more, I think I'll make RUSTY_V8_MIRROR interpret it as a path unless it starts with `http:` or `https:`. Seems easier.

PR #496, opened 2020-10-12
Add Isolate::set_promise_hook()

bnoordhuis:
cc @bartlomieju

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -1734,8 +1734,8 @@ fn promise_hook() {
     _parent: v8::Local<v8::Value>,
   ) {
     let scope = &mut unsafe { v8::CallbackScope::new(promise) };
-    let scope = &mut v8::HandleScope::new(scope);
     let context = promise.creation_context(scope);
+    let scope = &mut v8::ContextScope::new(scope, context);
     let global = context.global(scope);
     let name = v8::String::new(scope, "hook").unwrap();
     let func = global.get(scope, name.into()).unwrap();
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -1726,6 +1726,57 @@ fn promise_reject_callback_no_value() {
   }
 }
 
+#[test]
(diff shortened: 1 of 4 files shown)


bartlomieju (approved):
LGTM

piscisaureus:
wait

tests/test_api.rs
```diff
@@ -1726,6 +1726,57 @@ fn promise_reject_callback_no_value() {
...
 
+#[test]
+fn promise_hook() {
+  extern "C" fn hook(
+    type_: v8::PromiseHookType,
+    promise: v8::Local<v8::Promise>,
+    _parent: v8::Local<v8::Value>,
+  ) {
+    let scope = &mut unsafe { v8::CallbackScope::new(promise) };
+    let scope = &mut v8::HandleScope::new(scope);
```
piscisaureus, line 1737: This is unnecessary. V8 creates a HandleScope for us already. Passing the CallbackScope created 1 line above will work anywhere a HandleScope is needed.

tests/test_api.rs
```diff
@@ -1726,6 +1726,57 @@ fn promise_reject_callback_no_value() {
...
+    let context = promise.creation_context(scope);
```
piscisaureus, line 1738: Since you are going to call a function that is bound to `context`, I think you should also enter said context.
Otherwise the function now suddenly "sees" a different global object which can cause weird behavior.
Vice versa it may not be undesirable to leak the innards of the current context to the creation context.
So I'd suggest to use a ContextScope here.

piscisaureus: Actually nevermind - this is a test... 🤦

bnoordhuis: I added a ContextScope although it doesn't really matter here as there's only one context. But it makes sense to be theoretically correct if the test suite also doubles as ersatz documentation.

Apropos the HandleScope, I see that pattern in several other tests.

PR #499, opened 2020-10-15
Add Global::map()

bnoordhuis:
Add a convenience method for unwrapping the `v8::Global` without a scope
object. This is the same trick that Node.js applies to speed up unboxing
strong persistent handles.

Refs: https://github.com/denoland/deno/issues/7969

<hr>

It's an idea I had while thinking about denoland/deno#7969. I'm not saying it's a good idea but it seems convenient to have.


ry:
Closing because stale.

PR #508, opened 2020-10-25
Add Function::new_instance

leops:
This allows object instances to be created from constructor functions. This is useful for instantiating ECMAScript Classes from native code since they cannot be called as functions and must use this `new_instance` path.


bnoordhuis (approved):
LGTM, thanks!

PR #510, opened 2020-10-27
Simplify Function::new_with_data()

bnoordhuis:
It's not necessary to have a separate C++ glue function, it can be
shared with Function::new().


ry (approved):
LGTM

PR #522, opened 2020-11-13
Fix runtime assert when hashing module object

bnoordhuis:
The blanket `std::hash::Hash` impl for instances of `v8::Data` invokes
`v8::internal::Object::GetHash()` but that crashes for `v8::Module`
objects. Use a custom impl that calls `v8::Module::get_identity_hash()`.

Fixes the following runtime assertion:

    # Fatal error in ../../../v8/src/objects/objects-inl.h, line 1043
    # Debug check failed: object.IsJSReceiver().

Refs: https://github.com/denoland/deno/pull/8354#discussion_r522157813

cc @bartlomieju @piscisaureus

### DIFF
--- a/src/data.rs
+++ b/src/data.rs
@@ -348,7 +348,6 @@ pub struct Module(Opaque);
 impl_deref! { Data for Module }
 impl_try_from! { Data for Module if d => d.is_module() }
 impl_eq! { for Module }
-impl_hash! { for Module }
 impl_partial_eq! { Data for Module use identity }
 impl_partial_eq! { Module for Module use identity }
 
--- a/src/module.rs
+++ b/src/module.rs
@@ -1,4 +1,6 @@
 use std::convert::TryInto;
+use std::hash::Hash;
+use std::hash::Hasher;
 use std::mem::MaybeUninit;
 use std::ptr::null;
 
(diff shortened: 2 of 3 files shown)


piscisaureus:
LGTM, although I think it would be appropriate to use DummyHasher rather than DefaulHasher just like we do for other object types.

piscisaureus:
Another thing - if a Module has been "upcast" to v8::Data then this still invokes the wrong hash function.

bartlomieju:
Original idea for my PR was not to use `get_identity_hash` but it seems this function is still called here. Won't this cause the same problems with "birthday paradox"?

bnoordhuis:
@piscisaureus I don't think that's an issue? There's no `impl Hash for Data`, only for its subtypes.

I don't understand your DummyHasher comment. That's not used anywhere in rusty_v8, AFAICT.

@bartlomieju That's inherent when using a hashing function because of the pigeon hole principle. I could mix in more bits, like the script id, but that won't make a huge difference.

piscisaureus:
> There's no impl Hash for Data, only for its subtypes.

I see, you're right. I had planned to implement that but never did. 

> I don't understand your DummyHasher comment. That's not used anywhere in rusty_v8, AFAICT.

Sorry, I was wrong about that. I meant `BuildTypeIdHasher` but that's to be used when constructing a `HashMap/HashSet`, rather than when implementing the `Hash` trait .

piscisaureus (approved):
🚀

PR #523, opened 2020-11-13
Remove mut from Example in lib.rs

danbev:
This commit removes `mut` from the `script` variable to avoid the following
warning if used:
```console
25 |   let mut script = v8::Script::compile(scope, code, None).unwrap();
   |       ----^^^^^^
   |       |
   |       help: remove this `mut`
   |
   = note: `#[warn(unused_mut)]` on by default

warning: 1 warning emitted
```


bnoordhuis (approved):
Thanks Daniel, LGTM.

PR #529, opened 2020-11-15
Make v8::ScriptOrigin support optional arguments

bnoordhuis:
The C++ API requires the resource name argument but the other eight (!)
arguments are optional. Mimic that by allowing `None`.

This commit adds a small helper to turn `Option<Local<T>>` into a raw
`T*` pointer. I'll move over more of the code base in a follow-up pull
request.

### DIFF
--- a/examples/shell.rs
+++ b/examples/shell.rs
@@ -111,14 +111,14 @@ fn execute_string(
   let script = v8::String::new(&mut scope, script).unwrap();
   let origin = v8::ScriptOrigin::new(
     v8::String::new(&mut scope, filename).unwrap().into(),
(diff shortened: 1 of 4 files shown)


bnoordhuis:
The run failed with a SIGSEGV... I'm inclined to believe that isn't caused by this PR. Restarted but saving the error message for posterity:
```
2020-11-15T18:43:32.1529832Z   process didn't exit successfully: `/home/runner/work/rusty_v8/rusty_v8/target/x86_64-unknown-linux-gnu/release/deps/test_api-40dee9a527ddf804` (signal: 11, SIGSEGV: invalid memory reference)
```
(no other error output)

PR #530, opened 2020-11-15
Add Value::to_rust_string_lossy()

bnoordhuis:
If it's convenient to have on String objects, then it's convenient to
have on all Value instances.


bartlomieju (approved):
This is super useful and not having it tripped me many times in `deno_core`, similar to lack of `Debug` implementations. LGTM

PR #532, opened 2020-11-17
Add dispose/shutdown_platform to hello_world

danbev:
This commit adds calls to `v8::V8::dispose` and `v8::V8::shutdown_platform`
to the `hello_world.rs` example.

The motivation for adding this is that it was not obvious to me that I
needed to have the Isolate in a separate scope. If there is not, the
platform will have been shutdown before the Isolate's `Drop` function
is called at the end of the function where the scope ends. `Drop` will in
turn call `Dispose` on the V8 Isolate, which calls `Isolate::Deinit`, and
`Deinit` calls `tracing_cpu_profilers.reset()` which will call
`~TracingCpuProfilerImpl` which has a call to `V8::GetCurrentPlatform()`
which will produce the following error:

```console
#
# Fatal error in ../../../v8/src/init/v8.cc, line 144
# Debug check failed: platform.
#
#
#
#FailureMessage Object: 0x7fff45b03700Illegal instruction (core dumped)
```

Having an example might safe others some time if they run into the same
issue.

### DIFF
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -6,60 +6,67 @@ fn main() {
   v8::V8::initialize_platform(platform);
   v8::V8::initialize();
 
-  // Create a new Isolate and make it the current one.
-  let isolate = &mut v8::Isolate::new(v8::CreateParams::default());
+  {
+    // Create a new Isolate and make it the current one.
+    let isolate = &mut v8::Isolate::new(v8::CreateParams::default());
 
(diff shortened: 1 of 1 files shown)


bnoordhuis:
I force-pushed your branch to restart the CI, it was failing on an infrastructural issue.

bnoordhuis (approved):

bnoordhuis:
Thanks Dan!

PR #536, opened 2020-11-19
Add set_flags_from_command_line_with_usage fn

danbev:
This commit suggests adding a usage string which can be passed to V8
when setting the command line flags. If the `--help` flag is specified
then the usage string will be printed before the V8 options.

The motivation for this would be to enable embedders of rusty_v8 to
pass their own usage string and have their usage printed before V8's
options.

For example:
```console
Usage: binaryname --startup-src=file

SSE3=1 SSSE3=1 SSE4_1=1 SSE4_2=1 SAHF=1 AVX=1 FMA3=1 BMI1=1 BMI2=1
The following syntax for options is accepted (both '-' and '--' are ok):
  --flag        (bool flags only)
  --no-flag     (bool flags only)
  --flag=value  (non-bool flags only, no spaces around '=')
  --flag value  (non-bool flags only)
  --            (captures all remaining args in JavaScript)

Options:
...
```

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -1000,6 +1000,15 @@ fn set_flags_from_command_line() {
 }
 
 #[test]
+fn set_flags_from_command_line_with_usage() {
+  let r = v8::V8::set_flags_from_command_line_with_usage(
+    vec!["binaryname".to_string(), "--help".to_string()],
+    Some("Usage: binaryname --startup-src=file\n\n"),
+  );
+  assert_eq!(r, vec!["binaryname".to_string()]);
+}
+
+#[test]
 fn inspector_string_view() {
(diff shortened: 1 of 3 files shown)


tests/test_api.rs
```diff
@@ -1004,6 +1004,15 @@ fn set_flags_from_command_line() {
   );
 }
 
+#[test]
+fn set_flags_from_command_line_with_usage() {
```
danbev, line 1008: This actually prints to stdout regardless if `--nocapture` is specified on the command line. If this feature is something that seems reasonable to add, then I'll see if I can work around that.

ry (approved):
Thanks @danbev - seems reasonable. Sorry for the slow review.

PR #540, opened 2020-11-20
Reword Source struct comment in script_compiler.rs

danbev:
This commit fixes what I think is a typo but I'm not sure.


ry (approved):
thx

PR #543, opened 2020-11-23
use lockfile to prevent linking of partially downloaded lib

mash-graz:


mash-graz:
this PR concerns the issue https://github.com/denoland/deno/issues/8449

bartlomieju (approved):
Sorry for slow review @mash-graz, discussed offline with @bnoordhuis and we decided that it's the best to go with your approach and monitor potential problems. 

That said, it might take some time to verify this approach as we need to release new version `rusty_v8`; upgrade it in `deno` and then monitor for CI failures.

mash-graz:
thanks for this encouraging message! :)

> That said, it might take some time to verify this approach as we need to release new version rusty_v8; upgrade it in deno and then monitor for CI failures.

yes -- it's definitely a tricky issue and it's very wise to carefully watch out for unwanted side effects.

just two remarks concerning this tests and necessary steps forward:

1. unfortunately the issue doesn't only affect `deno`. other projects using `rusty_v8`, like `jstime`, are fighting the same symptoms. 

2. compiling `deno` in a typical CI environment often works without this patch, because the virtual machines used for this purpose commonly provide only one single CPU core. in this case cargo behaves like using -j1 one a native multicore machine. the real troubles only happen, if you compile `deno` on such a multicore machine without limiting the concurrency of the build processes by an explicit `-j1`.

bartlomieju:
> unfortunately the issue doesn't only affect deno. other projects using rusty_v8, like jstime, are fighting the same symptoms.

Right, that's good, more projects = more confidence in the fix :) 

> compiling deno in a typical CI environment often works without this patch, because the virtual machines used for this purpose commonly provide only one single CPU core. in this case cargo behaves like using -j1 one a native multicore machine. the real troubles only happen, if you compile deno on such a multicore machine without limiting the concurrency of the build processes by an explicit -j1.

`deno` repo uses multicore linux runner so the problem should present itself there (hopefully not after this fix!)

Issue #545, opened 2020-11-30
Verify checksum of static lib in build script


bartlomieju:
As an alternative approach to fixing `truncated or malformed archive (offset to next archive member past the end of the archive after member /)` error (https://github.com/denoland/rusty_v8/pull/543, https://github.com/denoland/deno/issues/8449) we shall create checksum files in CI and later verify those checksums against downloaded static lib during build process.

mash-graz:
hmm -- @bnoordhuis has [already suggested](https://github.com/denoland/deno/issues/8449#issuecomment-731445262) this kind of workaround in an early answer to my reports about this defect in our current build system.

maybe he still hasn't changed his mind about the actual cause of this issue and still denies any explanation, which interpret the troubles as a consequence of inter-process race conditions in our build system. i don't know -- he simply didn't answer my additional analyses and suggested practical solutions resp. patch attempts until now..

i spend a lot of time last weekend to study this symptoms and their causes. it's a rather puzzling and hard to reproduce issue, like most errors caused by parallel running computing processes resp. unsynchronized data access.

i still can't give an exact explanation, why cargo is running two instances of `build[.rs]` execution in our particular case at all, but reports about other issues caused by parallel execution in cargo (https://github.com/rust-lang/cargo/issues/851, https://github.com/rust-lang/cargo/issues/1569,  https://github.com/rust-lang/cargo/issues/5575) and recent changes in the build system (https://internals.rust-lang.org/t/evaluating-pipelined-rustc-compilation/10199) suggest the assumption, that we always should take care of this kind behavior resp. avoid any side effects of parallel invocation in our build scripts, because parallel runs of the build script may happen in cargo for many reasons. 

utilizing checksums or renaming/moving the completed file by an atomic operation, when the download is done, are IMHO just very unsatisfying workarounds, which do not solve the actual cause of the defect. in fact, they will most likely just result in a second parallel download attempt instead of blocking any further processing until the still ongoing first download finally becomes ready. that's why i suggested and used the utilization of lock files as much more adequate answer to this challenge in my PR (https://github.com/denoland/rusty_v8/pull/543).

bnoordhuis:
I went over this with @bartlomieju yesterday and we concluded that there's no way for the build script to exhibit such behavior barring bugs in cargo. Possible, of course, but not reproducible. Neither of us really liked the idea of landing fixes for a poorly understood problem.

[...]

PR #547, opened 2020-11-30
Write librusty_v8.a sha-256 checksum to file.

bnoordhuis:


piscisaureus:
What's the purpose of this change?
I suspect it has something to do with re-downloading librusty_v8.a after upgrading the rusty_v8 crate.
But there is no way for build.rs to know the "correct" shasum so I don't see how it would work...

bartlomieju:
@piscisaureus https://github.com/denoland/rusty_v8/issues/545

piscisaureus:
@bartlomieju 

There is no way for build.rs to know the "correct" shasum so I don't see how it would work. The only way to find out would be to download the .shasum file on every build, but this will make people that use a mirror unhappy.

My suggestion would be to append the crate version to the filename instead.

bartlomieju:
> The only way to find out would be to download the .shasum file on every build, but this will make people that use a mirror unhappy.

That was the plan to download checksum alongside static lib and verify that checksum of static lib's checksum matches. Probably could skip it for mirrors?

piscisaureus:
@bartlomieju 

Right - so this is not intended to solve the issue that the static library build should be re-downloaded after a rusty_v8 upgrade?

I am aware that some people experience link errors, but that's pretty much always because the download failed entirely and left an empty librusty_v8.a in the build directory. There's no need for shasumming to detect that.

bnoordhuis:
> so this is not intended to solve the issue that the static library build should be re-downloaded after a rusty_v8 upgrade?

No, it's to catch partial or corrupted downloads.

bnoordhuis:
Ping @piscisaureus - your thoughts?

Cargo.lock
```diff
@@ -6,12 +6,52 @@ version = "1.2.1"
...
+
+[[package]]
+name = "generic-array"
+version = "0.14.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "501466ecc8a30d1d3b7fc9229b122b2ce8ed6e9d9223f1138d4babb253e51817"
+dependencies = [
+ "typenum",
+ "version_check",
+]
```
ry, line 53: It's rather unfortunate that this patch adds a bunch of dependencies just to calculate the checksum. 

https://crates.io/crates/sha1 for example is self-contained. Maybe we could use that - or an already installed command-line utility?

ry:
Closing because stale.

PR #549, opened 2020-12-02
Make Private work with context-less HandleScope

bnoordhuis:
v8::Private is like v8::String and other primitives, it doesn't need an
active v8::Context.


ry (approved):
Macht sin. Gracias.

PR #550, opened 2020-12-03
Make Symbol work with context-less HandleScope

bnoordhuis:
v8::Symbol is like v8::String and other primitives, it doesn't need an
active v8::Context.

See also #549.


piscisaureus:
LGTM

piscisaureus (approved):
LGTM

Issue #551, opened 2020-12-06
Support loading snapshots before snapshotting


NickGeek:
Hi,
I noticed that in `runtime.rs` there is a TODO around supporting loading snapshots and snapshotting on the same instance. It's a feature I'd find quite useful so I did some digging.

It looks like V8 supports receiving a `StartupData` in the constructor of `SnapshotCreator` ([source](https://source.chromium.org/chromium/chromium/src/+/master:v8/include/v8.h;drc=ed64b9822219f26660592ee5048418c1d4ade2a9;l=10025)). I tried naively updating `rusty_v8` to use that parameter, but I get an assertion failure:
```
# Fatal error in ../../../../v8/src/api/api.cc, line 636
# Debug check failed: data->created_.
```

Is there anything I can do to help get that todo item done?

bartlomieju:
@NickGeek I've transferred the issue to `rusty_v8` repo as it must first be implemented here before we can add that feature to `deno` itself.

bartlomieju:
@NickGeek please open a PR with your code; the error you got suggests that snapshot was not taken before dropping the isolate.

NickGeek:
@bartlomieju Sweet, thanks. I'll try and get that up tonight. Just a warning: it's definitely in a draft state, I was mostly just seeing if I could get something to work.

bartlomieju:
This is now supported.

PR #555, opened 2020-12-09
WIP: Support loading snapshots before snapshotting

NickGeek:
This is my current attempt at allowing snapshots to be loaded before another snapshot is taken on an isolate.

It currently doesn't work with the error message (see: #551)
```
# Fatal error in ../../../../v8/src/api/api.cc, line 636
# Debug check failed: data->created_.
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -56,7 +56,7 @@ dependencies = [
 
 [[package]]
 name = "rusty_v8"
-version = "0.14.0"
+version = "0.15.0"
 dependencies = [
  "bitflags",
  "cargo_gn",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rusty_v8"
-version = "0.14.0"
+version = "0.15.0"
 description = "Rust bindings to V8"
 readme = "README.md"
 authors = ["the Deno authors"]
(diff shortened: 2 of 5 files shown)


bnoordhuis:
When exactly are you getting that error? I can tell you how it happens: the `SnapshotCreator` is dropped, without calling `create_blob()` first - it's a kind of unfortunate debug check in V8.

I can probably help you troubleshoot but it'd be helpful if there was a test in `tests/test_api.rs` so we know we're testing the same thing.

NickGeek:
@bnoordhuis Hi, I've pushed up a test case, although it annoyingly passes on my machine.
In `deno_core` I got the issue by using `rusty_v8` with my changes and updating from line 207 of `runtime.rs`:
```rust
let (mut isolate, maybe_snapshot_creator) = if options.will_snapshot {
let existing_blob = options.startup_snapshot.map(|snapshot| {
  match snapshot {
    Snapshot::JustCreated(startup_data) => startup_data,
    _ => todo!()
  }
});

let mut creator =
v8::SnapshotCreator::new(Some(&bindings::EXTERNAL_REFERENCES), existing_blob);
```

ry:
@NickGeek Thanks for this work - it's definitely a desired feature. But I'm going to close this PR because it's stale. You're very welcome to rebase your changes and open a new PR - I'll see to it that it gets reviewed quickly. - Ryan

PR #557, opened 2020-12-13
Add 'origin' to V8Inspector::context_created

### LINKED ISSUE #552: V8Inspector::context_created should accept V8ContextInfo argument
https://github.com/denoland/rusty_v8/blob/efe0e76a75603b70666297e9782ba01bd8ad0343/src/inspector.rs#L893-L895

It'd be quite useful if we could pass `V8ContextInfo` - it allows to set `origin` field which would help reduce some technical debt in `deno` related to REPL handling:

https://github.com/denoland/deno/blob/7135d34ccab7629da57c75ee239fcb0dda733eae/cli/module_loader.rs#L59-L64

https://github.com/denoland/deno/blob/7135d34ccab7629da57c75ee239fcb0dda733eae/core/module_specifier.rs#L58-L60

CC @bnoordhuis

bartlomieju:
Closes #552

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1793,9 +1793,12 @@ v8_inspector::V8InspectorSession* v8_inspector__V8Inspector__connect(
 
 void v8_inspector__V8Inspector__contextCreated(
     v8_inspector::V8Inspector* self, const v8::Context& context,
-    int contextGroupId, v8_inspector::StringView humanReadableName) {
-  self->contextCreated(v8_inspector::V8ContextInfo(
-      ptr_to_local(&context), contextGroupId, humanReadableName));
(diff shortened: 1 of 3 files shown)


bnoordhuis (approved):
LGTM but consider adding a test that verifies the origin is propagated correctly.

(I'm sympathetic to the "too hard" argument and the changes in this PR look obviously correct but without regression tests it might break again in the future.)

bartlomieju:
Actually I don't know if I want to land this PR, I've tried it in `deno` but it doesn't solve the problem described in the issue. I'll get back to it some time

ry:
Closing because stale. Please rebase and reopen if you want to merge. Seems like we could add this binding even if it doesn't fix something in Deno...

PR #558, opened 2020-12-20
improve ninja/gn detection

avindra:
This should allow people to build without explicitly setting `GN` and `NINJA` env vars.

If the tool(s) are detected, the build will continue without trying to download the binaries.


bnoordhuis (approved):
Thanks Avindra, LGTM.

PR #560, opened 2020-12-25
Add Isolate::set_wasm_streaming_callback()

### LINKED ISSUE #556: Expose wasi_streaming_callback
In Deno, our implementation of `WebAssembly.instantiateStreaming` is broken because I thought the implementation provided by wasi-test-streaming would be runtime API compatible with any implementation of Response.

Turns out, from what I can tell in the implementation found here https://github.com/v8/v8/blob/4b9b23521e6fd42373ebbcb20ebe03bf445494f9/src/wasm/wasm-js.cc#L585 that we're supposed to override wasi_streaming_callback defined here https://github.com/v8/v8/blob/4b9b23521e6fd42373ebbcb20ebe03bf445494f9/src/isolate.h#L372

So for rusty_v8 we need Isolate to expose this callback.

bnoordhuis:
Add the hook necessary to implement `WebAssembly.compileStreaming()`.

Fixes #556

cc @caspervonb

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -266,6 +266,11 @@ void v8__Isolate__SetAllowAtomicsWait(v8::Isolate* isolate, bool allow) {
   isolate->SetAllowAtomicsWait(allow);
 }
 
+void v8__Isolate__SetWasmStreamingCallback(v8::Isolate* isolate,
+                                           v8::WasmStreamingCallback callback) {
+  isolate->SetWasmStreamingCallback(callback);
+}
+
 void v8__Isolate__CreateParams__CONSTRUCT(
     uninit_t<v8::Isolate::CreateParams>* buf) {
   construct_in_place<v8::Isolate::CreateParams>(buf);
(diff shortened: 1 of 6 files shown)


src/binding.cc
```diff
@@ -2029,6 +2034,38 @@ MaybeBool v8__Module__SetSyntheticModuleExport(const v8::Module& self,
       isolate, ptr_to_local(export_name), ptr_to_local(export_value)));
 }
 
+struct WasmStreamingSharedPtr {
+  std::shared_ptr<v8::WasmStreaming> inner;
+};
+
+static_assert(sizeof(WasmStreamingSharedPtr) <= 2 * sizeof(void*),
```
ry, line 2041: Why <= ?

src/wasm.rs
```diff
@@ -0,0 +1,105 @@
...
+    let zero = null_mut();
+    let mut that = WasmStreamingSharedPtr([zero, zero]);
+    unsafe {
+      v8__WasmStreaming__Unpack(scope.get_isolate_ptr(), data, &mut that)
+    };
+    let source = args.get(0);
+    (F::get())(scope, source, WasmStreaming(that));
+  }
+  c_fn::<F>
+}
```
ry, line 86: nice!

tests/test_api.rs
```diff
@@ -4326,3 +4327,68 @@ fn clear_kept_objects() {
...
+  eval(scope, script).unwrap();
+
+  let global = context.global(scope);
+  let name = v8::String::new(scope, "result").unwrap().into();
+  assert!(global.get(scope, name).unwrap().is_null());
+
+  let mut ws = WS.with(|slot| slot.borrow_mut().take().unwrap());
+  assert!(global.get(scope, name).unwrap().is_null());
+
+  ws.on_bytes_received(&[0x00, 0x61, 0x73, 0x6d, 0x01, 0x00, 0x00, 0x00]);
```
ry, line 4368: What is this wasm code - maybe put a comment with the WAT?

ry (approved):
LGTM

src/binding.cc
bnoordhuis: `std::shared<T>` is normally two pointers big but it's conceivable that an alternative libc++ might implement it as a single pointer. IOW, over-allocating is okay, under-allocating is not.

tests/test_api.rs
bnoordhuis: It doesn't contain any actual WASM bytecode, just the file magic and WASM version. :-) I'll add a comment.

src/binding.cc
```diff
@@ -2029,6 +2034,38 @@ MaybeBool v8__Module__SetSyntheticModuleExport(const v8::Module& self,
       isolate, ptr_to_local(export_name), ptr_to_local(export_value)));
 }
 
+struct WasmStreamingSharedPtr {
```
piscisaureus, line 2037: We do not usually wrap a shared_ptr in a struct. Why is that necessary here?

piscisaureus:
I don't understand why the existing infrastructure to wrap C++ `shared_ptr` can't be used here, but I'll admit that there might be a good reason.

src/binding.cc
bnoordhuis: It's not strictly necessary but I like the newtype idiom. It makes it harder to use it wrong.

It's also nice to have the same type name in C++ and Rust, easier to grep.

bnoordhuis:
Explained in the comment here :-)

https://github.com/denoland/rusty_v8/pull/560/files#diff-6d7bc9e99526e7f06b818a6103bd30aaa067263cc8e34a14197e1a3bcf2dd4dbR14-R21

PR #562, opened 2020-12-25
Isolate::thread_safe_handle no longer requires mut ref

bartlomieju:


src/isolate.rs
```diff
@@ -672,7 +672,7 @@ impl IsolateHandle {
     self.0.isolate
   }
 
-  fn new(isolate: &mut Isolate) -> Self {
+  fn new(isolate: &Isolate) -> Self {
```
bnoordhuis, line 675: FWIW, this doesn't look Obviously Correct to me because now you can go from `&Isolate` to `*mut Isolate` because of the `get_isolate_ptr(&self)` method a few lines up. It's `pub(crate)`, but still.

Aside: `HandleHost::get_isolate_handle()` in `src/handle.rs` probably no longer needs `as_mut()` with this change.

bartlomieju: Thanks, you're correct, I will wait for @piscisaureus for some hints

piscisaureus: > FWIW, this doesn't look Obviously Correct to me because now you can go from `&Isolate` to `*mut Isolate` because of the `get_isolate_ptr(&self)` method a few lines up. It's `pub(crate)`, but still.

I appreciate that this gets careful review, however I don't see the problem with a transition from `&Isolate` to `*mut Isolate` as such. However if this introduces an (indirect) way to transition from `&Isolate` to &mut Isolate`, it would indeed be bad.

piscisaureus (approved):

PR #570, opened 2021-01-07
Remove set_flags_from_command_line_with_usage test

danbev:
This commit removes the set_flags_from_command_line_with_usage and puts
it in an example code section instead.

The motivation for doing this is that the test output currently contains
the usage string and all the V8 options which creates a lot of output
when the tests is run regardless if --nocapture is used or not.


bnoordhuis (approved):
That test was on the chopping block anyway because of https://github.com/denoland/rusty_v8/pull/567#discussion_r549444149 so LGTM and thanks!

PR #572, opened 2021-01-08
Add Script::get_unbound_script and UnboundScript::bind_to_current_context.

losfair:
I'm working on [a project](https://github.com/losfair/rusty-workers) where I want to reuse the same `Isolate` with different `Context`'s and cache compiled scripts.

rusty_v8 currently only exposes `Script` that is bound to a specific context, and adding conversion to and from `UnboundScript` allows using a compiled script from different contexts.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1503,11 +1503,13 @@ const v8::Script* v8__Script__Compile(const v8::Context& context,
                           const_cast<v8::ScriptOrigin*>(&origin)));
 }
 
-const v8::UnboundScript* v8__Script__GetUnboundScript(const v8::Script& script) {
+const v8::UnboundScript* v8__Script__GetUnboundScript(
+    const v8::Script& script) {
   return local_to_ptr(ptr_to_local(&script)->GetUnboundScript());
 }
 
-const v8::Script* v8__UnboundScript__BindToCurrentContext(const v8::UnboundScript& unbound_script) {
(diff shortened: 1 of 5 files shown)


src/binding.cc
```diff
@@ -1503,6 +1503,14 @@ const v8::Script* v8__Script__Compile(const v8::Context& context,
                           const_cast<v8::ScriptOrigin*>(&origin)));
 }
 
+const v8::UnboundScript* v8__Script__GetUnboundScript(const v8::Script& script) {
+  return local_to_ptr(ptr_to_local(&script)->GetUnboundScript());
+}
+
+const v8::Script* v8__UnboundScript__BindToCurrentContext(const v8::UnboundScript& unbound_script) {
```
bnoordhuis, line 1510: Can you wrap lines at 80 columns? Thanks.

bnoordhuis (commented):
You'll also need to make the appropriate changes to `src/data.rs`. Have a look at how it's done for `Script`, you can probably just follow that.

Can you add one or two tests to `tests/test_api.rs` and run `cargo format`? Thanks!

src/binding.cc
losfair: Done!

losfair:
Added a test for converting to and from `UnboundScript`.

Just checked `src/data.rs` - it seems that the `UnboundScript` type and the related `impl`'s are already there?

piscisaureus (approved):

PR #576, opened 2021-01-09
dontmerge! try to fix sccache ci issue

bnoordhuis:


bnoordhuis:
@piscisaureus We're getting this "corrupted sccache" error again. I tried to `rm -rf` it and that works for that particular run, but it comes back on the next run (see e.g. #572.)

PR #578, opened 2021-01-11
Expose the ArrayBuffer::Allocator interface.

losfair:
This pull request exposes the `ArrayBuffer::Allocator` interface to Rust by adding a `new_rust_allocator` method and passing explicit vtables to C++.

Based on https://github.com/denoland/rusty_v8/pull/572 .

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1503,11 +1503,13 @@ const v8::Script* v8__Script__Compile(const v8::Context& context,
                           const_cast<v8::ScriptOrigin*>(&origin)));
 }
 
-const v8::UnboundScript* v8__Script__GetUnboundScript(const v8::Script& script) {
+const v8::UnboundScript* v8__Script__GetUnboundScript(
+    const v8::Script& script) {
   return local_to_ptr(ptr_to_local(&script)->GetUnboundScript());
 }
 
-const v8::Script* v8__UnboundScript__BindToCurrentContext(const v8::UnboundScript& unbound_script) {
+const v8::Script* v8__UnboundScript__BindToCurrentContext(
+    const v8::UnboundScript& unbound_script) {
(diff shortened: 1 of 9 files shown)


src/binding.cc
```diff
@@ -1105,10 +1105,57 @@ size_t v8__ArrayBufferView__CopyContents(const v8::ArrayBufferView& self,
...
+  const RustAllocatorVtable* vtable;
+
+ public:
+  RustAllocator(void* handle, const RustAllocatorVtable* vtable) {
+    this->handle = handle;
+    this->vtable = vtable;
+  }
+
+  RustAllocator(const RustAllocator& that) = delete;
+  RustAllocator(RustAllocator&& that) = delete;
```
bnoordhuis, line 1129:
```suggestion
  RustAllocator(const RustAllocator& that) = delete;
  RustAllocator(RustAllocator&& that) = delete;
  void operator=(const RustAllocator& that) = delete;
  void operator=(RustAllocator&& that) = delete;
```

src/binding.cc
```diff
@@ -1105,10 +1105,57 @@ size_t v8__ArrayBufferView__CopyContents(const v8::ArrayBufferView& self,
...
+
+  virtual ~RustAllocator() { vtable->drop(handle); }
+
+  virtual void* Allocate(size_t length) {
```
bnoordhuis, line 1133: I'd mark these `final`:
```suggestion
  void* Allocate(size_t length) final {
```
`final` is only valid on virtual functions so it's equivalent to `virtual void* Allocate(size_t) final override`if you omit the `virtual`.

bnoordhuis (commented):
Thanks, mostly LGTM and very useful!

bnoordhuis (approved):

PR #579, opened 2021-01-11
Rekey GitHub actions cache

piscisaureus:


piscisaureus:
Apparently the issue is not caused by a corrupted GH actions cache. Must be an sccache bug, then.

PR #585, opened 2021-01-15
Support setting OOM callback on `Isolate`.

losfair:
Exposes `v8::Isolate::SetOOMErrorHandler`.

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -4430,7 +4430,10 @@ fn run_with_rust_allocator() {
     count.fetch_add(n, Ordering::SeqCst);
     Box::into_raw(vec![0u8; n].into_boxed_slice()) as *mut [u8] as *mut c_void
   }
-  unsafe extern "C" fn allocate_uninitialized(count: &AtomicUsize, n: usize) -> *mut c_void {
+  unsafe extern "C" fn allocate_uninitialized(
+    count: &AtomicUsize,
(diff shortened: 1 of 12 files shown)


tests/test_api.rs
```diff
@@ -4515,3 +4515,47 @@ fn run_with_rust_allocator() {
...
+  extern "C" {
+    fn setjmp(env: &mut [usize; 32]) -> i32;
+    fn longjmp(env: &[usize; 32], value: i32) -> !;
+  }
+
+  static mut JMPBUF: [usize; 32] = [0; 32];
+
+  extern "C" fn oom_callback(_: *const std::os::raw::c_char, _: bool) {
+    unsafe {
+      longjmp(&JMPBUF, 42);
```
bnoordhuis, line 4530: This isn't safe (and indeed the test suite is failing on the windows buildbot) because V8 - possibly the Rust runtime too - is in no way longjmp-proof.

V8 heavily depends on RAII. When you longjmp, no destructors run and you basically clobber its internal state. That affects other tests too because they run in the same process.

Basically, you can test up to the `isolate.set_oom_error_handler(oom_callback)` call, but no further.

losfair: Yeah it seems that the test is broken on Windows, triggering some undefined behavior here.

Updated.

bnoordhuis (approved):

PR #593, opened 2021-01-21
Exclude tools/clang from publish

bnoordhuis:
The only file that is actually needed is tools/clang/scripts/update.py
and that's already explicitly excluded from the exclude list.

Shrinks the (unpacked) crate size by 2.4 MB.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ exclude = [
  "buildtools/third_party/libc++/trunk/utils/",
  "buildtools/third_party/libc++/trunk/www/",
  "buildtools/third_party/libc++abi/trunk/test/",
+ "tools/clang",
  "v8/ChangeLog",
  "v8/benchmarks/",
  "v8/docs/",


ry (commented):
I'm a bit worried this will break V8_FROM_SOURCE=1 for crate users... Isn't the clang binary stored there?

piscisaureus:
The clang binary is downloaded by update.py.

Not-from-source builds don't touch clang at all.

piscisaureus (approved):

ry:
> The clang binary is downloaded by update.py.

doesn’t that break V8_FROM_SOURCE for dependents?

piscisaureus:
> doesn’t that break V8_FROM_SOURCE for dependents?

No - build.rs will execute update.py if necessary: https://github.com/denoland/rusty_v8/blob/1fb0e9436c865f68db2150ec3dcc25b503660324/build.rs#L349

ry:
But that file is now excluded from the package..?

avindra:
It should™ be included per the Cargo.toml `exclude` rule:

https://github.com/denoland/rusty_v8/blob/1fb0e9436c865f68db2150ec3dcc25b503660324/Cargo.toml#L44

PR #598, opened 2021-02-03
Add support for import assertions

bartlomieju:


src/binding.cc
```diff
@@ -2060,6 +2069,13 @@ MaybeBool v8__Module__InstantiateModule(const v8::Module& self,
       ptr_to_local(&self)->InstantiateModule(ptr_to_local(&context), cb));
 }
 
+MaybeBool v8__Module__InstantiateModuleNew(const v8::Module& self,
+                                        const v8::Context& context,
+                                        v8::Module::ResolveModuleCallback cb) {
```
ry, line 2074: run clang-format

src/isolate_create_params.rs
```diff
@@ -192,6 +192,7 @@ pub(crate) mod raw {
     pub embedder_wrapper_type_index: int,
     pub embedder_wrapper_object_index: int,
     pub cpp_heap_params: SharedPtr<CppHeapCreateParams>,
+    // NOTE(bartlomieju): this field is deprecated in V8 API.
```
ry, line 195: Use deprecated attribute? https://doc.rust-lang.org/edition-guide/rust-2018/the-compiler/an-attribute-for-deprecation.html

src/fixed_array.rs
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  pub fn get<'s>(
+    &self,
+    scope: &mut HandleScope<'s>,
+    index: usize,
+  ) -> Local<'s, Data> {
+    unsafe {
+      scope.cast_local(|sd| {
+        v8__FixedArray__Get(self, sd.get_current_context(), index as int)
```
bnoordhuis, line 31: The `index as int` cast is problematic for two reasons:

1. V8 aborts with a runtime error for indices > length
2. Negative indices (which can happen for values > `int::MAX`) result in out-of-bounds reads

My suggestion: check the index against the length first and return `None` if it's not in range.

src/module.rs
```diff
@@ -85,6 +86,58 @@ where
...
+  }
+
+  #[cfg(target_os = "windows")]
+  fn mapping() -> Self {
+    let f = |ret_ptr, context, specifier, import_assertions, referrer| {
+      let r = (F::get())(context, specifier, import_assertions, referrer)
+        .map(|r| -> *const Module { &*r })
+        .unwrap_or(null());
+      unsafe { std::ptr::write(ret_ptr, r) }; // Write result to stack.
+      ret_ptr // Return stack pointer to the return value.
```
bnoordhuis, line 135: IMO, these comments are kind of confusing. They suggest we're directly manipulating the stack but in fact we're writing to the storage allocated by our caller, and that _may_ be on the stack.

`s/stack/caller-allocated memory/` and all should be good. :-)

src/isolate_create_params.rs
bartlomieju: I don't think that's a good idea, it will show up as a warning and this field cannot be used/configured by user whatsoever.

src/fixed_array.rs
bartlomieju: Sure, I'll do that.

src/module.rs
bartlomieju: That's a copy-paste from `ResolveCallback` written by Bert, I can update them tho.

bartlomieju:
@ry @bnoordhuis I don't think it makes sense to add `ModuleResolveCallback` and keep old `ResolveCallback` as well (which is now deprecated in V8 API; the only difference being `import_assertions` field). I will remove all code related to using the old APIs and replace with the new API.

src/fixed_array.rs
bartlomieju: Done, I left cast to `as int` to make compiler happy

ry (approved):
LGTM - nice work

We'll have to bump the minor version on next release due to API change. But we do that all the time anyway.

src/fixed_array.rs
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    Some(
+      unsafe {
+        scope.cast_local(|sd| {
+          v8__FixedArray__Get(self, &*sd.get_current_context(), index as int)
+        })
+      }
+      .unwrap(),
+    )
```
bnoordhuis, line 40:
```suggestion
    unsafe {
      scope.cast_local(|sd| {
        v8__FixedArray__Get(self, &*sd.get_current_context(), index as int)
      })
    }
```
(`scope.cast_local()` already returns `Option`)

src/module.rs
```diff
@@ -121,7 +126,7 @@ where
         .map(|r| -> *const Value { &*r })
         .unwrap_or(null());
       unsafe { std::ptr::write(ret_ptr, r) }; // Write result to stack.
-      ret_ptr // Return stack pointer to the return value.
+      ret_ptr // Return caller-allocated memory pointer to the return value.
```
bnoordhuis, line 129: You're right you're not the one introducing the verbiage. I'm okay with backing it out for now so it's at least consistent everywhere.

bnoordhuis (approved):
Thanks Bartek, LGTM with two suggestions.

PR #606, opened 2021-02-09
docs: add info about settings --harmony-import-assertions flag

bartlomieju:


src/module.rs
```diff
@@ -284,6 +284,8 @@ impl Module {
   /// Returns an empty Maybe<bool> if an exception occurred during
   /// instantiation. (In the case where the callback throws an exception, that
   /// exception is propagated.)
+  ///
+  /// NOTE: requires to set `--harmony-import-assertions` V8 flag.
```
ry, line 288: Calling this at all need the flag?

bartlomieju: Yes it does. I might have made a mistake removing old methods, could have left them with `deprecated_` prefix.

ry: Ok.

Well - we apologize to any onlookers for the bumpy road.

ry (changes requested):
LGTM

ry (approved):
LGTM

PR #609, opened 2021-02-09
Fix ValueDeserializer::read_value() lifetime

### LINKED ISSUE #607: Value returned by "value_deserializer" cannot outlive the "value_deserializer"
The returned value from "value_deserializer.read_value(...)" should be able to outlive value_deserializer.
Current lifeline configuration does not allow this.
Is needed for this https://github.com/denoland/deno/pull/9323 prototype.

bnoordhuis:
The return value should have the lifetime of the context, not the
deserializer.

Fixes #607.

### DIFF
--- a/src/value_deserializer.rs
+++ b/src/value_deserializer.rs
@@ -261,7 +261,10 @@ pub trait ValueDeserializerHelper {
     .into()
   }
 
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

Issue #610, opened 2021-02-09
Build from source broken for dependent crates


ry:
```
~/src/deno> V8_FROM_SOURCE=1 GN_ARGS="host_cpu=\"arm64\"" cargo build -vv
...
c/deno/target/debug/deps/libhyper-cd26dee30b6158d3.rlib --extern lazy_static=/Users/ry/src/deno/target/debug/deps/liblazy_static-09260e36393089cb.rlib --extern os_pipe=/Users/ry/src/deno/target/debug/deps/libos_pipe-931cb976d8d76367.rlib --extern pty=/Users/ry/src/deno/target/debug/deps/libpty-86d6ee2213cbbcc7.rlib --extern regex=/Users/ry/src/deno/target/debug/deps/libregex-50fc994c8e78ec67.rlib --extern serde=/Users/ry/src/deno/target/debug/deps/libserde-05c5b8a383cc2a59.rlib --extern tempfile=/Users/ry/src/deno/target/debug/deps/libtempfile-d3043d73d17898d5.rlib --extern test_util=/Users/ry/src/deno/target/debug/deps/libtest_util-c17416a61e5d687b.rlib --extern tokio=/Users/ry/src/deno/target/debug/deps/libtokio-15e5426244b92668.rlib --extern tokio_rustls=/Users/ry/src/deno/target/debug/deps/libtokio_rustls-8458d849548e734e.rlib --extern tokio_tungstenite=/Users/ry/src/deno/target/debug/deps/libtokio_tungstenite-ba34e393c09cfcf7.rlib -L native=/Users/ry/src/deno/target/debug/build/ring-6d0f8897d550f2a4/out`
[rusty_v8 0.17.0] Done. Made 160 targets from 89 files in 2274ms
[rusty_v8 0.17.0] cargo:rustc-env=GN_OUT_DIR=/Users/ry/src/deno/target/debug/gn_out
[rusty_v8 0.17.0] running: "/Users/ry/src/deno/target/debug/ninja_gn_binaries-20210101/mac/ninja" "-C" "/Users/ry/src/deno/target/debug/gn_out" "rusty_v8"
[rusty_v8 0.17.0] ninja: Entering directory `/Users/ry/src/deno/target/debug/gn_out'
[...]

PR #611, opened 2021-02-09
include v8/tools/gen-v8-gn.py in package

ry:
Maybe fixes https://github.com/denoland/rusty_v8/issues/610 

I'm not sure how I can test this without uploading to crates.io. Suggestions?


bnoordhuis (approved):
> I'm not sure how I can test this without uploading to crates.io. Suggestions?

I think it should be possible to set up a local registry and publish it there first. Lot of work though.

PR #615, opened 2021-02-10
Hack ICU build scripts so we don't have to depend on icudata file

ry:
Here is the hack: https://github.com/denoland/icu/commit/28b0e9ea59878fdd1682593be2ac489a6a6bbb21

Note the size of [0.18.0](https://crates.io/crates/rusty_v8/0.18.0)  on crates.io  is 19.5 MB. This change will bring that back down to a reasonable level.


bartlomieju (approved):
LGTM

PR #627, opened 2021-02-19
upgrade to v8 9.0.217

### LINKED ISSUE #626: Not entering isolate breaks profiling
rusty_v8's clever approach to avoiding entering the isolate breaks `--prof` and the programmatic sampling profiler API.

The profiler is a thread that sends SIGPROF signals to the isolate thread but the signal handler doesn't know where to store the sample when `Isolate::Enter()` has not been called.

Case in point: it starts working when I manually call `unsafe { v8__Isolate__Enter(&mut *isolate) }` in deno.

ry:
Fixes #626.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Rusty V8 Binding
 
-V8 Version: 9.0.123
+V8 Version: 9.0.217
 
 [![ci](https://github.com/denoland/rusty_v8/workflows/ci/badge.svg?branch=main)](https://github.com/denoland/rusty_v8/actions)
 [![crates](https://img.shields.io/crates/v/rusty_v8.svg)](https://crates.io/crates/rusty_v8)
--- a/src/exception.rs
+++ b/src/exception.rs
@@ -305,11 +305,12 @@ impl Exception {
     message: Local<String>,
     contructor: unsafe extern "C" fn(*const String) -> *const Value,
   ) -> Local<'s, Value> {
-    scope.enter_isolate();
-    let error =
-      unsafe { scope.cast_local(|_| (contructor)(&*message)) }.unwrap();
-    scope.exit_isolate();
-    error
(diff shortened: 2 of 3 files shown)


ry:
I'm experiencing reliable segfaults in both OSX and Linux on a number of tests inspector_schedule_pause_on_next_statement, clear_kept_objects, terminate_execution and probably other tests.

build and buildtools should be correctly updated.

I'm starting to think perhaps we're interacting with the API incorrectly, and the problem is only now showing itself due to internal changes. Could this be related to https://github.com/denoland/rusty_v8/issues/626 ?

tests/test_api.rs
```diff
@@ -4425,6 +4425,7 @@ fn clear_kept_objects() {
   let _setup_guard = setup();
 
   let isolate = &mut v8::Isolate::new(Default::default());
+  unsafe { crate::v8::isolate::v8__Isolate__Enter(&mut **isolate) }
```
piscisaureus, line 4428: Move this into the rusty_v8 function that needs it.

bnoordhuis:
The segfaults are caused by calls to `Isolate::Current()` inside V8. The failing inspector tests are due to one such call in the `AbstractBytecodeArray` constructor. It's a TODO, going by the comment preceding it but yeah, that doesn't help us now.

ry:
I'm proposing a fairly radical change in this PR: enter the isolate on construction, exit it on drop. This makes sense for our use case in Deno, but for users who intend to use the Isolate on a different thread would need to first call `Isolate::exit_isolate()`. Probably a better solution would be to have some scope object to handle the entry and exit, but this might suffice as a simple interim solution.

bnoordhuis (approved):
It'll stop working when Lockers make a rentrée but a good interim solution until then.

Also fixes #626.

PR #633, opened 2021-02-24
Upgrade V8 to 9.0.257.3 and bump version to v0.20.0

ry:
Preparing for Deno 1.8.0


bnoordhuis (approved):

bnoordhuis:
The V8 branch cut just happened. 9.0.257 is stable and will be what gets rolled into Chrome 90 (but we should just track `﻿branch-heads/9.0﻿` now that it exists.)

PR #634, opened 2021-02-25
test(function): add tests for return values of a function

cola119:
Hi there. I'm reading the code to grasp and learn internal architecture.
I found trivial tasks around tests of `src/function` so fixed.


ry (approved):
LGTM - thanks

PR #635, opened 2021-02-25
Add code cache API.

losfair:
This pull request exposes the V8 `CachedData` interface and supports:

- Creating code cache from `UnboundScript` and `UnboundModuleScript`.
- Creating `Source` with cached code.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -345,6 +345,11 @@ v8::ScriptCompiler::CachedData* v8__ScriptCompiler__CachedData__NEW(
       data, length, v8::ScriptCompiler::CachedData::BufferNotOwned);
 }
 
+void v8__ScriptCompiler__CachedData__DELETE(
+    v8::ScriptCompiler::CachedData* self) {
+  delete self;
+}
+
 const v8::ScriptCompiler::CachedData* v8__ScriptCompiler__Source__GetCachedData(
     const v8::ScriptCompiler::Source* source) {
   return source->GetCachedData();
@@ -1583,6 +1588,17 @@ const v8::UnboundScript* v8__Script__GetUnboundScript(
 const v8::Script* v8__UnboundScript__BindToCurrentContext(
     const v8::UnboundScript& unbound_script) {
   return local_to_ptr(ptr_to_local(&unbound_script)->BindToCurrentContext());
+}
+
+v8::ScriptCompiler::CachedData* v8__UnboundScript__CreateCodeCache(
+    const v8::UnboundScript& unbound_script) {
+  return v8::ScriptCompiler::CreateCodeCache(ptr_to_local(&unbound_script));
+}
+
+v8::ScriptCompiler::CachedData* v8__UnboundModuleScript__CreateCodeCache(
+    const v8::UnboundModuleScript& unbound_module_script) {
+  return v8::ScriptCompiler::CreateCodeCache(
+      ptr_to_local(&unbound_module_script));
 }
 
 const v8::Value* v8__Script__Run(const v8::Script& script,
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -64,6 +64,17 @@ static_assert(sizeof(v8::CFunction) == sizeof(size_t) * 2,
 static_assert(sizeof(three_pointers_t) == sizeof(v8_inspector::StringView),
               "StringView size mismatch");
 
+static_assert(sizeof(v8::ScriptCompiler::CachedData) == 24,
+              "CachedData size mismatch");
+static_assert(offsetof(v8::ScriptCompiler::CachedData, data) == 0,
+              "CachedData.data offset mismatch");
+static_assert(offsetof(v8::ScriptCompiler::CachedData, length) == 8,
+              "CachedData.length offset mismatch");
+static_assert(offsetof(v8::ScriptCompiler::CachedData, rejected) == 12,
+              "CachedData.rejected offset mismatch");
+static_assert(offsetof(v8::ScriptCompiler::CachedData, buffer_policy) == 16,
+              "CachedData.buffer_policy offset mismatch");
+
 enum InternalSlots {
   kSlotDynamicImport = 0,
   kNumInternalSlots,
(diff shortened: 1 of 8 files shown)


tests/test_api.rs
```diff
@@ -4778,3 +4778,102 @@ fn icu_collator() {
...
+
+#[test]
+fn unbound_module_script_conversion() {
+  let _setup_guard = setup();
+  let isolate = &mut v8::Isolate::new(Default::default());
+  let scope = &mut v8::HandleScope::new(isolate);
+  let context = v8::Context::new(scope);
+  let mut scope = v8::ContextScope::new(scope, context);
+  create_unbound_module_script(&mut scope, "'Hello ' + value", None);
+}
```
ry, line 4831: This test has no asserts. Seems kind of useless?

tests/test_api.rs
```diff
@@ -4778,3 +4778,102 @@ fn icu_collator() {
...
+    let isolate = &mut v8::Isolate::new(Default::default());
+    let scope = &mut v8::HandleScope::new(isolate);
+    let context = v8::Context::new(scope);
+    let mut scope = v8::ContextScope::new(scope, context);
+    let unbound_module_script =
+      create_unbound_module_script(&mut scope, CODE, None);
+    unbound_module_script.create_code_cache().unwrap().to_vec()
+  };
+
+  {
```
ry, line 4857: This block seems superfluous

src/lib.rs
```diff
@@ -125,6 +126,7 @@ pub use scope::EscapableHandleScope;
 pub use scope::HandleScope;
 pub use scope::TryCatch;
 pub use script::ScriptOrigin;
+pub use script_compiler::CachedData;
```
ry, line 129: BufferPolicy should also be exported?

Maybe rename to CachedDataBufferPolicy?

tests/test_api.rs
losfair: `UnboundModuleScript` doesn't have any methods except `create_code_cache` so this test just makes sure that none of the `.unwrap()`s fail.

src/lib.rs
losfair: The current `CachedData` API doesn't support owned buffers so `BufferPolicy` isn't used. Owned buffers are freed by C++ destructor and would need extra wrappers if using them from the Rust side is needed.

Should it still be exported if unused?

src/script_compiler.rs
```diff
@@ -29,15 +37,87 @@ extern "C" {
...
+/// functions in ScriptCompiler contains produce_data_to_cache = true. The data to cache
+/// can then can be retrieved from UnboundScript.
+#[repr(C)]
+#[derive(Debug)]
+pub struct CachedData<'a> {
+  data: *const u8,
+  length: i32,
+  rejected: bool,
+  buffer_policy: BufferPolicy,
+  _phantom: PhantomData<&'a ()>,
```
bnoordhuis, line 51: Can you add a static_assert to `src/binding.cc` that checks that the Rust and C++ definitions agree on the struct size?

src/script_compiler.rs
```diff
@@ -29,15 +37,87 @@ extern "C" {
...
+    origin: &ScriptOrigin,
+    cached_data: UniqueRef<CachedData>,
+  ) -> Self {
+    let mut buf = MaybeUninit::<Self>::uninit();
+    unsafe {
+      v8__ScriptCompiler__Source__CONSTRUCT(
+        &mut buf,
+        &*source_string,
+        origin,
+        cached_data.into_raw(),
```
bnoordhuis, line 112:
```suggestion
        cached_data.into_raw(), // Source constructor takes ownership.
```
Not immediately obvious from looking at just the code, I think.

src/support.rs
```diff
@@ -106,7 +106,7 @@ impl<T> From<UniqueRef<T>> for UniquePtr<T> {
 pub struct UniqueRef<T: ?Sized>(NonNull<T>);
 
 impl<T> UniqueRef<T> {
-  unsafe fn try_from_raw(ptr: *mut T) -> Option<Self> {
+  pub unsafe fn try_from_raw(ptr: *mut T) -> Option<Self> {
```
bnoordhuis, line 109:
```suggestion
  pub(crate) unsafe fn try_from_raw(ptr: *mut T) -> Option<Self> {
```
Or is it intentional to promote it to the public API?

bnoordhuis (commented):
Mostly LGTM, thanks!

src/support.rs
losfair: I think `try_from_raw` and `from_raw` should have the same visibility since `from_raw` is just a thin wrapper that unwraps the result of `try_from_raw`?

But instead of promoting `try_from_raw` to the public API I think making both functions `pub(crate)` might make more sense, since there is special requirement on the pointer (`T`'s own `Drop` implementation should free the pointer) which doesn't seem explicit, and none of the current APIs expect externally-constructed `UniqueRef`.

bnoordhuis: I agree, they're better off as `pub(crate)`.

losfair: Fixed

src/binding.cc
```diff
@@ -64,6 +64,17 @@ static_assert(sizeof(v8::CFunction) == sizeof(size_t) * 2,
...
+              "CachedData size mismatch");
+static_assert(offsetof(v8::ScriptCompiler::CachedData, data) == 0,
+              "CachedData.data offset mismatch");
+static_assert(offsetof(v8::ScriptCompiler::CachedData, length) == 8,
+              "CachedData.length offset mismatch");
+static_assert(offsetof(v8::ScriptCompiler::CachedData, rejected) == 12,
+              "CachedData.rejected offset mismatch");
+static_assert(offsetof(v8::ScriptCompiler::CachedData, buffer_policy) == 16,
+              "CachedData.buffer_policy offset mismatch");
+
```
bnoordhuis, line 77: Ideally this would use `sizeof(three_pointers_t) == sizeof(v8::ScriptCompiler::CachedData)` etc. so it also works with 32 bits builds but I'll leave that up to you because we don't actually test such builds.

losfair: Added check for 32-bit platforms.

ry (approved):
LGTM - thanks @losfair 

We'll probably wait another week or two before we make a release that includes this patch.

PR #640, opened 2021-03-06
build: patch GN's host_cpu detection on Apple Silicon

AaronO:
Since we download `x86_64` `ninja` and `gn` on M1 macs `gn` incorrectly targets `x86_64` by default.

This codifies [ry's workaround](https://github.com/denoland/rusty_v8/issues/520#issuecomment-740190036) so that `V8_FROM_SOURCE=1 cargo build --release` does the right thing out of the box, no surprises.

I hadn't checked the comments and all the threads, so I ended up "wasting" a decent amount of time building versions I couldn't link against, when wanting to test local changes with other crates. So hopefully this saves other contributors time.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -84,7 +84,7 @@ fn build_v8() {
   if !is_debug() {
     gn_args.push("v8_enable_handle_zapping=false".to_string());
   }
-  
+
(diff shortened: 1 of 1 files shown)


ry (approved):
LGTM - thanks!

PR #641, opened 2021-03-06
Initial support for external v8 strings

### LINKED ISSUE #636: Support external v8 strings
Bindings for v8's external strings would help make heap and performance improvements (reducing new allocs of `v8::String`s in some cases), such as https://github.com/denoland/deno/issues/9675.

Initially I would mainly be interested in binding to `&'static str`, that would be sufficient to drive potential perf improvements in `serde_v8` and the aforementioned heap improvements.

But external strings might also allow more efficient message-passing between isolates/workers.

## Implementation

This would mainly involve exposing bindings for `v8::String::MakeExternal` and possibly other snapshot related changes (to be investigated).

AaronO:
This PR partially fixes #636, by adding support for external static rust strings.

Rust doesn't have utf16 (or two-byte string) literals, so we could shorten the name in this static case, i.e:
```rust
v8::String::new_external_static()
instead of
v8::String::new_external_onebyte_static()
```

We could support non-static external strings too, but I'm not sure we want to encourage that.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -808,6 +808,27 @@ int v8__String__WriteUtf8(const v8::String& self, v8::Isolate* isolate,
   return self.WriteUtf8(isolate, buffer, length, nchars_ref, options);
 }
 
+class ExternalStaticOneByteStringResource: public v8::String::ExternalOneByteStringResource {
+  public:
+    ExternalStaticOneByteStringResource(const char *data, int length):
+                                        _data(data), _length(length) {}
+    const char* data() const override { return _data; }
+    size_t length() const override { return _length; }
+
+  private:
+    const char* _data;
+    const int _length;
+};
+
+const v8::String* v8__String__NewExternalOneByteStatic(v8::Isolate *isolate,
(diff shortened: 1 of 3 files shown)


ry (commented):
Looks good. Could you add a small test?

I'd prefer to keep the function name the same as in V8, as you have done.

AaronO:
@ry Added some simple tests and exposed some relevant v8 methods:
```
is_external()
is_external_onebyte()
is_external_twobyte()
is_onebyte()
```

See notes on `is_external()`, it needs to be updated on next v8 release, since `v8::String::IsExternal()` was only recently re-added (https://source.chromium.org/chromium/_/chromium/v8/v8.git/+/1dd8624b524d14076160c1743f7da0b20fbe68e0).

The C++ binding fallsback to `v8::Value::IsExternal()` which is incorrect, the rust fallback I added is perfectly functional however, so this can ship without updating v8

ry (approved):
LGTM - thanks @AaronO

PR #642, opened 2021-03-07
Cache should be updated when build or buildtools is changed

ry:


.github/workflows/ci.yml
```diff
@@ -110,7 +110,7 @@ jobs:
             target/*/deps
           key:
             ${{ matrix.config.target }}-${{ matrix.config.variant }}-${{
-            hashFiles('./Cargo.lock', './v8/include/v8-version.h') }}
+            hashFiles('./Cargo.lock', './v8/include/v8-version.h', './build/**', './buildtools/**', './third_party/**') }}
```
piscisaureus, line 113: This is a bit worrysome- are you sure that no temp files or other volatile artifacts (.pyc files for instance) are ever saved to those directories?
I’m also not convinced that I’ll work because a lot of the tooling is in “target” and “tools/clang”.
Maybe it would be better to only consider the submodule SHAs and build.rs.

That said , if you just want to try it out to see whether it works, I have no objection to that.

piscisaureus (approved):
Committer discretion is advised.

PR #648, opened 2021-03-16
Add ArrayBuffer::detach() and is_detachable()

### LINKED ISSUE #646: Add bindings for `ArrayBuffer::Detach` and friends
`ArrayBuffer::Detach` also requires `ArrayBuffer::Externalize` (because buf needs to have been externalized before detaching). `ArrayBuffer::IsDetachable` and `ArrayBuffer::IsExternal` would be very useful too.

bnoordhuis:
Fixes #646.


lucacasonato (approved):
LGTM! Thanks :-)

PR #649, opened 2021-03-17
feat: add new_backing_store_from_raw_pointer

lucacasonato:
This commit adds a `ArrayBuffer::new_backing_store_from_raw_pointer`
method that can be used to construct an externally backed
ArrayBuffer without any safety guarantees.

I am not too happy with the use of an `extern "C"` callback. It would
be great if the function took a rust `Fn(Box<T>)` callback, and the
deleter_data was a `Box<T>` (or `Box<dyn Any>`), and we just hooked
this all up correctly. It is also unfortunate that the callback does
not get passed a reference to the isolate.

I still need to add a test.

### DIFF
--- a/src/array_buffer.rs
+++ b/src/array_buffer.rs
@@ -418,4 +418,32 @@ impl ArrayBuffer {
       ))
     }
   }
+
+  /// Returns a new standalone BackingStore that is backed by the given buffer.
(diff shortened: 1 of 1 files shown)


lucacasonato:
Closing for now

PR #653, opened 2021-03-28
Remove dead platform code

bnoordhuis:
The Task struct and Platform::pump_message_loop() method were
non-functional, remove them.


build.rs
```diff
@@ -11,6 +11,8 @@ use std::process::Stdio;
 use which::which;
 
 fn main() {
+  println!("cargo:rerun-if-changed=src/binding.cc");
```
bnoordhuis, line 14: This is a minor bug fix. Previously, changes to binding.cc wouldn't always get picked up.

ry (approved):
LGTM

PR #654, opened 2021-04-01
Add String::new_from_one_byte

ry:


src/string.rs
```diff
@@ -20,6 +20,13 @@ extern "C" {
     length: int,
   ) -> *const String;
 
+  fn v8__String__NewFromOneByte(
+    isolate: *mut Isolate,
+    data: *const char,
```
bnoordhuis, line 25:
```suggestion
    data: *const u8,
```
And `const uint8_t* data` in C++ land so you don't have to reinterpret_cast.

bnoordhuis (approved):

PR #657, opened 2021-04-08
Upgrade V8 to 9.1.269.2

ry:


bnoordhuis:
> js-heap-broker.cc:4515: undefined reference to `__aarch64_cas4_acq_rel'

Needs to link in libatomic - which I thought the aarch64 build already did. I'll investigate.

piscisaureus:
@bnoordhuis 

> Needs to link in libatomic - which I thought the aarch64 build already did. I'll investigate.
> 
> edit: It did but the cross-compile toolchain is too old.

It has something to do with clang emitting outline intrinsics which are are not supported by our runtime library. Next thing I would suggest to try is adding `-mno-outline` like here: https://github.com/denoland/chromium_build/blob/4e13e3f7dd3689818d4662e9c853fa75df77f92a/config/compiler/BUILD.gn#L1097. Linking with `libgcc` instead of `compiler-rt` might also work.

However, as you said, this toolchain is way old and I question whether it's worth it sinking all this time into it.

piscisaureus (approved):

PR #666, opened 2021-04-19
Add Signature::new()

bnoordhuis:
Refs #478.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1485,8 +1485,8 @@ const v8::Object* v8__Function__NewInstance(const v8::Function& self,
 }
 
 const v8::Signature* v8__Signature__New(v8::Isolate* isolate,
-                                        const v8::FunctionTemplate& templ) {
-  return local_to_ptr(v8::Signature::New(isolate, ptr_to_local(&templ)));
+                                        const v8::FunctionTemplate* templ) {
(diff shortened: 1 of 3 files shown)


src/binding.cc
```diff
@@ -1484,6 +1484,11 @@ const v8::Object* v8__Function__NewInstance(const v8::Function& self,
       ptr_to_local(&context), argc, const_ptr_array_to_local_array(argv)));
 }
 
+const v8::Signature* v8__Signature__New(v8::Isolate* isolate,
+                                        const v8::FunctionTemplate& templ) {
```
piscisaureus, line 1488: Prefer the simple C type `const v8::FunctionTemplate*'.

piscisaureus (approved):
LGTM but see comment

src/binding.cc
```diff
@@ -1484,6 +1484,11 @@ const v8::Object* v8__Function__NewInstance(const v8::Function& self,
...
+  return local_to_ptr(v8::Signature::New(isolate, ptr_to_local(&templ)));
```
bnoordhuis, line 1489:
```suggestion
                                        const v8::FunctionTemplate* templ) {
  return local_to_ptr(v8::Signature::New(isolate, ptr_to_local(templ)));
```
Done, but I'll note that rusty_v8 is pretty schizophrenic when it comes to pointers-vs-references. Maybe I'll open a PR to switch everything over to pointers.

Issue #586, opened 2021-01-18
Dynamically link to CRT on Windows


ozgrakkurt:
Use the /MD flag on windows. I think the only build is statically linked to CRT which causes link issues with other libraries. Should't it be the default to use /MD?

piscisaureus:
As far as I know librusty_v8.lib is neither linked statically nor dynamically, unless a bug snuck in.

What are the linking issues you are referring to?

PR #690, opened 2021-05-11
build: add flag to build CRT flavor agnostic Windows libraries

piscisaureus:
Fixes: #586
Fixes: denoland/deno#10378

### DIFF
--- a/.gn
+++ b/.gn
@@ -32,5 +32,6 @@ default_args = {
   v8_use_external_startup_data = false
   v8_use_snapshot = true
   is_component_build = false
+  win_crt_flavor_agnostic = true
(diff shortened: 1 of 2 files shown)


bnoordhuis (approved):
LGTM. Do you plan on upstreaming the patch to build/?

piscisaureus:
> Do you plan on upstreaming the patch to build/?

No. I'd do it if I thought they were interested, but (wrt `//build` specifically) they just land what google/chromium needs and only what they need, and they don't distribute any static libraries. On our end, we're not getting out of maintaining a fork anytime soon.  So all in all it seems not worth the effort.

PR #270, opened 2020-02-09
android support 64 bit

90dy:
see https://chromium.googlesource.com/chromium/src/+/master/docs/android_build_instructions.md
ref denoland/deno#3839

Successful build on aarch-linux-android and x86_64-linux-android 🙌 
Other arch (arm and i686) are not currently available

# How to

### Clone the repo
```
git clone git@github.com:denoland/rusty_v8 ./rusty_v8
cd ./rusty_v8
git submodule update --init --recursive
```

### Install cross

```
cargo install cross
```

### Build aarch64
```
docker build --target aarch64-linux-android --tag denoland/rusty_v8:aarch64-linux-android .
V8_FROM_SOURCE=1 cross build -vv --target aarch64-linux-android
```

### Build x86_64
```
docker build --target x86_64-linux-android --tag denoland/rusty_v8:x86_64-linux-android .
V8_FROM_SOURCE=1 cross build -vv --target x86_64-linux-android
```

PR #695, opened 2021-05-31
Adding explicit return statement in build.rs line:157

beastOP:
It caused my deno installation build to fail.
Screenshot_2021-05-31-13-36-56-12_84d3000e3f4017145260f7618db1d683

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -155,18 +155,20 @@ fn maybe_install_sysroot(arch: &str) {
 }
 
 fn platform() -> &'static str {
+  let target_os;
   #[cfg(target_os = "windows")]
   {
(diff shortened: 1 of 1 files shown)


lucacasonato:
This is not going to solve the problem of not being able to run on android. We do not provide V8 builds for android. See #270

beastOP:
@lucacasonato it might not solve my problem but it can cause others build to fail as it is a valid rust error, right?

lucacasonato:
No - this error will only appear on unsupported architechtures.

piscisaureus:
Yeah this fixes nothing.

Issue #694, opened 2021-05-26
Unable to upgrade to V8 9.2


ry:
During `gn gen` we're getting an error:
```
ERROR at //gni/v8.gni:89:25: Duplicate build argument declaration.
  cppgc_is_standalone = false
                        ^----
Here you're declaring an argument that was already declared elsewhere.
You can only declare each argument once in the entire build so there is one
canonical place for documentation and the default value. Either move this
argument to the build config file (for visibility everywhere) or to a .gni file
that you "import" from the files where you need it (preferred).
See //v8/gni/v8.gni:89:25: Previous declaration.
  cppgc_is_standalone = false
                        ^----
See also "gn help buildargs" for more on how build arguments work.
See //third_party/googletest/BUILD.gn:5:1: whence it was imported.
import("../../gni/v8.gni")
^------------------------
See //testing/gmock/BUILD.gn:21:5: which caused the file to be included.
    "//third_party/googletest:gmock_config",
    ^--------------------------------------
```
The problem seems to be that we're importing v8.gni twice. The `import("../../gni/v8.gni")` was in `v8/third_party/googletest/BUILD.gn` was recently added 7f9d7f0e40c57a15a4c052f92321f1508cd7d917 (https://chromium-review.googlesource.com/c/v8/v8/+/2850700)

I suspect this problem is because we use V8 as a secondary_source and somehow GN doesn't recognize that v8.gni was already imported?
https://github.com/denoland/rusty_v8/blob/be1c2073ec1675970b4358016dc8a98ce5356f97/.gn#L14-L17

When I revert 7f9d7f0e40c57a15a4c052f92321f1508cd7d917, the build works correctly.

I'm not sure how we can workaround this. 

cc @MayaLekova

MayaLekova:
Sorry for the error you're experiencing. I'm curious though why other BUILD.gn files that do `import("../../gni/v8.gni")` don't cause the same issue, such as `v8/tools/BUILD.gn` for instance. I agree with your analysis that probably using `secondary_source` triggers the issue, as normally importing the same .gni file twice shouldn't have any effect.

Is it a viable temporary solution to have a floating patch with the faulty CL reverted? In the meantime, I'm trying to find a person more familiar with GN to give us a hint what would be a proper way to solve this (couldn't find any public forum about GN usage).

ry:
@MayaLekova Thank you for the reply and asking around for us. Yes, if necessary, we can float the revert patch - but that causes other complications.

[...]

PR #700, opened 2021-06-10
Upgrade V8 to 9.2.230.10

ry:
We are floating a revert of v8/v8@7f9d7f0 because of the issue described
in #694.


piscisaureus:
Disabling shared readonly heap because of a faulty assert in debug mode: 😢

Upgrading at all: 🎉🎉🎉

piscisaureus (approved):

PR #704, opened 2021-06-17
Add missing methods for one- and two-byte strings.

### LINKED ISSUE #703: Support missing methods for working with one- and two-byte strings
rusty_v8 currently doesn't support any of `v8::String`'s methods for creating strings from a two-byte (UTF-16) source, or for retrieving a string's contents as either one-byte (Latin-1) or two-byte representations. Working with the Latin-1 representation of strings seems to be necessary for denoland/serde_v8#1, and the UTF-16 representation is expected to be useful for some of the performance improvements in denoland/deno#10978.

The missing APIs in question are:

- `v8::String::NewFromTwoByte` (static)
- `v8::String::NewExternalTwoByte` (static)
- `v8::String::ContainsOnlyOneByte`
- `v8::String::Write`
- `v8::String::WriteOneByte`

andreubotella:
Fixes #703.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -852,10 +852,30 @@ const v8::String* v8__String__NewFromOneByte(v8::Isolate* isolate,
       v8::String::NewFromOneByte(isolate, data, new_type, length));
 }
 
+const v8::String* v8__String__NewFromTwoByte(v8::Isolate* isolate,
+                                             const uint16_t* data,
+                                             v8::NewStringType new_type,
+                                             int length) {
+  return maybe_local_to_ptr(
+      v8::String::NewFromTwoByte(isolate, data, new_type, length));
+}
+
 int v8__String__Length(const v8::String& self) { return self.Length(); }
 
 int v8__String__Utf8Length(const v8::String& self, v8::Isolate* isolate) {
   return self.Utf8Length(isolate);
+}
+
+int v8__String__Write(const v8::String& self, v8::Isolate* isolate,
+                          uint16_t* buffer, int start, int length,
+                          int options) {
+  return self.Write(isolate, buffer, start, length, options);
+}
+
+int v8__String__WriteOneByte(const v8::String& self, v8::Isolate* isolate,
+                          uint8_t* buffer, int start, int length,
+                          int options) {
+  return self.WriteOneByte(isolate, buffer, start, length, options);
 }
 
 int v8__String__WriteUtf8(const v8::String& self, v8::Isolate* isolate,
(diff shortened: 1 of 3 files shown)


andreubotella:
The Windows build is failing with a `STATUS_ACCESS_VIOLATION` error, but I can't reproduce on a local Windows 10 machine. It doesn't seem like I can rerun the workflow manually, so I'll push an empty commit instead to see if it was a fluke.

src/string.rs
```diff
@@ -191,6 +285,23 @@ impl String {
...
+  pub fn new_external_twobyte_static<'s>(
+    scope: &mut HandleScope<'s, ()>,
+    buffer: &'static [u16],
+  ) -> Option<Local<'s, String>> {
+    let buffer_len = buffer.len().try_into().ok()?;
+    unsafe {
+      scope.cast_local(|sd| {
+        v8__String__NewExternalTwoByteStatic(
+          sd.get_isolate_ptr(),
+          buffer.as_ptr() as *const u16,
```
bnoordhuis, line 298:
```suggestion
          buffer.as_ptr(),
```
The cast is superfluous, I think?

src/string.rs
```diff
@@ -130,6 +162,25 @@ impl String {
...
+    buffer: &[u16],
+    new_type: NewStringType,
+  ) -> Option<Local<'s, String>> {
+    unsafe {
+      scope.cast_local(|sd| {
+        v8__String__NewFromTwoByte(
+          sd.get_isolate_ptr(),
+          buffer.as_ptr(),
+          new_type,
+          buffer.len() as int,
```
bnoordhuis, line 178: I realize you copied this from `new_from_one_byte()` but a good follow-up change would be to turn the cast into a fallible operation, in case someone passes in a buffer > 2^31.

src/binding.cc
```diff
@@ -852,12 +852,32 @@ const v8::String* v8__String__NewFromOneByte(v8::Isolate* isolate,
...
+
 int v8__String__Length(const v8::String& self) { return self.Length(); }
 
 int v8__String__Utf8Length(const v8::String& self, v8::Isolate* isolate) {
   return self.Utf8Length(isolate);
 }
 
+int v8__String__Write(const v8::String& self, v8::Isolate* isolate,
+                          uint16_t* buffer, int start, int length,
+                          int options) {
```
bnoordhuis, line 871:
```suggestion
int v8__String__Write(const v8::String& self, v8::Isolate* isolate,
                      uint16_t* buffer, int start, int length,
                      int options) {
```

src/binding.cc
```diff
@@ -852,12 +852,32 @@ const v8::String* v8__String__NewFromOneByte(v8::Isolate* isolate,
...
+  return self.Write(isolate, buffer, start, length, options);
+}
+
+int v8__String__WriteOneByte(const v8::String& self, v8::Isolate* isolate,
+                          uint8_t* buffer, int start, int length,
+                          int options) {
```
bnoordhuis, line 877:
```suggestion
int v8__String__WriteOneByte(const v8::String& self, v8::Isolate* isolate,
                             uint8_t* buffer, int start, int length,
                             int options) {
```

bnoordhuis (approved):
LGTM, thanks. I'll fix up the style suggestions.

PR #706, opened 2021-06-20
platform: pump_message_loop, run_idle_tasks, GLOBAL_PLATFORM

inteon:
Added `pump_message_loop` and `run_idle_tasks` to `v8::Platform`.
The `GLOBAL_PLATFORM` now manages the active platform.

### DIFF
--- a/examples/hello_world.rs
+++ b/examples/hello_world.rs
@@ -2,7 +2,7 @@ use rusty_v8 as v8;
 
 fn main() {
   // Initialize V8.
-  let platform = v8::new_default_platform().unwrap();
+  let platform = v8::new_default_platform(0, false).make_shared();
   v8::V8::initialize_platform(platform);
   v8::V8::initialize();
 
--- a/src/V8.rs
+++ b/src/V8.rs
@@ -79,7 +79,7 @@ pub fn assert_initialized() {
   let global_state_guard = GLOBAL_STATE.lock().unwrap();
   match *global_state_guard {
     Initialized(_) => {}
-    _ => panic!("Invalid state"),
+    _ => panic!("Invalid global state"),
   };
 }
 
--- a/src/V8.rs
+++ b/src/V8.rs
@@ -7,7 +7,7 @@ use std::sync::Mutex;
 use std::vec::Vec;
 
 use crate::platform::Platform;
-use crate::support::UniqueRef;
+use crate::support::SharedRef;
 use crate::support::UnitType;
 
 extern "C" {
@@ -61,24 +61,26 @@ where
   }
 }
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Debug)]
 enum GlobalState {
   Uninitialized,
-  PlatformInitialized,
-  Initialized,
-  Disposed,
+  PlatformInitialized(SharedRef<Platform>),
+  Initialized(SharedRef<Platform>),
+  Disposed(SharedRef<Platform>),
   PlatformShutdown,
 }
 use GlobalState::*;
 
 lazy_static! {
-  static ref GLOBAL_STATE: Mutex<GlobalState> =
-    Mutex::new(GlobalState::Uninitialized);
+  static ref GLOBAL_STATE: Mutex<GlobalState> = Mutex::new(Uninitialized);
 }
 
 pub fn assert_initialized() {
   let global_state_guard = GLOBAL_STATE.lock().unwrap();
-  assert_eq!(*global_state_guard, Initialized);
+  match *global_state_guard {
+    Initialized(_) => {}
+    _ => panic!("Invalid state"),
+  };
 }
 
 /// Pass the command line arguments to v8.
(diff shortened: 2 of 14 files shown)


bartlomieju:
@inteon very nice! The code looks mostly good to me, however a segfault happens in `test_single_threaded_default_platform` test. It's an already existing test so my hunch is that some default values have changed in this PR.

src/platform.rs
```diff
@@ -1,31 +1,101 @@
...
+      std::mem::transmute::<_, Pin<Arc<Platform>>>(platform)
+    }
   }
 }
 
-#[repr(C)]
-#[derive(Debug)]
-pub struct Platform(Opaque);
+impl Platform {
+  pub fn pump_message_loop(
```
ry, line 66: please copy docs from V8
https://source.chromium.org/chromium/chromium/src/+/main:v8/include/libplatform/libplatform.h;l=67-76;drc=a18a674bb8423d0eec98e3d3933f648a7029d79d

Cargo.toml
```diff
@@ -73,6 +73,7 @@ exclude = [
 lazy_static = "1.4.0"
 libc = "0.2.93"
 bitflags = "1.2.1"
+num_cpus = "1.13.0"
```
bartlomieju, line 76: I suggest to skip this dependency and force users to pass the value themselves

inteon: for the tests, I'll just put 2 as a default value?

bartlomieju: Sure, or maybe just use single-threaded version anyway?

inteon: Ah, wait; don't you have to add the `--single-threaded` V8 flag everywhere then too?

ry: two is fine for the tests

src/V8.rs
```diff
@@ -173,16 +175,29 @@ pub fn get_version() -> &'static str {
...
+    unsafe { v8__V8__InitializePlatform(Arc::as_ptr(&unpinned_arc) as *mut _) };
+  }
+
   *global_state_guard = PlatformInitialized;
+  *global_platform_guard = Some(platform);
+}
+
+/// Sets the v8::Platform to use. This should be invoked before V8 is
+/// initialized.
+pub fn get_current_platform() -> Pin<Arc<Platform>> {
```
bartlomieju, line 196: I'm wondering if `Pin<Arc<Platform>>` could be simplified? @piscisaureus what do you think?

src/binding.cc
```diff
@@ -1967,17 +1968,33 @@ v8::StartupData v8__SnapshotCreator__CreateBlob(
...
 
-v8::Platform* v8__platform__NewDefaultPlatform() {
-  // TODO(bnoordhuis) Support optional arguments.
-  return v8::platform::NewDefaultPlatform().release();
+void v8__Platform__CONSTRUCT(
+  uninit_t<v8::platform::DefaultPlatform>* buf,
+  int thread_pool_size,
+  bool idle_task_support
+) {
+  static_assert(sizeof(v8::platform::DefaultPlatform) <= 20 * sizeof(size_t), "Platform size mismatch");
```
bnoordhuis, line 1976: Can you wrap at 80 columns in this file? Thanks.

bnoordhuis (commented):
Thanks for the pull request, @inteon.

I like the general thrust but I have mixed feelings about bypassing V8's public API and using the internal `DefaultPlatform` directly.

I assume you're doing that so you can use placement new but is that really necessary? Is the current `UniquePtr<Platform>` approach not workable?

src/binding.cc
ry: We don't check it in CI but this is formatted with clang-format 

https://github.com/denoland/rusty_v8/blob/main/.clang-format

src/V8.rs
piscisaureus: I don't think `Pin<Arc<_>>` is necessarily inappropriate, but it doesn't look like the reference count will ever exceed 1.

src/V8.rs
```diff
@@ -72,8 +73,9 @@ enum GlobalState {
 use GlobalState::*;
 
 lazy_static! {
-  static ref GLOBAL_STATE: Mutex<GlobalState> =
-    Mutex::new(GlobalState::Uninitialized);
+  static ref GLOBAL_STATE: Mutex<GlobalState> = Mutex::new(Uninitialized);
+  static ref GLOBAL_PLATFORM: Mutex<Option<Pin<Arc<Platform>>>> =
```
piscisaureus, line 77: Why is this an Arc? The reference count never exceeds one by the look of it.

inteon:
@bnoordhuis @piscisaureus @bartlomieju The main challenge I encountered implementing this PR, is the `pump_message_loop` function that needs a pointer to platform as a function argument. This means that unique ownership (like `UniquePtr<Platform>`) cannot be used, because the call to `initialize_platform` would transfer ownership, and the platform pointer is still needed for the `pump_message_loop` call.

For that reason, I'm using `Pin<Arc<Platform>>`. This however means that I have to allocate the heap space in rust and do a v8 in-place construction of the Platform object. This also means that the logic for the number of threads in the thread pool is skipped. Therefore this logic has to be implemented in rust (if thread_pool_size = 0 for call to `new_default_platform` instead the number of cpu cores - 1 should be used as the thread_pool_size; this is the logic from v8).

@piscisaureus I think the reference counter for the Arc increases when calling `get_current_platform` or when cloning the Arc. I could be wrong, but I think that the reference count will exceed 1, right?

examples/hello_world.rs
```diff
@@ -2,7 +2,7 @@ use rusty_v8 as v8;
 
 fn main() {
   // Initialize V8.
-  let platform = v8::new_default_platform().unwrap();
+  let platform = v8::new_default_platform(2, false);
```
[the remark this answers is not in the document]
inteon: As described in https://github.com/denoland/rusty_v8/pull/706#issuecomment-868713451, 0 does not translate to the number of cpu cores -1. This cpp logic had to be omitted so I van do and inplace cpp construct.

src/V8.rs
inteon: I think the counter does exceed 1, when returnining the Arc, it is closed and this the count is incremented.

inteon: Also, Arc allows to call Drop on platform once the V8::global_state_guard is called and no other references (Arc) exist anymore. (Fixes this comment: https://github.com/denoland/rusty_v8/pull/706/files#diff-5968ad1a2c8f5b1977d1dc3776ad078298f49e52a8a7bc137ecb46d6a45fc8a9L176)

bnoordhuis:
@inteon Would `SharedPtr` work? They're immutable shared references but I'd say it's okay to bend the rules a little behind the scenes and convert to a mutable pointer at the Rust/C++ edge.

inteon:
> @inteon Would `SharedPtr` work? They're immutable shared references but I'd say it's okay to bend the rules a little behind the scenes and convert to a mutable pointer at the Rust/C++ edge.

For a `SharedPtr` to work, I think you have to implement the Shared trait, and that requires a shared pointer to be kept on the cpp side. I also started by trying that solution, but I think it is not a simpler/ more elegant solution.

inteon:
> @inteon Would `SharedPtr` work? They're immutable shared references but I'd say it's okay to bend the rules a little behind the scenes and convert to a mutable pointer at the Rust/C++ edge.

Just finished creating a solution based on `SharedPtr` 🥳 . It required a bit more code, but requires less hacks.

bartlomieju:
@inteon nice work! This looks much better to me. I'll defer to review from @bnoordhuis and @piscisaureus

piscisaureus:
LGTM
That said, I'm still a bit nervous about it - @bnoordhuis can I get your sign off as well?

src/V8.rs
```diff
@@ -61,24 +61,26 @@ where
...
-    Mutex::new(GlobalState::Uninitialized);
+  static ref GLOBAL_STATE: Mutex<GlobalState> = Mutex::new(Uninitialized);
 }
 
 pub fn assert_initialized() {
   let global_state_guard = GLOBAL_STATE.lock().unwrap();
-  assert_eq!(*global_state_guard, Initialized);
+  match *global_state_guard {
+    Initialized(_) => {}
+    _ => panic!("Invalid state"),
```
bnoordhuis, line 82:
```suggestion
    _ => panic!("Invalid global state"),
```
For consistency with the other error messages.

src/platform.rs
```diff
@@ -1,31 +1,181 @@
...
+  /// If |idle_task_support| is enabled then the platform will accept idle
+  /// tasks (IdleTasksEnabled will return true) and will rely on the embedder
+  /// calling v8::platform::RunIdleTasks to process the idle tasks.
+  pub fn new(
+    thread_pool_size: u32,
+    idle_task_support: bool,
+  ) -> UniqueRef<Self> {
+    unsafe {
+      UniqueRef::from_raw(v8__Platform__NewDefaultPlatform(
+        std::cmp::min(thread_pool_size, 16) as i32,
```
bnoordhuis, line 95:
```suggestion
        thread_pool_size.min(16) as i32,
```

bnoordhuis (approved):
LGTM, thanks @inteon. Ideally, the test wouldn't use natives syntax (not stable across V8 releases) but if this is the only reliable way to test then so be it.

piscisaureus (approved):

PR #708, opened 2021-06-21
build: apply 'win_crt_flavor_agnostic' flag to components built as static libraries

piscisaureus:
Depends on #709

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -307,8 +307,14 @@ fn print_link_flags() {
   println!("cargo:rustc-link-lib=static=rusty_v8");
 
   if cfg!(target_os = "windows") {
-    println!("cargo:rustc-link-lib=dylib=winmm");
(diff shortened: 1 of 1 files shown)


ry:
Seems like you're getting a link error.

Is this something that could be submitted upstream?

piscisaureus:
> Is this something that could be submitted upstream?

No it's a "merge fix" to make #690 compatible again with `//build`.

ry (approved):

piscisaureus:
Linking `oldnames.lib` (`println!("cargo:rustc-link-lib=dylib=oldnames")`) does work, but it seems that rustc somehow does not propagate this library, so if the PR is landed as-is Deno can no longer be built.

piscisaureus:
... so we are out of options and will probably have to fork v8 forever.

PR #712, opened 2021-06-24
fix: pass serializer helpers as mut refs

lucacasonato:
They are only useful in read_host_object and write_host_object if
the helpers are available as a mut ref.


ry:
Can you add a test for this?

bartlomieju (approved):

piscisaureus (approved):
LGTM

PR #717, opened 2021-06-27
add bindings for V8Inspector::exception_thrown

bartlomieju:


tests/test_api.rs
```diff
@@ -3538,6 +3542,66 @@ fn inspector_dispatch_protocol_message() {
...
+  let message_string_view = StringView::from(message);
+  let detailed_message = "detailed message".to_string();
+  let detailed_message = &detailed_message.into_bytes()[..];
+  let detailed_message_string_view = StringView::from(detailed_message);
+  let url = "file://exception.js".to_string();
+  let url = &url.into_bytes()[..];
+  let url_string_view = StringView::from(url);
+  let exception_msg = v8::String::new(&mut context_scope, "This is a test error").unwrap();
+  let exception = v8::Exception::error(&mut context_scope, exception_msg);
+  let stack_trace = v8::UniquePtr::from(V8StackTrace);
```
bnoordhuis, line 3586: That won't work - `V8StackTrace` can't be instantiated in Rust or C++, it's a virtual base class. Only V8 can hand them out.

You can get an instance from V8 through `V8Inspector::createStackTrace()` or `V8Inspector::captureStackTrace()`. The latter is straightforward to implement:
<details>

```diff
diff --git a/src/binding.cc b/src/binding.cc
index 176c95a..1246828 100644
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -2056,6 +2056,11 @@ void v8_inspector__V8Inspector__contextCreated(
       ptr_to_local(&context), contextGroupId, humanReadableName));
 }
 
+v8_inspector::V8StackTrace* v8_inspector__V8Inspector__captureStackTrace(
+    v8_inspector::V8Inspector* self, bool full_stack) {
+  return self->captureStackTrace(full_stack).release();
+}
+
 void v8_inspector__V8InspectorSession__DELETE(
     v8_inspector::V8InspectorSession* self) {
   delete self;
diff --git a/src/inspector.rs b/src/inspector.rs
index b6cccd3..7957d4a 100644
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -106,6 +106,10 @@ extern "C" {
     contextGroupId: int,
     humanReadableName: StringView,
   );
+  fn v8_inspector__V8Inspector__captureStackTrace(
+    this: *mut V8Inspector,
+    full_stack: bool,
+  ) -> *mut V8StackTrace;
 }
 
 #[no_mangle]
@@ -925,6 +929,17 @@ impl V8Inspector {
       )
     }
   }
+
+  pub fn capture_stack_trace(
+    &mut self,
+    full_stack: bool,
+  ) -> UniqueRef<V8StackTrace> {
+    unsafe {
+      UniqueRef::from_raw(v8_inspector__V8Inspector__captureStackTrace(
+        self, full_stack,
+      ))
+    }
+  }
 }
 
 impl Drop for V8Inspector {
```

</details>

bartlomieju: Thanks! I got stuck on this one and picked up something else :) that should get me of the ground

bartlomieju: I'm not really sure what's going on :/ failing on:
`thread 'inspector_exception_thrown' panicked at 'assertion failed: needs_drop::<T>()', src/support.rs:183:3`
I think this is coming from `UniqueRef` related to `V8StackTrace`

bartlomieju:
```
fn assert_unique_ptr_layout_compatible<U, T>() {
  // Assert that `U` (a `UniqueRef` or `UniquePtr`) has the same memory layout
  // as a raw C pointer.
  assert_eq!(size_of::<U>(), size_of::<*mut T>());
  assert_eq!(align_of::<U>(), align_of::<*mut T>());

  // Assert that `T` (probably) implements `Drop`. If it doesn't, a regular
  // reference should be used instead of UniquePtr/UniqueRef.
  assert!(needs_drop::<T>());
}
```

So I guess we need to use a regular reference instead of `UniqueRef` :P

src/binding.cc
```diff
@@ -2056,6 +2056,28 @@ void v8_inspector__V8Inspector__contextCreated(
...
 
+unsigned v8_inspector__V8Inspector__exceptionThrown(
+    v8_inspector::V8Inspector* self, const v8::Context& context,
+    v8_inspector::StringView message, const v8::Value& exception,
+    v8_inspector::StringView detailed_message, v8_inspector::StringView url,
+    unsigned line_number, unsigned column_number,
+    v8_inspector::V8StackTrace* stack_trace, int script_id) {
+  return self->exceptionThrown(ptr_to_local(&context), message,
+    ptr_to_local(&exception), detailed_message, url, line_number,
+    column_number, static_cast<std::unique_ptr<v8_inspector::V8StackTrace>>(stack_trace),
```
piscisaureus, line 2067: I'm afraid this "consumes" the `V8StackTrace` object.
Is it possible to call `exception_thrown()` in rust twice without crashing?

piscisaureus:
I don't think the `V8StackTrace` implementation in rust is sufficiently complete (it's only a stub at the moment). 
In particular, it should have its C++ destructor called when dropped, otherwise it leaks memory.
The type to use in rust is `UniqueRef<V8StackTrace>` rather than `NonNull<V8StackTrace>`.

tests/test_api.rs
piscisaureus: > So I guess we need to use a regular reference instead of `UniqueRef` :P

I'm afraid it is the other way around: `Drop` needs to be implemented for `V8StackStrace`.

src/inspector.rs
```diff
@@ -925,6 +950,58 @@ impl V8Inspector {
...
+      ))
+    }
+  }
+
+  pub fn create_stack_trace(
+    &mut self,
+    stack_trace: Local<StackTrace>,
+  ) -> UniqueRef<V8StackTrace> {
+    unsafe {
+      UniqueRef::from_raw(v8_inspector__V8Inspector__createStackTrace(
```
piscisaureus, line 999: In C++ `V8Inspector::createStackTrace` is actually returning `nullptr` it seems. 
You weren't catching that in a previous version of this patch because you were using `NonNull::new_unchecked()` so the null value wasn't caught.

src/inspector.rs
```diff
@@ -925,6 +950,58 @@ impl V8Inspector {
...
+        url,
+        line_number,
+        column_number,
+        stack_trace.into_raw(),
+        script_id,
+      )
+    }
+  }
+
+  pub fn capture_stack_trace(
```
piscisaureus, line 983: This function needs a test as well.

src/inspector.rs
bartlomieju: I see, but it only partially makes sense to me - after all in the test I'm getting a stack trace but it has zero frames. So in this case shouldn't we actually use `UniquePtr<V8StackTrace>` which can be `null`?

piscisaureus: > I see, but it only partially makes sense to me - after all in the test I'm getting a stack trace but it has zero frames. So in this case shouldn't we actually use `UniquePtr<V8StackTrace>` which can be `null`?

Yes it looks like we should be using `UniquePtr<V8StackTrace>`.

Issue #643, opened 2021-03-08
Reintroduce Locker API (multi-thread support)


bnoordhuis:
Lockers were removed in #272 but it's become clear they're necessary to access isolates from different threads and to make things like `--prof` work, that require an ambient isolate.

The straw-man proposal is something like this:

1. `Isolate` becomes `Send` but is basically an opaque object with no functionality.

2. Creating a `Locker` consumes or borrows the `Isolate` and gives you a `LocalIsolate` (name TBD) that's `!Send + !Sync`.

3. Dropping the locker gives you back the `Isolate`.

There's a complementary `Unlocker` API for yielding the isolate. It'd be nice to have but it might be hard to model in Rust.

Stretch goal: make `HandleScope::new()` work with both `Isolate` and `LocalIsolate` so downstream users by and large won't have to update their code?

Would fix #486, #639, and some others. Stands a good chance of obsoleting #602. Removes the need for 83052c45351768c2e8923d1bf557728e03fecbf6.

nicksrandall:
I would love to help out and work on this but I'd new to this repo so I'd need some guidance. Would step 1 be reverting #272 to re-introduce the locker api?

bnoordhuis:
@nicksrandall That sounds like a good first step, yes. PR welcome!

ry:
@nicksrandall Help is much appreciated.

I think we need to figure out what the API would be still.

I was thinking we should have an API that consumes an OwnedIsolate to provide a Send object (which can then be sent to a different thread) and then something that consumes the object to turn it back into an OwnedIsolate.

Maybe something like
```rust
fn lock(i: v8::OwnedIsolate) -> v8::LockedIsolate;
fn unlock(i: v8::LockedIsolate) -> v8::OwnedIsolate;
```

nicksrandall:
Thanks for the help, I'll get started on this. Just to make sure I understand, do we still want to use v8's locking mechanism defined [here](https://v8docs.nodesource.com/node-0.8/d1/d80/classv8_1_1_locker.html) or do we want to implement the lock entirely on our end?

@ry I'll trust your opinion on this but I was thinking that for the API we might be able to attach a `lock` method to the `v8::OwnedIsolate` struct and create a new struct `v8::LockedIsolate` which would have a corresponding `unlock` method.

Something like:

```rust
let locked = owned.lock();
let owned = locked.unlock();
```

Should I try for the method approach or would you prefer that I stick with adding new functions to the crate?

[...]

PR #696, opened 2021-06-04
Reintroduce Locker API for Isolate

nicksrandall:
WIP fix for #643

I'm very knew to V8 internals and only moderately comfortable with Rust so I'm open to any and all feedback -- I hope I'm not a burden on the maintainers (let me know if I am!).

## Open Questions
- [ ] Do we want to use v8's internal locker API or handle everything ourselves? 
- [ ] What would be a good way to test that this feature actually works as intended?

PR #738, opened 2021-07-15
Add SharedIsolate w/ Locker API

SpencerSharkey:
This PR includes v8's `Locker`/`Unlocker` APIs with a means to move Isolate references across threads via a `SharedIsolate` handle. A `LockedIsolate<'_>` (similar to an OwnedIsolate with a lifetime) can be borrowed from a `SharedIsolate`. When the borrowed Isolate reference is returned, the interior locker goes with it.

I based this work on the code and discussion in PR #696. This would fix issue #643 and a few others mentioned in that issue. The important takeaway is that V8 *needs* to be notified via the creation of Lockers if an Isolate is going to be entered on a different thread. This PR should enable embedders to start using Isolates across multiple threads.

### SharedIsolate
Shared isolates implement Send and can be moved across threads freely. When created, the native isolate is not locked and no Locker is present until the embedder calls `lock(&mut self) -> LockedIsolate<'_>`. SharedIsolates don't hold an Arc and do not implement Sync, allowing the embedder to ensure exclusive mutability how they see fit. SharedIsolates can be dereferenced like an IsolateHandle, as those methods are safe to run from any thread.

SharedIsolates can be converted from exiting OwnedIsolates, or you can create one with `Isolate::new_shared(params)`.

### LockedIsolate<'a>
A temporary reference to an isolate, with a lifetime. While the LockedIsolate is in scope, thanks to the rules of Rust, the parent SharedIsolate can not be moved or borrowed from. When the LockedIsolate is dropped, any scopes are reset, the isolate is exited, and the Locker is destroyed.

### Changes to OwnedIsolate
When created, an OwnedIsolate may create a Locker and maintain its lifecycle. There is an exception to this case: SnapshotCreator creates an OwnedIsolate to be used for bootstrapping a new Snapshot, and creating a Locker, in this case, causes issues. This is why the internal OwnedIsolate::new() has a flag for optionally creating the locker. 

### Conversions
Both OwnedIsolate and SharedIsolate (given it's not currently borrowed) may be converted from one another. Lockers and Isolate entries/exits are handled for the embedder in these cases. This led me down a path of adding some bespoke handling of the disposal of Isolates when dropped (bool flag on the struct indicating if we should dispose on drop), which I would like to clean up if possible.

### Scopes
I was sure to implement all the scope-based traits for Isolate/OwnedIsolate for the new LockedIsolate, which makes it transparent for use in creating new HandleScopes and whatnot.

### Examples
There are a few crude examples of the usage of these APIs in the tests. Here's the basic usage of SharedIsolate:
```rs
  // create a new shared isolate handle
  let mut shared_isolate = v8::Isolate::new_shared(Default::default());
  {
    // lock the isolate for this block
    let isolate = &mut shared_isolate.lock();
    // ...
  }
  // create a new thread
  std::thread::spawn(move || {
    // borrow the same isolate here and do stuff with it
    // an embedder may wrap the shared_isolate in some sort of mutex
    let isolate = &mut shared_isolate.lock();
    // ...
  })
```

## Todo/Concerns
- [ ] Is it ok to mark v8::Global as !Send? It seems to pass in my basic tests but would like to know if any nuanced use-cases will cause problems here.
- [ ] I bound the Unlocker API and it can be used from a LockedIsolate's scope. Its usage is unsafe as we can't ensure other entered Isolates aren't touched while the Unlocker is in-scope. 
  - [ ] Is there a way to implement the Unlocker safely?
  - [ ] Should it be pulled completely, as it doesn't provide much use in this context?
- [ ] Are SharedIsolate and LockedIsolate fitting names?
- [ ] What else should we add tests for?


ry:
@SpencerSharkey sorry for the delayed review - coming soon!

src/handle.rs
```diff
@@ -151,6 +151,8 @@ impl<T> Global<T> {
   }
 }
 
+unsafe impl<T> Send for Global<T> {}
```
ry, line 154: I don't think Global can safely be sent between threads?

src/isolate.rs
```diff
@@ -864,24 +884,227 @@ impl IsolateHandle {
...
+      v8__Locker__DESTRUCT(self);
+    }
+  }
+}
+
+/// The Unlocker object is intended for use in a long-running callback from V8, where you want to release the V8 lock for other threads to use.
+/// The v8::Locker is a recursive lock, i.e. you can lock more than once in a given thread.
+/// This can be useful if you have code that can be called either from code that holds the lock or from code that does not.
+/// The Unlocker is not recursive so you can not have several Unlockers on the stack at once, and you can not use an Unlocker in a thread that is not inside a Locker's scope.
+/// An unlocker will unlock several lockers if it has to and reinstate the correct depth of locking on its destruction
```
ry, line 927: nit: please wrap at 80 columns

src/isolate.rs
```diff
@@ -864,24 +884,227 @@ impl IsolateHandle {
...
+  }
+}
+
+impl From<OwnedIsolate> for SharedIsolate {
+  /// Performs a conversion from an OwnedIsolate to a SharedIsolate.
+  /// As the OwnedIsolate is dropped during conversion, active scopes and the native isolate are exited.
+  fn from(mut isolate: OwnedIsolate) -> Self {
+    // marking this as true will skip disposal of the native isolate, which is handed off to the SharedIsolate.
+    isolate.skip_disposal = true;
+    SharedIsolate::new(isolate.cxx_isolate.as_ptr())
```
ry, line 1065: This makes me wish we had LSAN working...

ry (commented):
Thank you for this work @SpencerSharkey. API looks good and great that it's backwards compatible.

I'm worried about Global being marked Send. I want to get reviews from @piscisaureus and @bnoordhuis

src/handle.rs
piscisaureus: I think there is no material difference between `Isolate` being `Send` and `Global<T>` being `Send`.
Any one of those already creates the possibility to access `Global` while the `Isolate` is accessed by another thread, which creates safety issues that need to be addressed.

The issue arises from calling the non-thread-safe `IsolateHandle::get_isolate_ptr()` method at these call sites:

* https://github.com/denoland/rusty_v8/blob/main/src/handle.rs#L337
* https://github.com/denoland/rusty_v8/blob/main/src/handle.rs#L273
* https://github.com/denoland/rusty_v8/blob/main/src/handle.rs#L164

I believe the only way to address this issue is to add a check to `get_isolate_ptr()` that verifies that the `Isolate` is locked by the current thread.

src/isolate.rs
```diff
@@ -864,24 +884,227 @@ impl IsolateHandle {
...
+    // Resetting to the root scope exits any zombie scopes so that the next LockedIsolate can cleanly re-enter a context.
+    ScopeData::get_root_mut(self);
+    unsafe { self.exit() }
+  }
+}
+
+/// A reference to an Isolate that may be sent between threads and borrowed with a Locker before entering V8.
+/// May be created by calling Isolate::new_shared(params), or converted via SharedIsolate::from(OwnedIsolate).
+#[derive(Debug)]
+pub struct SharedIsolate {
```
piscisaureus, line 1026: Ideally the `IsolateHandle` type (returned from `Isolate::thread_safe_handle()`) would be used for this.

SpencerSharkey: I played around with this idea originally. Because `IsolateHandle` is `Clone` we'd need to introduce a mutex around the isolate, and I tried to veer away from adding any locks (maybe the embedded would like to bring their own lock). Also, an `OwnedIsolate` is entered for its entire lifespan so we'd need a way to break out of that locker without dropping the isolate, which would probably need to introduce a breaking change.

It's worth noting that `SharedIsolate` impl's the same deref as the `IsolateHandle`, so the related methods can be accessed. I'd love to hear out and implement any ideas on this front.

nicksrandall:
Thanks for picking up where I left off @SpencerSharkey. This is great work!

src/isolate.rs
SpencerSharkey: My mistake, `SharedIsolate` was actually implementing deref for `Isolate`, which was no good. I've corrected it by making `SharedIsolate` deref as `IsolateHandle`, but that still doesn't solve the main issue here.

src/handle.rs
SpencerSharkey: Thanks for the insight, I think I have a better understanding of the handles now.. I attempted to resolve these safety issues in the latest commits.

It's unfortunate that slightly misusing a Global is a runtime assertion Specifically, dropping while outside the isolate scope.

However, as an embedder, storing and accessing Globals within the bounds of the Isolate's entered state (ie. using `set/get_slot`) largely prevents these occurrences (i think).

SpencerSharkey:
Some new misc. bindings that are useful when using isolates across multiple threads:

+ Added (crate) `Locker::is_locked(*mut Isolate) -> bool`
+ Added `IsolateHandle::is_locked(&self) -> bool`
+ Added `IsolateHandle::is_disposed(&self) -> bool`
+ Added `IsolateHandle::is_in_use(&self) -> bool`
+ Added `SharedIsolate::discard_thread_specific_metadata(&mut self) -> bool`

SpencerSharkey:
I've moved some stuff around, most importantly removed `SharedIsolate` altogether in place of an `IsolateHandle<UnlockedHandle>` implementation. `IsolateHandle` defaults to using the existing safe handle ThreadSafeHandle impl. I am also able to clean up the OwnedIsolate handle a bit. There are still no breaking changes to the API

Additionally, `LockedIsolate` now comes in two type-state flavors. `Entered` and `Locked`. The Unlocked handle provides functions to just simply `lock()` or lock-and-enter with `enter()`. I realized it's useful to signal to v8 you've locked the isolate to a thread but do not want to enter it, so this should enable that.

It's still a bit WIP, but in the end, it may provide a more cohesive API.

piscisaureus:
> It's still a bit WIP, but in the end, it may provide a more cohesive API.

Again, sorry for the delay. This update looks really good so far; I'll get it reviewed as soon as possible.

src/handle.rs
```diff
@@ -160,14 +173,15 @@ impl<T> Clone for Global<T> {
...
-      } else {
-        // Destroy the storage cell that contains the contents of this Global.
-        v8__Global__Reset(self.data.cast().as_ptr())
+    let HandleInfo { data, host } = self.get_handle_info();
+    match host {
+      // This `Global` handle is associated with an `Isolate` that has already
+      // been disposed.
+      HandleHost::DisposedIsolate => {}
+      HandleHost::UnlockedIsolate => {
+        panic!("Global dropped outside of locked isolate context")
```
piscisaureus, line 182: Not that I necessarily see a better way, but this seems unfortunate.

src/isolate.rs
piscisaureus: I'm really having trouble now keeping all the different Isolate types in my head and understanding the differences between them. I wonder if we could do something radically more simple where:
* Using lockers is no longer optional, but required.
* Isolates are reference counted. `Isolate` clones and `Global<_>` handles both establish a reference to the v8 isolate. The isolate is automatically disposed when the reference count reaches 0. Note that we are already kind of keeping a reference count in `IsolateAnnex`.

image

SpencerSharkey: I've started to converge toward what you've described in the latest commit. In this case, I've kept `IsolateHandle` and `Isolate` similar to main branch to avoid having to shuffle all the existing methods around. OwnedIsolate, LockedIsolate, SharedIsolate, etc. are now gone.

- `Isolate::new_handle()` returns an IsolateHandle that can be cloned/shared and locked. 
- `Isolate::new()` sort-of maintains the existing behavior of returning a `Locker` that will also dispose of the Isolate when dropped. We can ditch this idea and force the shared IsolateHandle path, but that adds a new line to every test case and is a wider breaking change.

The IsolateHandle is used to guard the Isolate reference count and dispose of it when dropped.

`Locker` can be obtained from an `IsolateHandle`, which derefs `Isolate` and can be used to create `HandleScope`. 
It has a zero-sized type parameter to allow the creator (`Isolate::new()` or `IsolateHandle::lock(&mut self)`) to create a new 'borrowed' Locker *or*  use an owned one (default), which lets us make the lifetime on Locker trivially optional.

It was relatively straightforward to clean up SnapshotCreator's `get_isolate()` too with this system.

src/handle.rs
SpencerSharkey: I added a disposal bin for dropped global's that `reset`s and drops the v8 references the next time the isolate is entered/exited or after the Isolate is dropped. I don't know if this is a better approach, but I thought it complements the general flexibility of the Global type. It removes the mentioned error and eventually cleans up the reference.

The disposal list is tucked away in the `IsolateAnnex`.

Additionally, the `Global` handles hold a `Weak<IsolateAnnex>` reference (vs. cloned IsolateHandle arc) so the Isolate doesn't get held alive by dangling handles. It uses this to enqueue itself to the disposal queue if the global was dropped outside of the locked isolate scope.

SpencerSharkey:
The latest stuff here adds some breaking changes to the API, like removing OwnedIsolate, making Locker a first-class citizen, and cleaning up the SnapshotCreator isolate creation/scopes. Overall it makes the addition of Lockers more seamless than my previous attempts.

I'll start a Deno branch to get it up to date with the latest ideas here, including the lockers/snapshotcreator isolate scope cleanups

nicksrandall:
> I'll start a Deno branch to get it up to date with the latest ideas here, including the lockers/snapshotcreator isolate scope cleanups

Is the plan to make it so that JSRuntime in deno_core is Send + Sync?

bartlomieju:
> > I'll start a Deno branch to get it up to date with the latest ideas here, including the lockers/snapshotcreator isolate scope cleanups
> 
> Is the plan to make it so that JSRuntime in deno_core is Send + Sync?

No, currently there are no plans to make `deno_core::JsRuntime` work across the threads.

SpencerSharkey:
Correct, we just want to make Deno build/pass with the new Locker pattern, not change any of its threading models.

bartlomieju:
> I'll start a Deno branch to get it up to date with the latest ideas here, including the lockers/snapshotcreator isolate scope cleanups

@SpencerSharkey let me know if you have any problem updating `deno_core` to this API

SpencerSharkey:
Sorry for my absence here- my priorities were shifted towards some other matters for the past week.

I'll be getting this and the Deno draft PR wrapped up hopefully over the next few days!

ry:
@SpencerSharkey We're happy to help getting this patch in as well. I haven't even taken a look at this patch right now - but from your perspective how close is this / what's left to do?

nicksrandall:
I also would be happy to help get this over the line if somebody can help me understand what is left to do.

lbguilherme:
I have been using this patch for a while and it seems to be working fine. Is there anything missing besides proper testing and a happy CI?

SpencerSharkey:
this PR was a bigger bite than I could chew at the time and the code has gotten a little stale, so i'll close this out for now.

discussion continues in #788 

Thank you for the time spent reviewing and brainstorming- I hope to return to this problem in some time.

PR #742, opened 2021-07-29
support armv7

somombo:
Related to https://github.com/denoland/deno/issues/2295 
(See also https://github.com/denoland/deno/issues/1846)

@piscisaureus following up on https://github.com/denoland/rusty_v8/issues/265#issuecomment-682326419, I would appreciate your help and guidance on this. 

/cc  @afinch7  @liushuyu

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,8 +49,17 @@ jobs:
             target: aarch64-unknown-linux-gnu
             variant: release
 
+          - os: ${{ github.repository == 'denoland/rusty_v8' && 'ubuntu-latest-xl' || 'ubuntu-18.04' }}
+            target: armv7-unknown-linux-gnueabihf
+            variant: debug
+
+          - os: ${{ github.repository == 'denoland/rusty_v8' && 'ubuntu-latest-xl' || 'ubuntu-18.04' }}
+            target: armv7-unknown-linux-gnueabihf
+            variant: release
+
     env:
       V8_FROM_SOURCE: true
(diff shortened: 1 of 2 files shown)


somombo:
So far, I have proceeded mostly by analogy, copying and altering what was done for arm64. 

However, on my fork, I can't seem to get many of the tests in test_api.rs to  pass. All other tests that are not in that particular file are passing ok.

somombo:
Here is my latest build with the error I'm reproducibly getting:
https://github.com/somombo/rusty_v8/runs/3190160275?check_suite_focus=true

As you can see, I'm able to `cargo build`, `cargo run` hello world example, pass all tests other than those in test_api.rs all just fine. 

However, when I try to run many of the tests in test_api.rs I get the following error
```
Fatal error in ../../../../v8/src/heap/read-only-spaces.cc, line 69
Check failed: read_only_blob_checksum_ == snapshot_checksum (<unprintable> vs. 4042857134).
```
My crud research so far, is pointing to that the error is because I set environment `GN_ARGS='v8_enable_shared_ro_heap=true'`
Which it told me that I needed to do for target armv7 or else it wouldn't even build let alone test. 
(to be precise, it said I need to set either `v8_enable_shared_ro_heap` or `v8_enable_pointer_compression` and I found that the latter is not supported on this machine type)

cc @nazrhom, nghiant2710

somombo:
@piscisaureus `debug armv7-unknown-linux-gnueabihf` job got cancelled when it was almost done.. but that's the main one i needed to see the results of

PR #747, opened 2021-08-07
chore: test String object serialization

bnoordhuis:
Serializing `new String("")` requires that the wire format header is
written, otherwise V8 assumes an incompatible legacy wire format when
deserializing.


lucacasonato (approved):
LGTM!

PR #752, opened 2021-08-12
fix: Linux i686 compilation and abi callback errors

PerfectLaugh:
This fixes compilation in i686 target in Linux, and several callbacks error caused by SystemV ABI differences between i686 and amd64.

Reference:
https://github.com/denoland/deno/pull/11653

### DIFF
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -152,13 +152,26 @@ pub type PrepareStackTraceCallback<'s> = extern "C" fn(
   Local<'s, Array>,
 ) -> *mut *const Value;
 
-// System V ABI: MaybeLocal<Value> returned in a register.
-#[cfg(not(target_os = "windows"))]
+// System V AMD64 ABI: MaybeLocal<Value> returned in a register.
+#[cfg(all(not(target_os = "windows"), target_pointer_width = "64"))]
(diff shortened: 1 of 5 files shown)


src/isolate.rs
```diff
@@ -152,14 +152,27 @@ pub type PrepareStackTraceCallback<'s> = extern "C" fn(
...
+pub struct PrepareStackTraceCallbackRet(*const Value);
+
+#[cfg(all(not(target_os = "windows"), target_pointer_width = "32"))]
+pub type PrepareStackTraceCallback<'s> =
+  extern "C" fn(
+    Local<'s, Context>,
+    Local<'s, Value>,
+    Local<'s, Array>,
+  ) -> PrepareStackTraceCallbackRet;
+
```
bnoordhuis, line 175: I think it should be possible to use PrepareStackTraceCallbackRet for both 32 and 64 bits. Word-sized structs are returned in registers so under the hood it compiles down to the same thing.

ry:
Thank you for the patch but without being able to test this in CI, we cannot accept complex changes like this. If you setup qemu in our CI and get this to build in a reasonable amount of time, we can consider this in the future.

PR #753, opened 2021-08-15
Upgrade V8 to 9.4.146.2

bnoordhuis:


src/binding.cc
```diff
@@ -2298,23 +2298,10 @@ const v8::Value* v8__Module__GetException(const v8::Module& self) {
...
-
 const v8::FixedArray* v8__Module__GetModuleRequests(const v8::Module& self) {
   return local_to_ptr(self.GetModuleRequests());
 }
 
-void v8__Module__GetModuleRequestLocation(const v8::Module& self, int i,
-                                          v8::Location* out) {
-  *out = self.GetModuleRequestLocation(i);
-}
-
```
bnoordhuis, line 2317: Removed for breaking the build. They used to be "deprecate soon" in v8.h and now they're really deprecated.

lucacasonato (approved):

PR #756, opened 2021-08-19
feat: implement counter_lookup_callback

bnoordhuis:
This is an objectively terrible API. The reason for implementing it unsafely is performance: some counters are updated _a lot_. Any overhead _will_ be felt.

### DIFF
--- a/src/isolate_create_params.rs
+++ b/src/isolate_create_params.rs
@@ -13,7 +13,12 @@ use std::convert::TryFrom;
 use std::iter::once;
 use std::mem::size_of;
 use std::mem::MaybeUninit;
+use std::os::raw::c_char;
 use std::ptr::null;
+
+/// Should return a pointer to memory that persists for the lifetime of the
+/// isolate.
+pub type CounterLookupCallback = extern "C" fn(name: *const c_char) -> *mut i32;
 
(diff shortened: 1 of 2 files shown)


src/isolate_create_params.rs
```diff
@@ -24,6 +29,16 @@ pub struct CreateParams {
 }
 
 impl CreateParams {
+  /// Enables the host application to provide a mechanism for recording
+  /// statistics counters.
+  pub fn counter_lookup_callback(
```
ry, line 34: should be `set_counter_function` ?

https://source.chromium.org/chromium/chromium/src/+/main:v8/include/v8.h;l=9336-9340;drc=9aaf664253063e5a28fcacd58a67106357c3eaed

bnoordhuis: That's the post-instantiation "set callback" method (which doesn't really work because you'll miss events.)

This is https://github.com/v8/v8/blob/a78cf618524ccf0cb5fb569f667f6e0bb6435a64/include/v8.h#L8337

lucacasonato (approved):
LGTM. Unfortunate that this leaks unsafety.

bnoordhuis:
I'll tackle histograms next.

PR #776, opened 2021-09-13
feat: Add `v8::CompiledWasmModule`

### LINKED ISSUE #759: Add bindings to serialise and deserialise WASM module objects
To support transferring WebAssembly modules between workers, I need following bindings:

- WasmModuleObject::GetCompiledModule
- WasmModuleObject::FromCompiledModule
- CompiledWasmModule

This carries with it the assumption that `CompiledWasmModule` is isolate agnostic and can be shared across threads. Is this correct? If not, I need `CompiledWasmModule::Serialize` too, and the reverse of that (what method is that?).

Ref https://github.com/denoland/deno/issues/11823

andreubotella:
`v8::CompiledWasmModule` is a representation of a compiled WebAssembly module, which can be shared by multiple `v8::WasmModuleObject`s.

Closes #759.

### DIFF
--- a/src/wasm.rs
+++ b/src/wasm.rs
@@ -4,10 +4,12 @@ use crate::function::FunctionCallbackArguments;
 use crate::function::FunctionCallbackInfo;
 use crate::scope::CallbackScope;
 use crate::scope::HandleScope;
+use crate::support::Opaque;
 use crate::support::UnitType;
 use crate::Isolate;
 use crate::Local;
 use crate::Value;
+use crate::WasmModuleObject;
 use std::ptr::null;
 use std::ptr::null_mut;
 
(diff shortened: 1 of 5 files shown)


andreubotella:
In order for `v8::CompiledWasmModule` to be useful for deno, it must impl `Send`, and a naive implementation (using shared code with `SharedArrayBufferStore`) would also need it to impl `Sync`. Not being too fluent in C++, I'm not sure how sound that would be.

src/wasm.rs
```diff
@@ -62,6 +64,72 @@ impl Drop for WasmStreaming {
...
+
+/// Wrapper around a compiled WebAssembly module, which is potentially shared by
+/// different WasmModuleObjects.
+pub struct CompiledWasmModule(*mut InternalCompiledWasmModule);
+
+impl CompiledWasmModule {
+  /**
+   * Get the (wasm-encoded) wire bytes that were used to compile this module.
+   */
+  pub fn get_wire_bytes_ref(&self) -> &[u8] {
```
ry, line 113: Use triple slash comment here

tests/test_api.rs
```diff
@@ -5408,3 +5408,65 @@ fn counter_lookup_callback() {
...
+
+    let foo_ab: v8::Local<v8::ArrayBuffer> =
+      eval(scope, "WebAssembly.Module.customSections(module, 'foo')[0]")
+        .unwrap()
+        .try_into()
+        .unwrap();
+    let foo_bs = foo_ab.get_backing_store();
+    let foo_section = unsafe {
+      std::slice::from_raw_parts(foo_bs.data() as *mut u8, foo_bs.byte_length())
+    };
```
ry, line 5469: Maybe just do the conversation to string in JavaScript?
```
let ab = WebAssembly.Module.customSections(module, 'foo')[0];
let ui8 = new Uint8Array(ab);
new TextDecoder().decode(ui8)
```

src/wasm.rs
```diff
@@ -62,6 +64,72 @@ impl Drop for WasmStreaming {
...
+    unsafe {
+      let ptr = v8__CompiledWasmModule__GetWireBytesRef(self.0, &mut len);
+      std::slice::from_raw_parts(ptr, len.try_into().unwrap())
+    }
+  }
+}
+
+// TODO(andreubotella): Safety???
+unsafe impl Send for CompiledWasmModule {}
+unsafe impl Sync for CompiledWasmModule {}
```
ry, line 125: I'm not sure about this either. If you don't need it, can we leave them out for now?

src/wasm.rs
```diff
@@ -62,6 +64,72 @@ impl Drop for WasmStreaming {
...
+        )
+      })
+    }
+  }
+
+  /**
+   * Get the compiled module for this module object. The compiled module can be
+   * shared by several module objects.
+   */
+  pub fn get_compiled_module(&self) -> CompiledWasmModule {
```
ry, line 90: triple-slash comment here and elsewhere.

src/wasm.rs
andreubotella: They're gonna be needed for denoland/deno#11823 though.

src/wasm.rs
andreubotella: My bad, I was being lazy and just directly copied those from `v8.h` 😅

tests/test_api.rs
andreubotella: `TextDecoder` is a web API, not a v8 built-in.

ry (approved):
LGTM - nice implementation - these are useful bindings to have!

PR #781, opened 2021-09-18
Throw DataCloneError if SAB cannot be cloned

bnoordhuis:
Best case, it produces serialized output that cannot be deserialized.

Worst case, it hits this assert in V8:

    # Fatal error in v8::FromJust
    # Maybe value is Nothing.


lucacasonato (approved):

ry:
It seems like this PR made main branch red?

PR #783, opened 2021-09-22
Default to linking against release v8 builds

AaronO:
Unless $V8_FORCE_DEBUG=true to speedup CI for deno main repo

Related to https://github.com/denoland/deno/pull/11884

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -214,7 +214,7 @@ fn static_lib_url() -> String {
   // Use v8 in release mode unless $V8_FORCE_DEBUG=true
   let profile = match env::var("V8_FORCE_DEBUG").unwrap_or_default() == "true" {
     true => "debug",
-    _ => "release"
+    _ => "release",
   };
   format!("{}/v{}/librusty_v8_{}_{}.a", base, version, profile, target)
 }
--- a/build.rs
+++ b/build.rs
@@ -220,10 +220,10 @@ fn static_lib_url() -> String {
 }
 
 fn env_bool(key: &str) -> bool {
-  match env::var(key).unwrap_or_default().as_str() {
-    "true" | "1" | "yes" => true,
-    _ => false,
-  }
+  matches!(
+    env::var(key).unwrap_or_default().as_str(),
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -206,14 +206,17 @@ fn static_lib_url() -> String {
...
-    // Note: we always use the release build on windows.
-    format!("{}/v{}/rusty_v8_release_{}.lib", base, version, target)
-  } else {
-    let profile = env::var("PROFILE").unwrap();
-    assert!(profile == "release" || profile == "debug");
-    format!("{}/v{}/librusty_v8_{}_{}.a", base, version, profile, target)
+    return format!("{}/v{}/rusty_v8_release_{}.lib", base, version, target);
   }
+  // Use v8 in release mode unless $V8_FORCE_DEBUG=true
+  let profile = match env::var("V8_FORCE_DEBUG").unwrap_or_default() == "true" {
```
ry, line 215: I know there are some who disagree but I'm of the opinion that binary env vars like this only need to be checked if they are set. That is we don't handle the case `V8_FORCE_DEBUG=0`, because no one would do that.. but we do easily handle `V8_FORCE_DEBUG=1` (common) as well as `V8_FORCE_DEBUG=true` (less common)
```rust
let profile = if env::var("V8_FORCE_DEBUG").is_some() { "debug" } else { "release "};
```

ry (commented):
A comment in the readme about this env var would be helpful

ry (approved):
preemptive approval (but I do think a comment in the readme would be useful)

build.rs
piscisaureus: I'm one of those people that think that "" and "0" and "false" should turn the flag off. 

But given that rusty_v8 is a Rust project, we should follow rust/cargo conventions.

bnoordhuis (approved):

build.rs
AaronO: This is handled, it now acts like a "standard" env boolean and matches `true|1|yes` as truthy, everything else is false

PR #786, opened 2021-09-22
feat: Add `WasmStreaming::set_url` and `CompiledWasmModule::source_url`

andreubotella:
Needed for denoland/deno#12151


ry (approved):
LGTM - thanks @andreubotella !

src/wasm.rs
```diff
@@ -112,6 +125,15 @@ impl CompiledWasmModule {
       std::slice::from_raw_parts(ptr, len.try_into().unwrap())
     }
   }
+
+  pub fn source_url(&self) -> &str {
```
piscisaureus, line 129: This returns a &str on the C++ heap with infinite lifetime. This is definitely unsafe - reverting this PR.

PR #788, opened 2021-09-22
Add Locker API

piscisaureus:
Fork of #738


piscisaureus:
@SpencerSharkey - I'm not trying to steal your thunder here, but I thought I'd try to move #738 forward a bit.

src/snapshot.rs
```diff
@@ -153,6 +163,30 @@ impl SnapshotCreator {
     }
   }
 
+  pub unsafe fn get_isolate_ptr(&self) -> *mut Isolate {
```
piscisaureus, line 166: Does not need to be unsafe.

src/snapshot.rs
```diff
@@ -153,6 +163,30 @@ impl SnapshotCreator {
...
+  }
+
+  /// Borrows the Isolate from the SnapshotCreator.
+  pub fn get_isolate(&mut self) -> Locker {
+    unsafe { Locker::new(self.get_isolate_ptr().as_mut().unwrap()) }
+  }
+
+  /// Borrows the Isolate from the SnapshotCreator, also returning a reference
+  /// to the SnapshotCreator.
+  pub fn get_isolate_and_handle(&mut self) -> (Locker, &mut SnapshotCreator) {
```
piscisaureus, line 177: This method is a bit weird, I suppose it is needed to make `SnapshotCreator::set_default_context(context: Local<Context>)` work?
I think it would be better to change the signature to `SnapshotCreator::set_default_context(context: Global<Context>)`, that way it could be called outside a `HandleScope`.

src/snapshot.rs
```diff
@@ -153,6 +163,30 @@ impl SnapshotCreator {
...
+  }
+
+  /// Creates a snapshot data blob.
+  /// This must not be called from within a handle scope.
+  pub fn into_blob(
+    mut self,
+    function_code_handling: FunctionCodeHandling,
+  ) -> Option<StartupData> {
+    self.create_blob(function_code_handling)
+  }
```
lbguilherme, line 188: Why was `into_blob` introduced instead of just using `create_blob`?

src/snapshot.rs
lbguilherme: I don't see the need for this method. The isolate can be obtained from `get_isolate` directly. And the snapshotcreator itself can be used for `set_default_context` just fine. The only requirement is that the `scope` must go out of scope (dropped) before `create_blob` is called. As far as I can tell, it can be removed.

src/snapshot.rs
lbguilherme: It looks like this method shouldn't be pub.

lbguilherme:
Bug: If an `IsolateHandle` is created with `v8::Isolate::new_handle`, the underlying isolate is never dropped because `drop_annex` is never called and the `Arc<IsolateAnnex>` has a cycle. Additionally, there is no way to drop the annex using only the public API.

merlinfuchs:
Really looking forward for this to be done. 👍  Sadly not experienced enough to help with this myself ...

src/isolate.rs
```diff
@@ -729,79 +751,140 @@ impl Isolate {
...
-  fn new(isolate: &Isolate) -> Self {
-    Self(isolate.get_annex_arc())
+  /// Creates an owned locker from the isolate, used in Isolate::new()
+  pub(crate) fn lock_as_owned(&mut self) -> Locker {
+    unsafe { Locker::new(self.get_isolate_ptr().as_mut().unwrap()) }
+  }
+
+  /// Borrows a new [`Locker`] reference for the Isolate. While Locked, the
+  /// embedder is free to interact with an Isolate on the current thread.
+  pub fn lock(&mut self) -> Locker {
```
merlinfuchs, line 865: Why does this require a mutual reference? It does not mutate anything and also doesn't require exclusive access because `IsolateHandle` can be cloned, right?

piscisaureus:
I just rebased this PR and fixed the tests again.

We are not merging this PR at this time because is has unaddressed safety issues and at least one memory leak.
In Deno and Deno Deploy we don't need to access isolates from different threads so it's hard to evaluate whether this PR addresses the need.

We would appreciate help from the community to get this PR into a shape where we can land it.

ry:
closing because stale.

PR #800, opened 2021-10-07
feat: v8::StackTrace::CurrentStackTrace() bindings

AaronO:
I originally added this binding to experiment with stacktrace stitching in `deno`, but as I suspected it's way too slow ...

Might be good to add for completeness sake or other use-cases ...

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1658,6 +1658,11 @@ const v8::Value* v8__ReturnValue__Get(const v8::ReturnValue<v8::Value>& self) {
   return local_to_ptr(self.Get());
 }
 
+// Note: StackTraceOptions is deprecated, kDetailed is always used
+const v8::StackTrace* v8__StackTrace__CurrentStackTrace(v8::Isolate* isolate, int frame_limit) {
+  return local_to_ptr(v8::StackTrace::CurrentStackTrace(isolate, frame_limit));
+}
+
 int v8__StackTrace__GetFrameCount(const v8::StackTrace& self) {
   return self.GetFrameCount();
 }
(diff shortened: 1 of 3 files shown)


src/binding.cc
```diff
@@ -1649,6 +1649,11 @@ const v8::Value* v8__ReturnValue__Get(const v8::ReturnValue<v8::Value>& self) {
   return local_to_ptr(self.Get());
 }
 
+// Note: StackTraceOptions is deprecated, kDetailed is always used
+const v8::StackTrace* v8__StackTrace__CurrentStackTrace(v8::Isolate* isolate, int frame_limit) {
```
bnoordhuis, line 1653:
```suggestion
const v8::StackTrace* v8__StackTrace__CurrentStackTrace(v8::Isolate* isolate,
                                                        int frame_limit) {
```

src/exception.rs
```diff
@@ -67,6 +71,18 @@ extern "C" {
 }
 
 impl StackTrace {
+  /// Grab a snapshot of the current JavaScript execution stack.
+  pub fn current_stack_trace<'s>(
+    scope: &mut HandleScope<'s>,
+    frame_limit: i32,
```
bnoordhuis, line 77: Consider using `usize` and casting to `i32`. V8 uses `int` extensively because Google and their C++ style guide but that doesn't really carry over to Rust land.

bnoordhuis (commented):
Useful addition but needs a test.

crowlKats:
Need this for https://github.com/denoland/deno/issues/1725

src/binding.cc
AaronO: Fixed

src/exception.rs
AaronO: Done

bartlomieju (approved):

PR #806, opened 2021-10-13
Rename Handle::get() to Handle::open() instead

piscisaureus:
This also adds a deprecated `Global::get()` method to avoid unnecessarily breaking a lot of users code.

cc @espindola


bartlomieju (approved):
LGTM

Issue #378, opened 2020-05-20
Build script does not rerun when env variables change


tiziano88:
I tried to switch `V8_FROM_SOURCE` on locally and was surprised that the build script did not actually re-run. This should be easy to fix by emitting [`cargo:rerun-if-env-changed=NAME`](https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorerun-if-env-changedname) lines as part of the build script, I can look into that if there is interest.

piscisaureus:
> I can look into that if there is interest.

I know it;'s very belated -- but yes that'd be very much appreciated.

ross-weir:
I'd love to have a shot at this if it's still desired. Thanks.

bnoordhuis:
@ross-weir Sure, go for it.

PR #807, opened 2021-10-19
Add missing cargo:rerun-if-env-changed

espindola:
In addition to the problem described in issue #378, we noticed link
error when updating rusty_v8 without doing a clean build. As far as I
can tell, there were two potential bugs:

* Cargo not being informed that it should run build.rs when
  CARGO_PKG_VERSION changed.

* Even if build.rs was rerun, it would not download a new package. It
  seems that build.rs should either verify that the right package was
  downloaded or should just trust cargo to not rerun it when not
  needed.

Fixes: #378

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -13,6 +13,35 @@ use which::which;
 fn main() {
   println!("cargo:rerun-if-changed=src/binding.cc");
 
+  // These are all the environment variables that we check. This is
+  // probably more than what is needed, but missing an important
+  // variable can lead to broken links when switching rusty_v8
+  // versions.
+  let envs = vec![
+    "CARGO",
+    "CARGO_MANIFEST_DIR",
+    "CARGO_PKG_VERSION",
+    "CCACHE",
+    "CLANG_BASE_PATH",
+    "DENO_TRYBUILD",
+    "DOCS_RS",
+    "GN",
+    "GN_ARGS",
+    "HOST",
+    "NINJA",
+    "OUT_DIR",
+    "PROFILE",
+    "RUSTY_V8_ARCHIVE",
+    "RUSTY_V8_MIRROR",
+    "SCCACHE",
+    "TARGET",
+    "V8_FORCE_DEBUG",
+    "V8_FROM_SOURCE",
+  ];
+  for env in envs {
+    println!("cargo:rerun-if-env-changed={}", env);
+  }
+
   // Detect if trybuild tests are being compiled.
   let is_trybuild = env::var_os("DENO_TRYBUILD").is_some();
 
@@ -313,11 +342,10 @@ fn download_static_lib_binaries() {
 
   let filename = static_lib_path();
   if filename.exists() {
-    println!("static lib already exists {}", filename.display());
-    println!("To re-download this file, it must be manually deleted.");
-  } else {
-    download_file(url, filename);
+    println!("Deleting old static lib {}", filename.display());
+    std::fs::remove_file(&filename).unwrap();
   }
+  download_file(url, filename);
 }
 
 fn print_link_flags() {


build.rs
```diff
@@ -310,11 +338,11 @@ fn download_static_lib_binaries() {
 
   let filename = static_lib_path();
   if filename.exists() {
-    println!("static lib already exists {}", filename.display());
+    println!("Deleting old static lib {}", filename.display());
     println!("To re-download this file, it must be manually deleted.");
```
douglascaetano, line 342: I believe this line can be removed, as the file is being deleted.

```suggestion
```


I think, though, that it's not a good idea to just delete the file and re-download it. There are cases when one wants to use a custom static lib, which is currently achieved by placing the custom file in the expected path. By deleting and re-downloading the file, this custom case is not covered anymore.

douglascaetano, line 342: I believe this line can be removed, as the file is being deleted.

```suggestion
```

I think, though, that it's not a good idea to just delete and re-download the file, as there may be cases where one may want to use a custom static lib separately compiled. This is achieved by placing the custom file in the expected path. If the file is deleted, this custom case is not covered anymore.

espindola: I have removed that line and added a new commit that hopefully supports the described use case.

andreubotella:
It seems like this PR resulted in builds of the `v8` crate having to download the library every time, regardless of whether the environment variables have changed since the last time. In particular, it adds a significant delay to running `cargo run` in the Deno repo immediately after running `cargo build` – which particularly affects the WPT test runner, since that calls `cargo build`.

espindola:
> It seems like this PR resulted in builds of the `v8` crate having to download the library every time, regardless of whether the environment variables have changed since the last time. In particular, it adds a significant delay to running `cargo run` in the Deno repo immediately after running `cargo build` – which particularly affects the WPT test runner, since that calls `cargo build`.

I will try to isolate the problem.

espindola:
> I will try to isolate the problem.
I can't reproduce the problem. I have tried executing on the deno repo:

    $ cargo build
    $ cargo run --version # warmup
    $ time cargo run --version

The results I go with various revisions were:

    be68b82eb43f8e6156bad562b292af11776d38b8: (upgrade to v8 0.33 which includes this change)
    cargo run -- --version  0.30s user 0.07s system 98% cpu 0.370 total

    09dd77c13edf59eb3748b5f21bd8b7141d822f61: (one before the upgrade)
    cargo run -- --version  0.31s user 0.08s system 99% cpu 0.389 total

    7c2abb9d579d13fa61339c24f8c39bc3c27c25db: (current main)
    cargo run -- --version  0.30s user 0.07s system 99% cpu 0.372 total

Could you please provide which commands I need to run to reproduce the issue? Maybe I need to have some environment variable set?

ry:
@andreubotella Could you look into your env vars and see if any are matching the above list and potentially changing? I haven't worked on rusty_v8 lately but I haven't heard anyone else complaining about this issue.

andreubotella:
I don't have any of the above environment variables set, but I noticed that after doing `cargo clean`, running `cargo run` after `cargo build` doesn't rebuild V8 – until I run `cargo run -- run -A --unstable ./tools/wpt.ts run`, which makes any later invocations of `cargo run` rebuild it. Can anyone reproduce?

espindola:
At 7c2abb9d579d13fa61339c24f8c39bc3c27c25db I got an error:

    Check file:///home/espindola/chisel/deno/tools/wpt.ts
    warning: Cannot load module "file:///home/espindola/chisel/deno/tools/wpt/runner.ts".
    at file:///home/espindola/chisel/deno/tools/wpt.ts:11:8
      If the source module contains only types, use `import type` and `export type` to import it instead.
    error: Uncaught SyntaxError: The requested module './wpt/runner.ts' does not provide an export named 'runSingleTest'
     runSingleTest,

At be68b82eb43f8e6156bad562b292af11776d38b8 I get

`error: Uncaught (in promise) NotFound: No such file or directory (os error 2), open '/home/espindola/chisel/deno/tools/wpt/manifest.json'
  const manifestText = Deno.readTextFileSync(MANIFEST_PATH);`

But it does run long enough to trigger the problem. Trying to run `cargo build` again will rebuild (download) v8.

Debugging.

06kellyjac:
Removing https://github.com/denoland/rusty_v8/blob/2a21f5abf9943f48295181c9c3054d6001e8cfd7/build.rs#L315-L318 breaks our deno build in nixpkgs since there is no network access at build time to keep the build hermetic

We separately grab the library in a Fixed Output Derivation (does have network access) and plug it in to the build

https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/web/deno/default.nix#L38-L51

Is there interest in maintaining a Bring-Your-Own library break-glass or do I need to stick to patching build.rs from now on?

espindola:
> Is there interest in maintaining a Bring-Your-Own library break-glass or do I need to stick to patching build.rs from now on?

Yes, a bring-your-own build should work with RUSTY_V8_ARCHIVE: https://github.com/denoland/rusty_v8/blob/main/README.md#the-rusty_v8_archive-environment-variable

06kellyjac:
Ahh awesome, I'll try that now. Thanks @espindola
I can also drop all the bash that copies it to target/x/gn_out/obj/ :rocket:

PR #817, opened 2021-10-28
refactor: Have `BackingStore::data` return `Option<NonNull<c_void>>`

andreubotella:
The pointer returned by `BackingStore::data` might be null if the backing store has zero length, but the return type `*mut c_void` does not force the user to consider this case. This change makes the return type `Option<NonNull<c_void>>`, which is semantically equivalent, but which forces users of the API to handle the `None` case.

This is a breaking API change.


bnoordhuis (approved):
Seems like a good additional safety measure.

PR #820, opened 2021-11-01
feat: Add bindings for WasmStreaming::set_compiled_module_bytes and CompiledWasmModule::serialize

12101111:
Add some bindings to serialize and load compiled WebAssembly module cache.

With these bindings, we can cache the compiling result of WebAssembly module: https://github.com/denoland/deno/issues/11918#issuecomment-955683827

Test is not working currently.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -2390,7 +2390,27 @@ const v8::FixedArray* v8__ModuleRequest__GetImportAssertions(
     const v8::ModuleRequest& self) {
   return local_to_ptr(self.GetImportAssertions());
 }
+}
 
+extern "C" {
+void Rust__WasmStreamingClient__DROP(void*);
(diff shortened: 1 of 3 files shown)


andreubotella:
I believe this would also let us wake the tokio context in Deno when the module is done compiling, which would make it no longer necessary to keep the event loop spinning while a wasm module is being compiled:

https://github.com/denoland/deno/blob/main/core/runtime.rs#L832-L834

What if the compilation fails, though? There doesn't seem to be a way to listen for that.

bnoordhuis:
@12101111 Is this ready for review? Or if not, do you plan to work on it?

12101111:
`WasmStreaming::set_client` somhow don't work, but I don't have time to investigate it.

PR #821, opened 2021-11-01
chore: Specify the rust toolchain version

espindola:
Tests in tests/compile_fail/ depend on a specific toolchain
version. For example, they currently fail with 1.58.0-nightly because
of extra warnings.


ry (commented):
I'm not a fan of having all these files in the root directory that need to be maintained. 

Isn't it enough to just say this is not compatible with nightly?

espindola:
> Isn't it enough to just say this is not compatible with nightly?

Maybe, but how does one do it? Also, nightly is just an example. The tests check the specific warnings that are produced, so it seems they do depend on a particular compiler version.

piscisaureus:
I think this is reasonable.

piscisaureus (approved):

PR #822, opened 2021-11-01
Make lifetime in FunctionCallback more restrictive

espindola:
This makes the callbacks themselves less restrictive. A test is added
the would fail to compile before.


piscisaureus (approved):
LGTM

PR #825, opened 2021-11-03
Ignore environment variables set by cargo

espindola:
This avoids constant rebuilds when alternating between running

$ cargo build
$ cargo run <something that runs cargo build>

Since the second command will see these environment variables set
during the logic deciding if build.rs should be rerun or not.

Since these are cargo specific variables, hopefully cargo already does
the right thing without our help.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -18,9 +18,6 @@ fn main() {
   // variable can lead to broken links when switching rusty_v8
   // versions.
   let envs = vec![
-    "CARGO",
-    "CARGO_MANIFEST_DIR",
-    "CARGO_PKG_VERSION",
     "CCACHE",
     "CLANG_BASE_PATH",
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -18,9 +18,6 @@ fn main() {
   // variable can lead to broken links when switching rusty_v8
   // versions.
   let envs = vec![
-    "CARGO",
-    "CARGO_MANIFEST_DIR",
-    "CARGO_PKG_VERSION",
```
lucacasonato, line 23: This one is actually necessary to ensure rebuilds on version changes.

lucacasonato: Actually, I guess cargo will already rebuild the buiild script on version changes, so disregard that.

ry (approved):
LGTM - let's try it

PR #843, opened 2021-11-30
Add bindings to `v8::Function::GetScript{Line,Column}Number`

sujayakar:
I'd like to use these methods in my application, so I added them! This is my first time contributing to `rusty_v8`, so let me know if there's anything I'm missing.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1566,8 +1566,7 @@ void v8__Function__SetName(const v8::Function& self, const v8::String& name) {
 }
 
 int v8__Function__GetScriptColumnNumber(const v8::Function& self) {
-  int ret = ptr_to_local(&self)->GetScriptColumnNumber();
-  return ret == v8::Function::kLineOffsetNotFound ? -1 : ret;
+  return ptr_to_local(&self)->GetScriptColumnNumber();
 }
 
 int v8__Function__GetScriptLineNumber(const v8::Function& self) {
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1570,8 +1570,7 @@ int v8__Function__GetScriptColumnNumber(const v8::Function& self) {
 }
 
 int v8__Function__GetScriptLineNumber(const v8::Function& self) {
-  int ret = ptr_to_local(&self)->GetScriptLineNumber();
-  return ret == v8::Function::kLineOffsetNotFound ? -1 : ret;
+  return ptr_to_local(&self)->GetScriptLineNumber();
 }
 
 const v8::Signature* v8__Signature__New(v8::Isolate* isolate,
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1565,6 +1565,16 @@ void v8__Function__SetName(const v8::Function& self, const v8::String& name) {
   return ptr_to_local(&self)->SetName(ptr_to_local(&name));
 }
 
+int v8__Function__GetScriptColumnNumber(const v8::Function& self) {
+  int ret = ptr_to_local(&self)->GetScriptColumnNumber();
+  return ret == v8::Function::kLineOffsetNotFound ? -1 : ret;
+}
+
+int v8__Function__GetScriptLineNumber(const v8::Function& self) {
+  int ret = ptr_to_local(&self)->GetScriptLineNumber();
+  return ret == v8::Function::kLineOffsetNotFound ? -1 : ret;
+}
+
 const v8::Signature* v8__Signature__New(v8::Isolate* isolate,
                                         const v8::FunctionTemplate* templ) {
   return local_to_ptr(v8::Signature::New(isolate, ptr_to_local(templ)));
--- a/src/function.rs
+++ b/src/function.rs
@@ -42,6 +42,8 @@ extern "C" {
   ) -> *const Object;
   fn v8__Function__GetName(this: *const Function) -> *const String;
   fn v8__Function__SetName(this: *const Function, name: *const String);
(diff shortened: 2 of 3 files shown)


tests/test_api.rs
```diff
@@ -1985,6 +1985,64 @@ fn function() {
...
+    let f_str = v8::String::new(scope, "f").unwrap();
+    let f_function_obj: v8::Local<v8::Function> = namespace_obj
+      .get(scope, f_str.into())
+      .unwrap()
+      .try_into()
+      .unwrap();
+    // The column number is zero-indexed and indicates the position of the end of the name.
+    assert_eq!(
+      f_function_obj.get_script_column_number(),
+      Some("export function f".len() as i32),
```
bartlomieju, line 2027: This assertion should use hardcoded column number

tests/test_api.rs
```diff
@@ -1985,6 +1985,64 @@ fn function() {
...
+
+    let g_str = v8::String::new(scope, "anotherFunctionG").unwrap();
+    let g_function_obj: v8::Local<v8::Function> = namespace_obj
+      .get(scope, g_str.into())
+      .unwrap()
+      .try_into()
+      .unwrap();
+    assert_eq!(
+      g_function_obj.get_script_column_number(),
+      Some("export function anotherFunctionG".len() as i32),
```
bartlomieju, line 2040: This one too

bartlomieju (commented):
Thank you for PR @sujayakar! It looks mostly good I have only nitpicks to assertions in test

bartlomieju (approved):
LGTM, thank you @sujayakar I'll wait for another review before landing

src/binding.cc
```diff
@@ -1565,6 +1565,16 @@ void v8__Function__SetName(const v8::Function& self, const v8::String& name) {
   return ptr_to_local(&self)->SetName(ptr_to_local(&name));
 }
 
+int v8__Function__GetScriptColumnNumber(const v8::Function& self) {
+  int ret = ptr_to_local(&self)->GetScriptColumnNumber();
+  return ret == v8::Function::kLineOffsetNotFound ? -1 : ret;
```
bnoordhuis, line 1570:
```suggestion
  return ptr_to_local(&self)->GetScriptColumnNumber();
```
kLineOffsetNotFound == -1 so the ternary is somewhat superfluous.

src/binding.cc
```diff
@@ -1565,6 +1565,16 @@ void v8__Function__SetName(const v8::Function& self, const v8::String& name) {
...
+}
+
+int v8__Function__GetScriptLineNumber(const v8::Function& self) {
+  int ret = ptr_to_local(&self)->GetScriptLineNumber();
+  return ret == v8::Function::kLineOffsetNotFound ? -1 : ret;
```
bnoordhuis, line 1575:
```suggestion
  return ptr_to_local(&self)->GetScriptLineNumber();
```

src/function.rs
```diff
@@ -460,4 +462,24 @@ impl Function {
...
+
+  /// Get the (zero-indexed) line number of the function's definition, if available.
+  pub fn get_script_line_number(&self) -> Option<int> {
+    let ret = unsafe { v8__Function__GetScriptLineNumber(self) };
+    if ret < 0 {
+      None
+    } else {
+      Some(ret)
+    }
+  }
```
bnoordhuis, line 484:
```suggestion
  pub fn get_script_column_number(&self) -> Option<u32> {
    let ret = unsafe { v8__Function__GetScriptColumnNumber(self) };
    (ret >= 0).then(|| ret as u32)
  }

  /// Get the (zero-indexed) line number of the function's definition, if available.
  pub fn get_script_line_number(&self) -> Option<u32> {
    let ret = unsafe { v8__Function__GetScriptLineNumber(self) };
    (ret >= 0).then(|| ret as u32)
```
`int` is `std::os::raw::c_int` and shouldn't leak out into the public API when it can be avoided.

I picked `u32` but, depending on what you do with the return value, `usize` might be more appropriate.

bnoordhuis (approved):
Left some suggestions but mostly LGTM. Thanks for the PR!

src/function.rs
```diff
@@ -460,4 +462,16 @@ impl Function {
   pub fn set_name(&self, name: Local<String>) {
     unsafe { v8__Function__SetName(self, &*name) }
   }
+
+  /// Get the (zero-indexed) column number of the function's definition, if available.
+  pub fn get_script_column_number(&self) -> Option<u32> {
+    let ret = unsafe { v8__Function__GetScriptColumnNumber(self) };
+    (ret >= 0).then(|| ret as u32)
```
sujayakar, line 469: also woah, TIL about `bool::then` :)

bnoordhuis:
Restarted the CI. Looks like the aarch64 buildbot hit a bug in qemu:
```
(process:17120): GLib-CRITICAL **: 18:55:59.804: g_hash_table_iter_next: assertion 'ri->version == ri->hash_table->version' failed**
ERROR:/build/qemu-KZtMUi/qemu-2.11+dfsg/qom/object.c:1645:object_get_canonical_path_component: code should not be reached
```
Hopefully a one-time thing.

bnoordhuis:
Thanks Sujay, I merged it with a slightly modified commit message (I try to keep the first line <= 50 columns for reasons of legibility in commands like `git log --oneline`.)

sujayakar:
great! could you cut a release at your earliest convenience so I can depend on it rather than my fork?

PR #851, opened 2021-12-11
fix: `backing_store_deleter_callback` drops a `Box<c_void>`, not a `Box<[u8]>`

### LINKED ISSUE #849: Backing store allocation is not properly deallocated and causes UB
`Box::new()` does not allocate for ZSTs and empty arrays (or basically, for any `T` where `size_of::<T> == 0`).
So when the destructor of a backing store (or ArrayBuffer?) that was previously created from a zero-sized array runs, it tries to deallocate 0x1 (placeholder pointer) which isn't a valid allocation.

Edit: @andreubotella mentioned in a comment that the backing store deleter callback drops the data as `Box<c_void>`, not a `Box<[u8]>`, which is UB for any slice, no matter if empty or not. 

MRE:
```rs
#![forbid(unsafe_code)]
use rusty_v8 as v8;

fn main() {
    let platform = v8::new_default_platform(0, false).make_shared();
    v8::V8::initialize();

    let mut scope = v8::HandleScope::new(&mut isolate);
    let context = v8::Context::new(&mut scope);
    let mut scope = v8::ContextScope::new(&mut scope, context);
    let store = v8::ArrayBuffer::new_backing_store_from_boxed_slice(Box::new([])).make_shared();
    let _ = v8::ArrayBuffer::with_backing_store(&mut scope, &store);
}
```

crashes with
```
[1]    19724 segmentation fault (core dumped)  cargo r
```

Running it under valgrind one can see that it indeed tries to delete the placeholder pointer 0x1:
```
==19436== Invalid free() / delete / delete[] / realloc()
==19436==    at 0x483C9FC: free (vg_replace_malloc.c:538)
==19436==    by 0x1EEB5A: alloc::alloc::dealloc (alloc.rs:105)
==19436==    by 0x1EFA1A: <alloc::alloc::Global as core::alloc::Allocator>::deallocate (alloc.rs:242)
==19436==    by 0x1EEFFD: alloc::alloc::box_free (alloc.rs:336)
==19436==    by 0x1EDDCB: core::ptr::drop_in_place<alloc::boxed::Box<core::ffi::c_void>> (mod.rs:188)
==19436==    by 0x1F4BAD: core::mem::drop (mod.rs:909)
==19436==    by 0x1F31BB: rusty_v8::array_buffer::backing_store_deleter_callback (array_buffer.rs:239)
==19436==    by 0x3B4D44: v8::internal::BackingStore::~BackingStore() (backing-store.cc:209)
==19436==    by 0x1FB786: std::__1::__shared_ptr_pointer<v8::BackingStore*, std::__1::shared_ptr<v8::BackingStore>::__shared_ptr_default_delete<v8::BackingStore, v8::BackingStore>, std::__1::allocator<v8::BackingStore> >::__on_zero_shared() (memory:2503)
==19436==    by 0xA22A67: v8::internal::ArrayBufferSweeper::ReleaseAll() (array-buffer-sweeper.cc:199)
==19436==    by 0x2CEECC: v8::internal::Heap::TearDown() (heap.cc:5959)
==19436==    by 0x27AEB5: v8::internal::Isolate::Deinit() (isolate.cc:3188)
==19436==  Address 0x1 is not stack'd, malloc'd or (recently) free'd
```

andreubotella:
This results in a segmentation fault when dropping a `BackingStore` constructed through `ArrayBuffer::new_backing_store_from_boxed_slice` from an empty slice, since empty boxed slices can be invalid pointers, but `Box<c_void>` can't.

Fixes #849.

### DIFF
--- a/src/array_buffer.rs
+++ b/src/array_buffer.rs
@@ -3,6 +3,7 @@
 use std::cell::Cell;
 use std::ffi::c_void;
 use std::ops::Deref;
+use std::ptr;
 use std::ptr::null_mut;
 use std::ptr::NonNull;
 use std::slice;
@@ -233,11 +234,12 @@ pub type BackingStoreDeleterCallback = unsafe extern "C" fn(
 
 pub unsafe extern "C" fn backing_store_deleter_callback(
   data: *mut c_void,
-  _byte_length: usize,
+  byte_length: usize,
   _deleter_data: *mut c_void,
 ) {
-  let b = Box::from_raw(data);
-  drop(b)
+  let slice_ptr = ptr::slice_from_raw_parts_mut(data as *mut u8, byte_length);
+  let b = Box::from_raw(slice_ptr);
+  drop(b);
 }
 
 /// A wrapper around the backing store (i.e. the raw memory) of an array buffer.
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -5660,3 +5660,18 @@ fn function_names() {
     assert_eq!(v8_name.to_rust_string_lossy(scope), "");
   }
 }
+
+// https://github.com/denoland/rusty_v8/issues/849
+#[test]
+fn backing_store_from_empty_boxed_slice() {
+  let _setup_guard = setup();
+
+  let mut isolate = v8::Isolate::new(Default::default());
+  let mut scope = v8::HandleScope::new(&mut isolate);
+  let context = v8::Context::new(&mut scope);
+  let mut scope = v8::ContextScope::new(&mut scope, context);
+
+  let store = v8::ArrayBuffer::new_backing_store_from_boxed_slice(Box::new([]))
+    .make_shared();
+  let _ = v8::ArrayBuffer::with_backing_store(&mut scope, &store);
+}


piscisaureus:
The test that you've added shows that there is something going on here.

However,

1. I am a bit concerned that this could actually cause a leak in certain situations, e.g. when you first construct a boxed slice with size greater than 1 bytes and then shrink it to become empty. After all, rust relies on the libc memory allocator and allocating 0 bytes is actually possible and those allocation do actually need to be freed.
   I'd suggest to hunt down the source code in rust_std where the decision "to free or not to free" is made so we know what to look out for.
2. The commit message is confusing; `backing_store_deleter_callback` does not construct anything, not before nor after this patch.

andreubotella:
>     1. I am a bit concerned that this could actually cause a leak in certain situations, e.g. when you first construct a boxed slice with size greater than 1 bytes and then shrink it to become empty. Also, rust relies on the libc memory allocator and allocating 0 bytes is actually possible and those allocation do actually need to be freed.
>        I'd suggest to hunt down the source code in rust_std where the decision "to free or not to free" is made so we know what to look out for.

The `Box` implementation is magic, but the code that actually allocates and frees boxes is in [`alloc::alloc::exchange_malloc` and `alloc::alloc::alloc_free`](https://github.com/rust-lang/rust/blob/master/library/alloc/src/alloc.rs#L312-L338), which just defer to the global allocator, regardless of whether the size is 0. So presumably it should properly handle shrunk allocations.

I suspect the actual issue here is that `c_void` is not zero-sized.

>     2. The commit message is confusing; `backing_store_deleter_callback` does not construct anything, not before nor after this patch.

The docs for `Box::from_raw()` use the wording "constructs a box from a raw pointer", but you're right that in this context it isn't clear that that's what's meant.

piscisaureus:
> I suspect the actual issue here is that `c_void` is size 1, not 0, so when the box is dropped, the global allocator tries to free one byte from a zero-sized allocation.
> 
> >

Ran this little test:

```rust
fn main() {
    print_box_ptr(Box::<std::ffi::c_void>::new(unsafe {
        std::mem::MaybeUninit::zeroed().assume_init()
    }));
    print_box_ptr(Box::new([0u8; 0]));
    print_box_ptr(Vec::<u8>::with_capacity(0).into_boxed_slice());
    print_box_ptr(
        std::iter::repeat(0u8)
            .take(10)
            .collect::<Vec<_>>()
            .into_boxed_slice(),
    );
    print_box_ptr({
        let mut v = Vec::<u8>::with_capacity(10);
        v.shrink_to_fit();
        v.into_boxed_slice()
    });
    print_box_ptr(Box::new(()));
}

fn print_box_ptr<T: ?Sized>(b: Box<T>) {
    let ptr = &*b as *const T;
    eprintln!(
        "{}, ptr: {:?}, size: {}",
        std::any::type_name::<T>(),
        ptr,
        std::mem::size_of_val(&*b)
    );
}
```

```
core::ffi::c_void, ptr: 0x600002378040, size: 1
[u8; 0], ptr: 0x1, size: 0
[u8], ptr: 0x1, size: 0
[u8], ptr: 0x600002378040, size: 10
[u8], ptr: 0x1, size: 0
(), ptr: 0x1, size: 0
```

So it turns out that all zero sized allocations have "address" `0x01`, which happens to be the same as what `NonNull::dangling()` would return.

piscisaureus (approved):
LGTM, thanks for the fix and test, @andreubotella.

PR #855, opened 2021-12-14
Fix the build with python 3.10

espindola:
This updates jinja2 to ee69aa00ee8536f61db6a451f3858745cf587de6, which
is the version in v8's DEPS. This version is compatible with python
3.10.


ry (approved):
Thanks LGTM

PR #860, opened 2021-12-26
add support for `aarch64-linux-android`

littledivy:
Adds support for `aarch64-linux-android`.
Continuation of #270 with testing and build cache enabled.

Also adds a little frame buffer example (`example/android_fractal.rs`):
<image src="https://user-images.githubusercontent.com/34997667/147415649-bd9b1a45-bab9-4e15-b6fa-144b3c49ba6e.png" width=246 height=533 />

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -225,7 +225,7 @@ fn maybe_clone_repo(dest: &str, repo: &str) {
   if !Path::new(&dest).exists() {
     assert!(Command::new("git")
       .arg("clone")
-      .arg("--depth=1")
+      .arg("--depth=3")
       .arg(repo)
       .arg(dest)
       .status()
--- a/build.rs
+++ b/build.rs
@@ -117,14 +117,7 @@ fn build_v8() {
     gn_args.push("host_cpu=\"arm64\"".to_string())
   }
 
-  if cfg!(target_os = "android") && cfg!(target_arch = "aarch64") {
-    gn_args.push(format!(
-      "clang_base_path={:?}",
-      "third_party/llvm-build/Release+Asserts"
-    ));
-    gn_args.push("clang_use_chrome_plugins=false".to_string());
-    gn_args.push("treat_warnings_as_errors=false".to_string());
-  } else if let Some(clang_base_path) = find_compatible_system_clang() {
+  if let Some(clang_base_path) = find_compatible_system_clang() {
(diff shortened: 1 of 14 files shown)


build.rs
```diff
@@ -171,6 +212,29 @@ fn build_v8() {
...
+    .status()
+    .unwrap()
+    .success());
+}
+
+fn maybe_clone_repo(dest: &str, repo: &str) {
+  if !Path::new(&dest).exists() {
+    assert!(Command::new("git")
+      .arg("clone")
+      .arg("--depth=3")
```
bnoordhuis, line 229: Out of curiosity, why --depth=3?

examples/android_fractal.rs
```diff
@@ -0,0 +1,149 @@
...
+
+  let script = v8::Script::compile(try_catch, script, None)
+    .expect("failed to compile script");
+
+  if script.run(try_catch).is_none() {
+    let exception = try_catch.exception().unwrap();
+    let exception_string = exception
+      .to_string(try_catch)
+      .unwrap()
+      .to_rust_string_lossy(try_catch);
```
bnoordhuis, line 101: Minor suggestion for improvement (also ~30 lines below):
```suggestion
    let exception_string = try_catch
      .stack_trace()
      .or_else(|| try_catch.exception())
      .map(|value| value.to_rust_string_lossy(try_catch))
      .unwrap_or_else(|| "no stack trace".into());
```
That way it always prints something, including a stack trace when one exists.

bnoordhuis (approved):
LGTM, the example is a nice touch!

build.rs
littledivy: It's a workaround to later checkout to the ndk 22 commit https://github.com/denoland/rusty_v8/blob/894374a1bfe4de6fc1dde653c766b4fefec50de4/build.rs#L182-L187

littledivy: Andd i now realize how dumb this is. It's going to break when there is a new commit :(

lucacasonato:
This is very cool!

build.rs
littledivy: fixed

bnoordhuis (approved):
Still LGTM, just needs a minor lint fix, it looks like.

iMrDJAi:
Any plans to support other Android architectures? @littledivy is that possible?

PR #862, opened 2022-01-03
feat: Add bindings for `{Dis,}allowJavascriptExecutionScope`

### LINKED ISSUE #833: Add bindings for `AllowJavascriptExecutionScope`
In denoland/deno#12505 it came up that trying to run Javascript code from [`v8::ValueDeserializerImpl::read_host_object`](https://docs.rs/v8/0.34.0/v8/trait.ValueDeserializerImpl.html#method.read_host_object) resulted in a crash because apparently that method is invoked from V8 inside a `DisallowJavascriptExecutionScope`. Implementing structured clone for web APIs in Deno essentially requires running JS code in the implementation of that method, and so bindings to [`v8::Isolate::AllowJavascriptExecutionScope`](https://source.chromium.org/chromium/chromium/src/+/main:v8/include/v8-isolate.h;l=338) would be needed to make that possible.

andreubotella:
I've got the raw functionality working, but I'm not sure at all about the API and the lifetimes/generics. For the time being I've managed to patch something that works for the `ValueDeserializer` use case, but I'm not sure at all that it's any good.

Closes #833.﻿


andreubotella:
@piscisaureus PTAL

andreubotella:
This is not ready for merging – the generics probably need changes, and I didn't add tests to `scope.rs` because they'd probably have to be reworked after the changes – but it is ready for review.

bnoordhuis (approved):
The changes LGTM, I don't really have complaints. Perhaps @piscisaureus should look it over.

src/scope.rs
```diff
@@ -1618,6 +1888,75 @@ mod raw {
...
+  #[repr(C)]
+  #[derive(Debug)]
+  pub(super) struct DisallowJavascriptExecutionScope([usize; 3]);
+
+  impl DisallowJavascriptExecutionScope {
+    /// This function is marked unsafe because the caller must ensure that the
+    /// returned value isn't dropped before `init()` has been called.
+    pub unsafe fn uninit() -> Self {
+      // This is safe because there is no combination of bits that would produce
+      // an invalid `[usize; 3]`.
```
andreubotella, line 1900: Although this follows the example of other scopes, this isn't actually safe now that I look at it: uninitialized memory is different conceptually than any (initialized) combination of bits, and uninitialized memory isn't a valid value for `[usize; 3]`. It should be this instead:

```rust
pub(super) struct DisallowJavascriptExecutionScope(MaybeUninit<[usize; 3]>);
```

This doesn't block this PR, though.

piscisaureus:
Looking good so far. 
By the look of it, it seems possible to create a HandleScope inside a Disallow scope. Is that intentional?
What about creating a Disallow scope inside a Disallow scope? 
And what about creating an Allow scope inside another Allow scope?

aapoalas:
Are there any blockers for this?

andreubotella:
> Are there any blockers for this?

Rebasing and a thorough review from @piscisaureus I guess. But I haven't looked at this PR in over a year, and scopes are complicated.

andreubotella:
Rebased. Since this PR is the main blocker for denoland/deno#12067, which is lately gaining prominence due to Deno KV, it would be good to put some more work into this. However, I don't fully understand the details of how scopes are meant to work together, in terms of creating scopes from other scopes and as/deref, and my understanding of those details has certainly not grown in the year that this PR has been open. So a lot of help from @piscisaureus would probably be needed before this PR can be landed.

src/scope.rs
andreubotella: The Rust docs [now mention](https://doc.rust-lang.org/std/mem/union.MaybeUninit.html#initializing-an-array-element-by-element) that creating an array of `MaybeUninit<T>` with `MaybeUninit::uninit().assert_init()` is safe, so this is not actually an issue.

ry:
Closing because it's very old. Please open a new PR if you want to continue this.

andreubotella:
I last rebased this PR three months ago. The single merge conflict is trivial, and after resolving it, the tests pass. Is _that_ too old, really?

The issue this PR fixes is a main blocker for structured cloning and serializing web APIs, which I would have assumed is of high relevance to the Deno company lately because of Deno KV. But recently I haven't seen much interest from Deno in getting this PR landed, even though at this point the ball was on Deno's court, see https://github.com/denoland/rusty_v8/pull/862#issuecomment-1529785212.

At this point, I don't have the motivation to reopen this. If anyone does, feel free to take this code, and I would appreciate it if you would ping me in the resulting PR.

bartlomieju:
Hey @andreubotella, sorry for the miscommunication. I'm still very interested in landing this PR - sorry I hadn't picked it up earlier, it slipped in my backlog. 

The only thing Bert wanted before merging this one was an update to `deref_types` test from `src/scope.rs`. I will prioritize getting it done this week.

bartlomieju:
@andreubotella I added more test cases in `deref_types()` test. I'll wait a couple days for Bert to take a look and then land the PR.

PR #867, opened 2022-01-11
ci: update runner from ubuntu-latest-xl to ubuntu-20.04-xl

bartlomieju:


piscisaureus:
We might need to set up sysroot.

bartlomieju:
> We might need to set up sysroot.

Yikes, that's gonna be some more work. I can look into at the end of the week.

bartlomieju:
Thanks @bnoordhuis

piscisaureus:
We need to test whether Deno built with these rusty_v8 binaries still runs on Ubuntu 18.04.
I suspect the answer is no.

piscisaureus:
> We need to test whether Deno built with these rusty_v8 binaries still runs on Ubuntu 18.04.

I tested it. Seems to work fine.

piscisaureus (approved):

PR #871, opened 2022-01-19
fix: separate out android example into crate

littledivy:
should fix dependencies pulled in https://github.com/denoland/deno/pull/13415


Cargo.toml
```diff
@@ -101,7 +95,3 @@ name = "shell"
 
 [[example]]
 name = "process"
```
lucacasonato, line 97:
```suggestion
name = "process"

[workspace]
members = [
  "examples/android"
]
```

examples/android/.gitignore
```diff
@@ -0,0 +1,2 @@
+target/
```
lucacasonato, line 1: Not needed with the above change.

Cargo.toml
littledivy: That includes it's dependency in Cargo.lock :/

lucacasonato: That's fine. It won't be in importers' Cargo.lock. It being in rusty_v8's lock file is desirable.

littledivy: Alrighty, done.

lucacasonato (approved):
LGTM

PR #872, opened 2022-01-19
Add v8::Map bindings

RomainMuller:
Support v8::Map operations (`new`, `clear`, `get`, `set`, `has`,
and `delete`).


bnoordhuis (approved):
LGTM, thanks.

PR #874, opened 2022-01-20
Add support for v8::Object.set_integrity_level

RomainMuller:
This allows making objects frozen or sealed from rust code. This is
useful to control mutations that are allowed to happen in JS on the
given object.

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -114,6 +114,11 @@ extern "C" {
     this: *const Object,
     index: int,
   ) -> *const Value;
+  fn v8__Object__SetIntegrityLevel(
+    this: *const Object,
+    context: *const Context,
+    level: IntegrityLevel,
+  ) -> MaybeBool;
   fn v8__Object__SetInternalField(
     this: *const Object,
     index: int,
@@ -487,6 +492,18 @@ impl Object {
     None
   }
 
+  /// Sets the integrity level of the object.
+  pub fn set_integrity_level(
+    &self,
+    scope: &mut HandleScope,
+    level: IntegrityLevel,
+  ) -> Option<bool> {
+    unsafe {
+      v8__Object__SetIntegrityLevel(self, &*scope.get_current_context(), level)
+    }
+    .into()
+  }
+
   /// Sets the value in an internal field. Returns false when the index
   /// is out of bounds, true otherwise.
   pub fn set_internal_field(&self, index: usize, value: Local<Value>) -> bool {
@@ -569,6 +586,19 @@ impl Object {
     }
     .into()
   }
+}
+
+/// Object integrity levels can be used to restrict what can be done to an
+/// object's properties.
+#[repr(C)]
+pub enum IntegrityLevel {
+  /// Frozen objects are like Sealed objects, except all existing properties are
+  /// also made non-writable.
+  Frozen,
+  /// Sealed objects prevent addition of any new property on the object, makes
+  /// all existing properties non-configurable, meaning they cannot be deleted,
+  /// have their enumerability, configurability, or writability changed.
+  Sealed,
 }
 
 impl Array {
--- a/src/object.rs
+++ b/src/object.rs
@@ -590,6 +590,7 @@ impl Object {
 
 /// Object integrity levels can be used to restrict what can be done to an
 /// object's properties.
+#[derive(Debug)]
 #[repr(C)]
 pub enum IntegrityLevel {
   /// Frozen objects are like Sealed objects, except all existing properties are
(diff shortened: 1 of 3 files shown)


src/object.rs
```diff
@@ -571,6 +588,19 @@ impl Object {
   }
 }
 
+/// Object integrity levels can be used to restrict what can be done to an
+/// object's properties.
+#[repr(C)]
+pub enum IntegrityLevel {
```
RomainMuller, line 594: :question: Should this be:

```suggestion
#[derive(Debug)]
#[repr(C)]
pub enum IntegrityLevel {
```

tests/test_api.rs
```diff
@@ -1553,6 +1563,29 @@ fn object() {
...
+    let object_string = v8::String::new(scope, "object").unwrap().into();
+    global.set(scope, object_string, object.into());
+    assert!(object
+      .set_integrity_level(scope, v8::IntegrityLevel::Sealed)
+      .unwrap());
+    {
+      let scope = &mut v8::TryCatch::new(scope);
+      // Creating a new property fails (in strict mode).
+      eval_strict(scope, "object.c = 'foo';");
+      assert!(scope.has_caught());
```
RomainMuller, line 1577: In non-strict mode, this silently does nothing.

tests/test_api.rs
```diff
@@ -1553,6 +1563,29 @@ fn object() {
...
+    }
+
+    assert!(object
+      .set_integrity_level(scope, v8::IntegrityLevel::Frozen)
+      .unwrap());
+    {
+      let scope = &mut v8::TryCatch::new(scope);
+      // Writing an existing property fails (in strict mode).
+      eval_strict(scope, "object.a = 'foo';");
+      assert!(scope.has_caught());
```
RomainMuller, line 1587: In non-strict mode, this silently does nothing.

src/object.rs
bnoordhuis: Yes :-)

tests/test_api.rs
bnoordhuis: It doesn't throw if that is what you mean but it's still observable because `object.c` won't have been set.

The way you can test this in sloppy mode is by checking that `Object.isSealed()` and `Object.isExtensible()` return false and true before and the inverse afterward - and that the property hasn't been set, of course. Ditto for frozen mode.

Aside: I'd prefer not to have this `eval_strict()` helper. It obscures more than it clarifies, IMO.

RomainMuller: >The way you can test this in sloppy mode is by checking that Object.isSealed() and Object.isExtensible() return false and true before and the inverse afterward - and that the property hasn't been set, of course. Ditto for frozen mode.

I did not think of `Object.isSealed`, `isExtensible` and `isFrozen`... I'm actually tempted to "trust" their return value, and so to skip trying to create/mutate a property and observing the expected side-effect (or absence thereof). I'm leaving those in for now, though...

> Aside: I'd prefer not to have this `eval_strict()` helper. It obscures more than it clarifies, IMO.

Yeah... Since this can be tested fully in sloppy mode, it does not make a ton of sense. I initially added it largely because I found `'use strict';\nobject.c = 'foo';` looked kinda bad...

tests/test_api.rs
```diff
@@ -1553,6 +1554,49 @@ fn object() {
...
+    let global = context.global(scope);
+    let object_string = v8::String::new(scope, "o").unwrap().into();
+    global.set(scope, object_string, object.into());
+
+    let true_value = v8::Boolean::new(scope, true);
+    let false_value = v8::Boolean::new(scope, false);
+
+    assert_eq!(eval(scope, "Object.isExtensible(o)").unwrap(), true_value);
+    assert_eq!(eval(scope, "Object.isSealed(o)").unwrap(), false_value);
+    assert_eq!(eval(scope, "Object.isFrozen(o)").unwrap(), false_value);
```
bnoordhuis, line 1567: It's slightly more idiomatic (and shorter) to use is_true() and is_false():
```suggestion
    assert!(eval(scope, "Object.isExtensible(o)").unwrap().is_true());
    assert!(eval(scope, "Object.isSealed(o)").unwrap().is_false());
    assert!(eval(scope, "Object.isFrozen(o)").unwrap().is_false());
```
I wanted to suggest this as a diff for the whole test but somehow GH won't let me. At any rate, that's the only nit I have and it's not really a blocker, more a suggestion.

bnoordhuis (approved):

tests/test_api.rs
bnoordhuis: You can use multi-line strings for that, they also let you embed quotes:
```rust
let source = r#"
  "use strict";
  object.c = "foo";
"#;
```

PR #879, opened 2022-01-22
Add v8::Value::InstanceOf bindings

littledivy:


src/binding.cc
```diff
@@ -673,6 +673,11 @@ const v8::Boolean* v8__Value__ToBoolean(const v8::Value& self,
   return local_to_ptr(self.ToBoolean(isolate));
 }
 
+bool v8__Value__InstanceOf(const v8::Value& self, const v8::Context& context,
```
bnoordhuis, line 676:
```suggestion
bool v8__Value__InstanceOf(v8::Value& self, const v8::Context& context,
```
And `*mut Value` in rust land.

piscisaureus:
I am not in favor of implementing deref_mut for Local handles.

src/handle.rs
```diff
@@ -111,6 +112,12 @@ impl<'s, T> Deref for Local<'s, T> {
   }
 }
 
+impl<'s, T> DerefMut for Local<'s, T> {
+  fn deref_mut(&mut self) -> &mut T {
+    unsafe { self.0.as_mut() }
+  }
+}
```
bartlomieju, line 119: As per @piscisaureus's comment we should remove this

littledivy: `instance_of` requires for `T` to be mutable in `v8::Local<T>`, hence the deref_mut implementation. I'm not sure how you can do `instance_of` on a local handle without this.

inteon: Is there a reason the `InstanceOf` function is not marked as `const` in the c++ api?
Seems counter-intuitive, could this be a mistake in the v8 c++ api?

littledivy: @inteon See https://github.com/denoland/rusty_v8/pull/879#discussion_r790182815

aapoalas:
I may well be wrong, but it seems like this binding will be necessary for implementing magic (de)serialiser for FFI-related custom class wrapping ArrayBuffers for passing to native libraries.

src/handle.rs
```diff
@@ -4,6 +4,7 @@ use std::hash::Hasher;
 use std::marker::PhantomData;
 use std::mem::transmute;
 use std::ops::Deref;
+use std::ops::DerefMut;
```
piscisaureus, line 7: Unused import

PR #885, opened 2022-01-26
Make v8::script_compiler::Source.get_cached_data return Option

### LINKED ISSUE #884: v8::script_compiler::Source.get_cached_data can lead to crash
The `cached_data` property of `v8::ScriptCompiler::Source` is optional/nullable. A proof of this is how `v8::script_compiler::Source` has two constructors (one with cached data, and one without).

Constructing a `Source` value without cached data will lead to `get_cached_data` returning an invalid `CachedData` value (i.e: `null`).

https://github.com/denoland/rusty_v8/blob/729c5b323f5bd172056a240c4bd9d9639b79be66/src/script_compiler.rs#L146-L148

That accessor should return `Option<&CachedData>`, I think?

RomainMuller:
The `cached_data` property of `Source` is optional, so reading the value
should return `Option<&CachedData>`.

Fixes #884

### DIFF
--- a/src/script_compiler.rs
+++ b/src/script_compiler.rs
@@ -143,8 +143,15 @@ impl Source {
     }
   }
 
-  pub fn get_cached_data(&self) -> &CachedData {
-    unsafe { &*v8__ScriptCompiler__Source__GetCachedData(self) }
+  pub fn get_cached_data(&self) -> Option<&CachedData> {
+    unsafe {
+      let cached_data = v8__ScriptCompiler__Source__GetCachedData(self);
+      if cached_data.is_null() {
+        None
+      } else {
+        Some(&*cached_data)
+      }
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
Thanks @RomainMuller, could you add a test case for `get_cached_data()` that returns `None`?

RomainMuller:
> Thanks @RomainMuller, could you add a test case for `get_cached_data()` that returns `None`?

There is already an indirect one, because of the test I've edited. But I'll agree it's not obvious that a code path leads to there with no cached data. Let me add something there real quick.

bartlomieju (approved):
LGTM

RomainMuller:
I'm not sure why the `license/cla` check is not returning here. I've signed the SLA and multiple of my contributions have already been merged... I guess it must have crashed?

lucacasonato:
@RomainMuller The CLA bot seems to have had a little hiccup. Closing and re-opening the PR triggered it to re-check.

PR #888, opened 2022-02-01
Add bindings for v8::ScriptCompiler::CreateCodeCacheForFunction

RomainMuller:
And also for CachedDataVersionTag while I was at it.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -435,6 +435,10 @@ const v8::UnboundScript* v8__ScriptCompiler__CompileUnboundScript(
   return maybe_local_to_ptr(maybe_local);
 }
 
+uint32_t v8__ScriptCompiler__CachedDataVersionTag() {
+  return v8::ScriptCompiler::CachedDataVersionTag();
+}
+
 bool v8__Data__EQ(const v8::Data& self, const v8::Data& other) {
   return ptr_to_local(&self) == ptr_to_local(&other);
 }
@@ -1861,6 +1865,10 @@ v8::ScriptCompiler::CachedData* v8__UnboundModuleScript__CreateCodeCache(
     const v8::UnboundModuleScript& unbound_module_script) {
   return v8::ScriptCompiler::CreateCodeCache(
       ptr_to_local(&unbound_module_script));
+}
+
+v8::ScriptCompiler::CachedData* v8__Function__CreateCodeCache(const v8::Function& self) {
+  return v8::ScriptCompiler::CreateCodeCacheForFunction(ptr_to_local(&self));
 }
 
 const v8::Value* v8__Script__Run(const v8::Script& script,
--- a/src/function.rs
+++ b/src/function.rs
@@ -484,7 +484,16 @@ impl Function {
   /// Creates and returns code cache for the specified unbound_script.
   /// This will return nullptr if the script cannot be serialized. The
   /// CachedData returned by this function should be owned by the caller.
-  pub fn create_code_cache<'s>(&self) -> Option<UniqueRef<CachedData<'s>>> {
-    unsafe { UniqueRef::try_from_raw(v8__Function__CreateCodeCache(self)) }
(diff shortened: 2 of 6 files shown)


bartlomieju (commented):
Looks good to me, but I will let others take a look before landing

piscisaureus:
Mostly looks good.

 I have some concern about the lifetime on `CacheData<‘t>` though - either the lifetime is bogus and unnecessary, or the buffer referenced by `CacheData` returned from `create_code_cache()` is never released (and thus leaked).

RomainMuller:
I have essentially stolen the lifetime convention from that of `UnboundScript`:

https://github.com/denoland/rusty_v8/blob/8b90dfd2f4fcbbaefd2c4d2be220d94a00a3ebba/src/unbound_script.rs#L28-L31

As well as `UnboundModuleScript`:

https://github.com/denoland/rusty_v8/blob/8b90dfd2f4fcbbaefd2c4d2be220d94a00a3ebba/src/unbound_module_script.rs#L12-L15

The V8 APIs look identical, so I assumed the same lifetime would make sense.

That being said, I'd tend to agree with you that this looks weird here... I'd expect the lifetime of the `CachedData` backing buffer to outlive (or at least, be unrelated to) the `Function`/`UnboundModule`/`UnboundModuleScript`.

The V8 API documentation states that the caller is expected to assume ownership of the `CachedData` object, which I reckon is why this ends up being wrapped in a `UniqueRef`.

I _suspect_ it might be more sensible to change all of those `create_code_cache` method signatures to be like:

```rs
pub fn create_code_cache<'s>(&self) -> Option<UniqueRef<CachedData<'s>>> {
  // ...
}
```

Which I reckon would be identical to the slightly more compact form:

```rs
pub fn create_code_cache(&self) -> Option<UniqueRef<CachedData<'_>>> {
  // ...
}
```

If you agree with me I'd extend this PR to align the signatures of the existing methods so it's all consistent...

RomainMuller:
>If you agree with me I'd extend this PR to align the signatures of the existing methods so it's all consistent...

I figured I'd do it... it's easy to roll back if it turns out I'm not understanding this properly :)

piscisaureus:
Upon further inspection I think you got it right the first time. [`CachedData`](https://github.com/denoland/rusty_v8/blob/a09d392711e6e4a416e9d051c8555c190efe182d/src/script_compiler.rs#L70-L76) has a [`buffer_policy`](https://github.com/denoland/rusty_v8/blob/a09d392711e6e4a416e9d051c8555c190efe182d/src/script_compiler.rs#L107-L110) field that determines whether the `CachedData` destructor releases the underlying buffer.

When calling `CachedData::new()` a new instance is created that doesn't own its buffer, so its lifetime must be constrained to the lifetime of the buffer. 

`create_code_cache()` supposedly returns an instance that owns its buffer, so returning an instance with a `'static` lifetime is appropriate here. I'd suggest to add an assert for this wherever `CachedData<'static>` is returned.

```rust
assert_eq!(cached_data.buffer_policy, BufferPolicy::BufferOwned);
```

You might need to `#[derive(Eq, PartialEq)]` for `BufferPolicy`.

WRT https://github.com/denoland/rusty_v8/pull/888/commits/fea4a1a8bdc6b792fe47a8d51b5fc711c8eb2e7f, this effectively does the same as returning `'static` as the lifetime `'s` is not constrained in any way. So I'd just use `'static` explicitly, it's more clear IMO.

RomainMuller:
Okay, I've added debug assertions around the `buffer_policy` (this involved making the `BufferPolicy` enum crate-public, and adding a `fn buffer_policy(&self)` crate-public accessor.

I've wrapped this under `#[cfg(debug_assertions)]` in places where I just couldn't easily use a `debug_assert_eq!` (i.e: when I get an `Option`), in the spirit of "Do not introduce additional call overhead." within release builds.

I also added an assertion in the `CachedData` constructor that it's buffer policy is `BufferNotOwned`, for good measure/symmetry.

piscisaureus (approved):
LGTM

Issue #469, opened 2020-09-14
Where is v8::Isolate::GetCurrent?


Icemic:
Hi, I am porting a program on v8 in C++ to Rust, and finding rusty_v8.

In C++, I always write `auto isolate = v8::Isolate::GetCurrent();` to get current isolate somewhere (for example, in the callback of libuv). But I cannot find this method in rusty_v8 as rusty_v8::Isolate only has `new()` and `create_params()`.

https://chromium.googlesource.com/v8/v8.git/+/refs/heads/8.7.75/include/v8.h#8592

I tried adding binding by myself: 

```rust
extern "C" {
    fn v8__Isolate__GetCurrent() -> *mut Isolate;
}
```

Only to get a linking error telling me there's no Isolate::GetCurrent in rusty_v8.lib

So my question are:
1. Does rusty_v8 use a modified version of v8 which removes GetCurrent method?
2. How to get current isolate in rusty_v8?

Thanks.

bnoordhuis:
> 1. Does rusty_v8 use a modified version of v8 which removes GetCurrent method?

No. :-)

> 2. How to get current isolate in rusty_v8?

You also need to update `src/binding.cc`.

That said, I don't think `Isolate::get_current()` would play nicely with Rust's ownership and aliasing rules.

Does `isolate.thread_safe_handle()` work for your use case? It returns an `IsolateHandle` that's `Clone`, `Send` and `Sync`, in order that you can (safely!) pass it around.

Icemic:
> > 1. Does rusty_v8 use a modified version of v8 which removes GetCurrent method?
> 
> No. :-)
> 
> > 1. How to get current isolate in rusty_v8?
> 
> You also need to update `src/binding.cc`.
> 
> That said, I don't think `Isolate::get_current()` would play nicely with Rust's ownership and aliasing rules.
> 
> Does `isolate.thread_safe_handle()` work for your use case? It returns an `IsolateHandle` that's `Clone`, `Send` and `Sync`, in order that you can (safely!) pass it around.

I don't think so too : P  but seems no choice before.

And I did notice IsolateHandle, but how to get a valid isolate via IsolateHandle? In document, it says a it's to terminal a execution instead of malloc a new handle scope. IsolateHandle only has one `terminate_execution` method.

btw, my case:

```cpp
  // it's a libuv callback with only the timer handle of libuv as parameter
  void _timer_callback(uv_timer_t* timer) {
    // get a isolate
[...]

PR #892, opened 2022-02-06
feat: add v8__Isolate__GetCurrent

zhangyuang:
Ref [docs](https://v8docs.nodesource.com/node-4.8/d5/dda/classv8_1_1_isolate.html#a66324a6a0a257b62865065bf539fa442)
Support `v8::Isolate::get_current()`, we can get current isolate snapshot in a existed process


littledivy:
@zhangyuang `rusty_v8` Isolates are `!Send` (cannot be sent between threads). See https://github.com/denoland/rusty_v8/issues/469. 
What is your usecase? Also, If this is added it should be marked `unsafe`.

zhangyuang:
@littledivy I want to use it to get current snapshot in a nodejs program like [v8-profiler](https://github.com/hyj1991/v8-profiler-next/blob/316454184e536cfbc785818929fd033ea5cb9c8b/src/heap_profiler.cc#L77)

littledivy:
If your program is single threaded use `thread_local` (See https://github.com/denoland/rusty_v8/issues/469#issuecomment-693215285).

zhangyuang:
Thanks, I will try it later

zhangyuang:
Excuse me, can you give me more entire demo?

`ISOLATE.with(|slot| slot.replace(Some(unsafe { transmute(*isolate) })));`

I don't know where is the isolate from?
image

zhangyuang:
emmm, i think my purpose is not use isolate::new to create a new isolate but use current v8 isolate which is nodejs process executed

hengfeiyang (approved):

littledivy:
See https://github.com/denoland/rusty_v8/issues/469 for why rusty_v8 doesn't bind `Isolate::GetCurrent`. Closing.

PR #895, opened 2022-02-09
Weak handles and finalizers

### LINKED ISSUE #861: Support weak references or a Rust-level WeakMap
denoland/deno#13239 will need to store data specific to each context, and for some operations (`sync_ops_cache`, for example) it would have to iterate through the existing known contexts. But contexts should still be able to be garbage collected, so it seems like this would require some implementation of an iterable `WeakMap` whose keys are V8-managed types and whose values are regular Rust types. If rusty_v8 had support for weak references with finalizer callbacks, this weak map would be implementable in Deno, but it seems like that would require the weak reference to be pinned to the stack or heap, so maybe it's better for rusty_v8 to just come with such a weak map instead.

andreubotella:
Just a strawperson proposal to test out the waters. The Discord conversation mentioned storing the callbacks in the isolate annex (or something), but doing it with pinned boxes seems easier to implement. We can still change it to use the annex, of course.

Closes #861.

### DIFF
--- a/src/handle.rs
+++ b/src/handle.rs
@@ -486,10 +486,12 @@ impl<T> Weak<T> {
 
   /// Create a weak handle with a finalization callback installed.
   ///
-  /// Unlike in C++, this Rust API does guarantee that any finalizers whose
-  /// corresponding `Weak` is still alive will have been called when the isolate
-  /// is dropped. Other than that, there is still no guarantee as to when it
-  /// will be invoked.
+  /// There is no guarantee as to *when* the finalization callback will be
+  /// invoked. However, unlike the C++ API, this API guarantees that when an
+  /// isolate is destroyed, any finalizers that haven't been called yet will be
+  /// run, unless a [`Global`] reference is keeping the object alive. Other than
+  /// that, there is still no guarantee as to when the finalizers will be
+  /// called.
   ///
   /// The callback does not have access to the inner value, because it has
   /// already been collected by the time it runs.
@@ -865,9 +867,7 @@ impl FinalizerMap {
     id
   }
 
-  pub(crate) fn take_map(
-    &mut self,
-  ) -> std::collections::HashMap<FinalizerId, Box<dyn FnOnce(&mut Isolate)>> {
-    std::mem::take(&mut self.map)
+  pub(crate) fn is_empty(&self) -> bool {
+    self.map.is_empty()
   }
 }
--- a/src/handle.rs
+++ b/src/handle.rs
@@ -1,4 +1,5 @@
 use std::borrow::Borrow;
+use std::cell::Cell;
 use std::hash::Hash;
 use std::hash::Hasher;
 use std::marker::PhantomData;
@@ -7,6 +8,9 @@ use std::mem::transmute;
 use std::ops::Deref;
 use std::ptr::NonNull;
 
+use libc::c_void;
+
+use crate::support::Opaque;
 use crate::Data;
 use crate::HandleScope;
 use crate::Isolate;
@@ -15,7 +19,23 @@ use crate::IsolateHandle;
 extern "C" {
   fn v8__Local__New(isolate: *mut Isolate, other: *const Data) -> *const Data;
   fn v8__Global__New(isolate: *mut Isolate, data: *const Data) -> *const Data;
+  fn v8__Global__NewWeak(
+    isolate: *mut Isolate,
+    data: *const Data,
+    parameter: *const c_void,
+    callback: extern "C" fn(*const WeakCallbackInfo),
+  ) -> *const Data;
   fn v8__Global__Reset(data: *const Data);
+  fn v8__WeakCallbackInfo__GetIsolate(
+    this: *const WeakCallbackInfo,
+  ) -> *mut Isolate;
+  fn v8__WeakCallbackInfo__GetParameter(
+    this: *const WeakCallbackInfo,
+  ) -> *mut c_void;
+  fn v8__WeakCallbackInfo__SetSecondPassCallback(
+    this: *const WeakCallbackInfo,
+    callback: extern "C" fn(*const WeakCallbackInfo),
+  );
 }
 
 /// An object reference managed by the v8 garbage collector.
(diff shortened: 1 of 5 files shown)


andreubotella:
cc @lucacasonato @piscisaureus

src/handle.rs
```diff
@@ -419,3 +437,287 @@ impl HandleHost {
...
+/// the key for a [`HashMap`] or [`HashSet`].
+///
+/// # `Clone`
+///
+/// Since finalization callbacks are specific to a `Weak<T>` instance, cloning
+/// will create a new object reference without a finalizer, as if created by
+/// [`Self::new`]. You can use [`Self::clone_with_finalizer`] to attach a
+/// finalization callback to the clone.
+pub struct Weak<T> {
+  data: Option<Pin<Box<WeakData<T>>>>,
```
piscisaureus, line 464: Not super convinced this needs to be `Pin`.

Boxed values are implicitly pinned, except for the magical move-out-of-box operation (`let foo = *boxed_foo`). But since the `Box` is embedded in a struct here you wouldn't be able to do that anyway.

piscisaureus:
I know this is WIP, but it would be nice to get some sort of test so we can see that it works (and that lets us poke at some edge cases).

src/handle.rs
```diff
@@ -419,3 +437,287 @@ impl HandleHost {
...
+  }
+
+  pub fn empty(isolate: &mut Isolate) -> Self {
+    Weak {
+      data: None,
+      isolate_handle: isolate.thread_safe_handle(),
+    }
+  }
+
+  pub fn clone_with_finalizer(&self, finalizer: Box<dyn FnOnce()>) -> Self {
```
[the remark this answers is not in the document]
andreubotella: `Weak<T>` doesn't implement `Handle`, and changing `Handle` to support weak references is probably more trouble than it's worth.

src/handle.rs
andreubotella: That's not quite true, since `std::mem::swap` is a safe function:

```rust
// Self-referential struct
struct SelfRef {
  bytes: [u8; 1024],
  pointer: *const u8
}
impl SelfRef {
  pub fn new() -> Box<SelfRef> { todo!() }
  pub fn get_pointed_byte(&self) -> &u8 {
    // When the struct is created, `pointer` points into the `bytes` block, so this is fine, right?
    unsafe { &*self.pointer }
  }
}

let mut a = SelfRef::new();
let mut b = SelfRef::new();
std::mem::swap(&mut a, &mut b);  // !!!
drop(a);
b.get_pointed_byte();  // This is now a dangling reference
```

With `Pin<Box<T>>` you have access to `Pin<&mut T>`, but not to `&mut T` (except with unsafe methods or if `T: Unpin`), so you can't do `std::mem::swap`. You _can_ do [`Pin::set`](https://doc.rust-lang.org/std/pin/struct.Pin.html#method.set) because the Pin contract is that the value doesn't move for as long as it lives, and `Pin::set` drops the previous value.

In this case, we're not using `std::mem::swap` or anything like that, but since we're relying on the guarantee that the boxed data doesn't move, I guess it doesn't hurt to express that via the type system.

src/handle.rs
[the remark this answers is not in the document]
piscisaureus: > Weak<T> doesn't implement Handle

Oh, right.

> because changing Handle to support weak references is probably more trouble than it's worth.

I agree with that!

src/handle.rs
andreubotella: Come to think about it, we don't really care that the value isn't moved, at least not for safety. We care that the box stays at that particular memory location and isn't deallocated until after the second pass callback, which is guaranteed by `Box` and by the fact that the box is never moved or deallocated until the weak handle is dropped, upon which `reset` is called, which means the callbacks won't be called.

src/handle.rs
```diff
@@ -419,3 +436,282 @@ impl HandleHost {
...
+        v8__WeakCallbackInfo__SetSecondPassCallback(
+          wci,
+          Self::second_pass_callback,
+        )
+      };
+    }
+  }
+
+  extern "C" fn second_pass_callback(wci: *const WeakCallbackInfo) {
+    // FIXME!!!!: Do we know for a fact that the parameter hasn't been dropped??
```
littledivy, line 625: I think it is safe? `WeakData<T>` is not dropped until `Weak<T>` itself drops.

andreubotella: The first pass callback must be called the moment the reference is GC'd, since it's that callback's job to reset the weak handle, but the second pass callback has no such guarantees as far as I'm aware. [Looking in the V8 code](https://source.chromium.org/chromium/chromium/src/+/main:v8/src/handles/global-handles.cc;l=1567) it seems like sometimes second passes run during GC, but other times they're scheduled on the platform task runner (which I believe would only get called when pumping the V8 message loop at the beginning of the next run in Deno's event loop).

andreubotella:
I was getting weird results when creating weak handles from globals, and apparently dropping globals will not cause the first callback to be called?

```rust
#[test]
fn weak_from_global() {
  let _setup_guard = setup();

  let mut scope = v8::HandleScope::new(&mut isolate);
  let context = v8::Context::new(&mut scope);
  let scope = &mut v8::ContextScope::new(&mut scope, context);

  let global = {
    let object = v8::Object::new(scope);
    v8::Global::new(scope, object)
  };

  let weak = v8::Weak::new(scope, &global);
  assert!(!weak.is_empty());
  assert_eq!(weak.open(scope).unwrap(), global.open(scope));

  drop(global);
  eval(scope, "gc()").unwrap();
  assert!(weak.is_empty());  // This assertion fails.
}
```

littledivy:
@andreubotella Try a `v8::Isolate::low_memory_notification` after `drop(global)` https://docs.rs/v8/latest/v8/struct.Isolate.html#method.low_memory_notification 
d8 uses it to force weak callback invocation. https://chromium.googlesource.com/v8/v8/+/3.27.34.7/src/d8.cc#1385

piscisaureus:
> I was getting weird results when creating weak handles from globals, and apparently dropping globals will not cause the first callback to be called.

I think the problem is that the HandleScope is still keeping the `Local<Object>` that was initially created alive.

piscisaureus:
@andreubotella FYI I've fixed and added the `weak_from_global` test.

src/handle.rs
```diff
@@ -432,3 +449,289 @@ impl HandleHost {
...
+    } else {
+      None
+    }
+  }
+
+  pub fn is_empty(&self) -> bool {
+    self.get_pointer().is_none()
+  }
+
+  pub fn open<'a>(&'a self, isolate: &mut Isolate) -> Option<&'a T> {
```
piscisaureus, line 578: I'm not 100% sure this is safe, because the returned pointer is not protected from getting garbage collected. Consider this:

```rust
let f = v8::Weak<v8::Function> = get_fn();
// Called function might trigger garbage collection, which could lead to `f` getting GCed.
f.open().unwrap().call(...); 
```

So maybe it's best to not have this function at all. Users can use `as_local()` instead.

src/handle.rs
```diff
@@ -432,3 +449,289 @@ impl HandleHost {
...
+    if let Some(data) = self.get_pointer() {
+      let handle_host: HandleHost = (&self.isolate_handle).into();
+      handle_host.assert_match_isolate(isolate);
+      Some(unsafe { &*data.as_ptr() })
+    } else {
+      None
+    }
+  }
+
+  pub fn as_global(&self, isolate: &mut Isolate) -> Option<Global<T>> {
```
piscisaureus, line 588: nit: `to_global()` would be more idiomatic.

src/handle.rs
```diff
@@ -432,3 +449,289 @@ impl HandleHost {
...
+    if let Some(data) = self.get_pointer() {
+      let handle_host: HandleHost = (&self.isolate_handle).into();
+      handle_host.assert_match_isolate(isolate);
+      Some(unsafe { Global::from_raw(isolate, data) })
+    } else {
+      None
+    }
+  }
+
+  pub fn as_local<'s>(
```
piscisaureus, line 598: nit: `to_local()` would be more idiomatic.

src/handle.rs
andreubotella: Yeah, I'd wondered about that too. Agreed.

src/handle.rs
```diff
@@ -432,3 +449,289 @@ impl HandleHost {
...
+    // parameter points to hasn't been dropped.
+    let weak_data = unsafe {
+      let ptr = v8__WeakCallbackInfo__GetParameter(wci);
+      &*(ptr as *mut WeakData<T>)
+    };
+
+    if let Some(data) = weak_data.pointer.take() {
+      unsafe { v8__Global__Reset(data.cast().as_ptr()) };
+    } else {
+      unreachable!();
```
piscisaureus, line 632: Use `.unwrap()` instead of `if let`?

src/handle.rs
piscisaureus: What mechanism exists that could cause the `WeakData` to be dropped earlier than expected?

piscisaureus:
Added some comments but this looks pretty good to me overall.
@andreubotella is there anything you are still trying to figure out or is it ready for final review?

src/handle.rs
andreubotella: My concern is a scenario where the value is GC'd while `Weak` is still alive, but the second pass callback is scheduled for later, and `Weak` drops before that second pass callback runs. We could have flags in `WeakData` to check whether the second callback has run, and whether the corresponding `Weak` has dropped, and then leak the `Box<WeakData>` in the drop impl and deallocate it in the second pass callback if this is the case.

It seems like the various ways to trigger GC in tests (`gc()`, sending a low-memory notification, even running the GC in predictable mode) will cause second passes to run synchronously, so I'm not sure that this is testable.

andreubotella:
> Added some comments but this looks pretty good to me overall. @andreubotella is there anything you are still trying to figure out or is it ready for final review?

I was trying to figure out how to test that there's no UB if the second pass callback is called asynchronously, after `Weak` is dropped, but since it doesn't seem to be testable, I've just added a fix that seems like would work, so this is now ready for review.

piscisaureus:
> I was trying to figure out how to test that there's no UB if the second pass callback is called asynchronously, after `Weak` is dropped, but since it doesn't seem to be testable, I've just added a fix that seems like would work, so this is now ready for review.

I think the biggest concern is now what happens when the isolate is disposed between the first and second stage.
I think in the latest iteration of this patch the `WeakData` would be leaked, unless V8 runs all pending second stage callbacks synchronously when disposed.

Is that correct?

andreubotella:
> I think the biggest concern is now what happens when the isolate is disposed between the first and second stage. I think in the latest iteration of this patch the `WeakData` would be leaked, unless V8 runs all pending second stage callbacks synchronously when disposed.
> 
> Is that correct?

That's right. But I don't think there's a way to prevent leaking the `WeakData`.

src/handle.rs
```diff
@@ -432,3 +449,301 @@ impl HandleHost {
...
+/// will create a new object reference without a finalizer, as if created by
+/// [`Self::new`]. You can use [`Self::clone_with_finalizer`] to attach a
+/// finalization callback to the clone.
+#[derive(Debug)]
+pub struct Weak<T> {
+  data: Option<Box<WeakData<T>>>,
+  isolate_handle: IsolateHandle,
+}
+
+impl<T> Weak<T> {
```
bartlomieju, line 472: Talked with @littledivy and it seems we're gonna need similar API to `v8::Global` to be able to turn `Weak` into raw pointer and then back into `v8::Weak` - maybe we should add public APIs `into_raw()` and `from_raw()`?

andreubotella: Assuming it doesn't matter for the NAPI that the pointer points to a `WeakData<T>` rather than pointing into V8's heap, I guess that's fine.

bartlomieju: > Assuming it doesn't matter for the NAPI that the pointer points to a `WeakData<T>` in the Rust heap rather than pointing into V8's heap, I guess that's fine.

Yeah, I think the only limitation is that it's expected that it's a size of a single pointer, so we'd need to manually pass back `Isolate` - but the pattern from `v8::Global` works fine now, so we should be good.

src/handle.rs
```diff
@@ -432,3 +449,301 @@ impl HandleHost {
...
+      )
+    } else {
+      Weak {
+        data: None,
+        isolate_handle: self.isolate_handle.clone(),
+      }
+    }
+  }
+
+  fn get_pointer(&self) -> Option<NonNull<T>> {
```
littledivy, line 561: In NAPI, we implement references to these Weak, if the ref count is greater than 0 we need to "clear the weak" and prevent finalization (sorta like what's happening in `drop()` rn?) What would be the right API for this?

andreubotella: Per conversation on Discord, keeping track of the reference count and clearing the weak handle can be done in Deno, but we would need a way to prevent finalization of a `Weak`, possibly even after the first pass callback has run. Since the second pass callback will only run the finalizer if it's `Some`, preventing finalization would just involve setting it to `None`.

src/handle.rs
```diff
@@ -432,3 +449,356 @@ impl HandleHost {
...
+          v8__Local__New(sd.get_isolate_ptr(), data.cast().as_ptr()) as *const T
+        })
+      };
+      Some(local.unwrap())
+    } else {
+      None
+    }
+  }
+
+  pub fn unset_finalizer(&mut self) {
```
andreubotella, line 656: This method doesn't really have to be `mut`, since the finalizer is behind a `Cell`, but it semantically does mutate the handle. WDYT?

bartlomieju: I think this is better that using `&self` 👍 let's keep it

andreubotella:
Rebasing this to main seems to introduce a V8 crash in `weak_is_empty_after_dropping_isolate`:

```
#
# Fatal error in ../../../v8/src/handles/global-handles.cc, line 311
# Debug check failed: AsChild()->IsInUse().
```

andreubotella:
I'm not 100% sure that the semantics of when finalizers are run make sense. If a `Weak<T>` is dropped while the value hasn't been GC'd, the finalizer won't run. But if a `Weak<T>` is dropped after the value has been GC'd, the finalizer might still run. Is this something that makes sense? The finalizer can be canceled with `unset_finalizer()`, so it's not world-breaking if you don't want that behavior. Also, the alternative of unsetting the finalizer on drop by default would be a lot harder to prevent.

Also, while I was thinking of this I noticed that the `Drop` impl assumes that if the finalizer is unset, there's no second-pass callback still to run (because there was never a finalizer, or because the callback has already run), and so assumes that it's safe to drop the `Box<WeakData<T>>`. With `unset_finalizer()`, this is no longer true. I'll think of how to fix this tomorrow.

piscisaureus:
> If a `Weak<T>` is dropped while the value hasn't been GC'd, the finalizer won't run. But if a `Weak<T>` is dropped after the value has been GC'd, the finalizer might still run after it's dropped. Is this something that makes sense?

I don't think that makes sense - we should prevent the finalizer from running after the Weak has been dropped.

> Also, while I was thinking of this I noticed that the `Drop` impl assumes that if the finalizer is unset, there's no second-pass callback still to run (because there was never a finalizer, or because the callback has already run, consuming the finalizer), and so assumes that it's safe to drop the `Box<WeakData<T>>`. With `unset_finalizer()`, this is no longer true. I'll think of how to fix this tomorrow.

I'm also not super happy with the potential for memory leaks when the Isolate is dropped between first and second stage callbacks. Maybe what we should do is the following:

* Store the `IsolateHandle` in the `WeakData` rather than in the `Weak`. 
* Assign each finalizer a unique id.
* Add a `HashMap<WeakId, FinalizerFn>` to the `IsolateAnnex`.
* When a Weak is dropped, remove its entry from the hashmap.
* In the second-stage callback, look up the finalizer in the hashmap by its id. If not found, assume that the Weak has been dropped and don't call any callbacks.
* When the isolate is disposed, clear the hashmap.

andreubotella:
> I'm also not super happy with the potential for memory leaks when the Isolate is dropped between first and second stage callbacks. Maybe what we should do is the following:
> 
>     * Store the `IsolateHandle` in the `WeakData` rather than in the `Weak`.
> 
>     * Assign each finalizer a unique id.
> 
>     * Add a `HashMap<WeakId, FinalizerFn>` to the `IsolateAnnex`.
> 
>     * When a Weak is dropped, remove its entry from the hashmap.
> 
>     * In the second-stage callback, look up the finalizer in the hashmap by its id. If not found, assume that the Weak has been dropped and don't call any callbacks.
> 
>     * When the isolate is disposed, clear the hashmap.

Done. Rather than storing the `IsolateHandle` in `WeakData`, I instead used `v8::WeakCallbackInfo::GetIsolate()`.

andreubotella:
@littledivy In the end we decided to remove the `unset_finalizer` method, and instead unset the finalizer automatically when dropping a `Weak<T>`. This covers all of the NAPI requirements, right?

@piscisaureus This should be now ready.

littledivy:
>  instead unset the finalizer automatically when dropping a Weak<T>. This covers all of the NAPI requirements, right?

yup

littledivy (changes requested):
Using this branch crashes the Deno repl. (Something going on with typescript compiler ops).

```
#
# Fatal error in ../../../../rusty_v8/v8/src/objects/objects-inl.h, line 455
# Debug check failed: IsNumber().
#
#
#
#FailureMessage Object: 0x1701736c8
==== C stack trace ===============================

    0   deno                                0x000000010470d18c v8::base::debug::StackTrace::StackTrace() + 24
    1   deno                                0x0000000104714d5c v8::platform::(anonymous namespace)::PrintStackTrace() + 116
    2   deno                                0x0000000104704ce4 V8_Fatal(char const*, int, char const*, ...) + 268
    3   deno                                0x00000001047047ac std::__1::enable_if<((!(std::is_function<std::__1::remove_pointer<char>::type>::value)) && (!(std::is_enum<char>::value))) && (has_output_operator<char, v8::base::CheckMessageStream>::value), std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > >::type v8::base::PrintCheckOperand<char>(char) + 0
    4   deno                                0x00000001047e1a18 v8::Integer::Value() const + 148
    5   deno                                0x00000001046e9778 v8__Integer__Value + 24
    6   deno                                0x00000001046b31dc _ZN2v86number35_$LT$impl$u20$v8..data..Integer$GT$5value17h664286859e3a18ffE + 24
    7   deno                                0x0000000100df1d6c deno::lsp::tsc::op_respond::v8_func::h35ecb08cd383903e + 120
    8   deno                                0x0000000100e1f938 core::ops::function::Fn::call::h519fb15c654e8819 + 52
    9   deno                                0x000000010136ec84 _ZN2v88function147_$LT$impl$u20$v8..support..MapFnFrom$LT$F$GT$$u20$for$u20$extern$u20$$u22$C$u22$$u20$fn$LP$$BP$const$u20$v8..function..FunctionCallbackInfo$RP$$GT$7mapping28_$u7b$$u7b$closure$u7d$$u7d$17h75289b9897a40af5E + 332
    10  deno                                0x00000001013ea9f4 _ZN105_$LT$extern$u20$$u22$C$u22$$u20$fn$LP$A0$RP$$u20$.$GT$$u20$R$u20$as$u20$v8..support..CFnFrom$LT$F$GT$$GT$7mapping4c_fn17h0f7a4e174e2d7ed8E + 188
    11  deno                                0x00000001048444c0 v8::internal::FunctionCallbackArguments::Call(v8::internal::CallHandlerInfo) + 800
    12  deno                                0x0000000104842a8c v8::internal::MaybeHandle<v8::internal::Object> v8::internal::(anonymous namespace)::HandleApiCallHelper<false>(v8::internal::Isolate*, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::HeapObject>, v8::internal::Handle<v8::internal::FunctionTemplateInfo>, v8::internal::Handle<v8::internal::Object>, v8::internal::BuiltinArguments) + 816
    13  deno                                0x00000001048413e8 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) + 524
    14  deno                                0x0000000105e2b6c8 Builtins_CEntry_Return1_DontSaveFPRegs_ArgvOnStack_BuiltinExit + 136
    15  deno                                0x0000000105bb4314 Builtins_InterpreterEntryTrampoline + 308
    16  deno                                0x0000000105bb4314 Builtins_InterpreterEntryTrampoline + 308
    17  deno                                0x0000000105bb4314 Builtins_InterpreterEntryTrampoline + 308
    18  deno                                0x0000000105bb4314 Builtins_InterpreterEntryTrampoline + 308
    19  deno                                0x0000000105bad7d8 Builtins_JSEntryTrampoline + 184
    20  deno                                0x0000000105bad468 Builtins_JSEntry + 168
    21  deno                                0x0000000104a47c9c v8::internal::(anonymous namespace)::Invoke(v8::internal::Isolate*, v8::internal::(anonymous namespace)::InvokeParams const&) + 4376
    22  deno                                0x0000000104a48640 v8::internal::Execution::CallScript(v8::internal::Isolate*, v8::internal::Handle<v8::internal::JSFunction>, v8::internal::Handle<v8::internal::Object>, v8::internal::Handle<v8::internal::Object>) + 472
    23  deno                                0x00000001047c32e4 v8::Script::Run(v8::Local<v8::Context>, v8::Local<v8::Data>) + 1216
    24  deno                                0x00000001046eb3f4 v8__Script__Run + 92
    25  deno                                0x00000001046ca600 _ZN2v86script34_$LT$impl$u20$v8..data..Script$GT$3run28_$u7b$$u7b$closure$u7d$$u7d$17hedcb8a831580682dE + 56
    26  deno                                0x00000001046d88f0 _ZN2v85scope27HandleScope$LT$$LP$$RP$$GT$10cast_local17h64b34c9057b07c6dE + 72
    27  deno                                0x00000001046b4434 _ZN2v86script34_$LT$impl$u20$v8..data..Script$GT$3run17h89116a7209c48deaE + 64
    28  deno                                0x00000001045bfc0c deno_core::runtime::JsRuntime::execute_script::h157dd1708db9f406 + 512
    29  deno                                0x0000000100d52b80 deno::lsp::tsc::request::h29a86fe263d8ab14 + 1340
    30  deno                                0x0000000100d29630 _ZN4deno3lsp3tsc8TsServer3new28_$u7b$$u7b$closure$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$17h4c11f6a6e80c014fE + 900
    31  deno                                0x0000000100f2448c _ZN97_$LT$core..future..from_generator..GenFuture$LT$T$GT$$u20$as$u20$core..future..future..Future$GT$4poll17h707b9a8b1146c4e8E + 84
    32  deno                                0x0000000100a3a350 _ZN72_$LT$core..pin..Pin$LT$P$GT$$u20$as$u20$core..future..future..Future$GT$4poll17h1e2f367c7e90200aE + 88
    33  deno                                0x0000000100bfb574 _ZN5tokio7runtime15basic_scheduler9CoreGuard8block_on28_$u7b$$u7b$closure$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$17h20239f57c9824759E + 72
    34  deno                                0x000000010118486c _ZN5tokio4coop11with_budget28_$u7b$$u7b$closure$u7d$$u7d$17h5b94841875dd443bE + 196
    35  deno                                0x0000000100a55708 _ZN3std6thread5local17LocalKey$LT$T$GT$8try_with17heb4e8f9baffa8d6bE + 188
    36  deno                                0x0000000100a4f9f0 _ZN3std6thread5local17LocalKey$LT$T$GT$4with17h7ee4b6176ab330f5E + 48
    37  deno                                0x0000000100bfb428 _ZN5tokio7runtime15basic_scheduler9CoreGuard8block_on28_$u7b$$u7b$closure$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$17hf11bb8487b25d55dE + 172
    38  deno                                0x0000000100bf354c tokio::runtime::basic_scheduler::Context::enter::h7faa10e32a8bb836 + 308
    39  deno                                0x0000000100bf9034 _ZN5tokio7runtime15basic_scheduler9CoreGuard8block_on28_$u7b$$u7b$closure$u7d$$u7d$17h3fb3f58991aba375E + 380
    40  deno                                0x0000000100bfd6c0 _ZN5tokio7runtime15basic_scheduler9CoreGuard5enter28_$u7b$$u7b$closure$u7d$$u7d$17ha03d9cf4d5dfbc31E + 48
    41  deno                                0x00000001009bb458 _ZN5tokio6macros10scoped_tls18ScopedKey$LT$T$GT$3set17hedbcace6b9a4df13E + 128
    42  deno                                0x0000000100bfd368 tokio::runtime::basic_scheduler::CoreGuard::enter::hfbe77a5b314a5684 + 284
    43  deno                                0x0000000100bf8038 tokio::runtime::basic_scheduler::CoreGuard::block_on::hf7a1f36eeb94d7fa + 56
    44  deno                                0x0000000100bf0c70 tokio::runtime::basic_scheduler::BasicScheduler::block_on::hf19a33b65c602371 + 204
    45  deno                                0x000000010104fd40 tokio::runtime::Runtime::block_on::h40ace54bea16a12c + 152
    46  deno                                0x0000000100d29254 _ZN4deno3lsp3tsc8TsServer3new28_$u7b$$u7b$closure$u7d$$u7d$17h54152d3e4debc869E + 168
    47  deno                                0x0000000100ae5820 std::sys_common::backtrace::__rust_begin_short_backtrace::h47533819778b5f96 + 32
    48  deno                                0x0000000101167a8c _ZN3std6thread7Builder15spawn_unchecked28_$u7b$$u7b$closure$u7d$$u7d$28_$u7b$$u7b$closure$u7d$$u7d$17hda4235906f1b64d0E + 32
    49  deno                                0x00000001008b0484 _ZN115_$LT$core..panic..unwind_safe..AssertUnwindSafe$LT$F$GT$$u20$as$u20$core..ops..function..FnOnce$LT$$LP$$RP$$GT$$GT$9call_once17h99da90c9c54a2aa7E + 32
    50  deno                                0x00000001008d55cc std::panicking::try::do_call::h732e87fb06b765f5 + 68
    51  deno                                0x00000001008fb9e4 __rust_try + 32
    52  deno                                0x00000001008cd518 std::panicking::try::hb4cfc88c580ffc43 + 116
    53  deno                                0x00000001013d02bc std::panic::catch_unwind::hf674a37210def406 + 32
    54  deno                                0x0000000101166e7c _ZN3std6thread7Builder15spawn_unchecked28_$u7b$$u7b$closure$u7d$$u7d$17h456ccffcf1429df8E + 280
    55  deno                                0x0000000100e21ed4 _ZN4core3ops8function6FnOnce40call_once$u7b$$u7b$vtable.shim$u7d$$u7d$17h4e4a95ed1146f2b7E + 24
    56  deno                                0x00000001062c7298 std::sys::unix::thread::Thread::new::thread_start::ha92b558f2b29b4cc + 48
    57  libsystem_pthread.dylib             0x00000001bad41240 _pthread_start + 148
    58  libsystem_pthread.dylib             0x00000001bad3c024 thread_start + 8
fish: Job 1, 'target/debug/deno' terminated by signal SIGTRAP (Trace or breakpoint trap)
```

andreubotella:
> Using this branch crashes the Deno repl. (Something going on with typescript compiler ops).

Not caused by this PR. It seems to be triggered by updating the V8 submodule to `10.0-lkgr-denoland`, but the ultimate cause is Deno keeping an `opIds` object in JS-land that doesn't get synchronized with the actual ops. I submitted a PR to fix that at the root in denoland/deno#14224.

andreubotella:
Since there's been a lot of back and forth on some topics, the resulting behavior might not be clear, so I've squashed all the commits and written a single commit message that hopefully informs the reviews.

andreubotella:
Currently, this PR runs any remaining finalization callbacks as the isolate is being disposed, in particular after dropping the isolate's annex, which as @littledivy found, triggers a crash when you run `v8::Global::from_raw` or `v8::Weak::from_raw` in such a finalizer. But as I looked into this, I found something that is more concerning.

`v8::Global` instances are not tied to the isolate's lifetime, and so they can outlive it. This is not a problem because in order to use them you usually need a mutable reference to the corresponding isolate, and the few cases where you don't need one will panic if you use them after the corresponding isolate is dropped. However, the fact that dropping the isolate will run _every_ registered finalizer, means that you can access a `v8::Global` to some object from a finalizer tied to that object. Furthermore, since the finalizer is passed a `&mut Isolate`, you might expect to be able to use the object. Trying that will panic, but if we fix the above crash by running the finalizers before dropping the annex, that might not be the case.

The V8 C++ API guarantees nothing about whether a finalizer will run. This PR currently guarantees that the finalizer will always run, but that causes the above mess. I think we should instead run the finalizers "on a best-effort basis", i.e. as long as there isn't a `v8::Global` keeping the object alive at the time that the isolate is dropped. To do this we could call `v8::Isolate::low_memory_notification()`, which triggers a full synchronous GC. Since this is expensive, we would only trigger that low-memory notification if the finalizer map is not empty.

The tricky part is how to run the GC after all scopes and contexts have been dropped, so you could have finalizers on the global object.

littledivy (approved):
LGTM!

bartlomieju (approved):
LGTM, thank you @andreubotella, this is great work!

PR #897, opened 2022-02-11
Add v8::Object::HasOwnProperty bindings

littledivy:


littledivy:
Needed for https://github.com/denoland/deno/pull/13633

src/binding.cc
```diff
@@ -1158,6 +1158,12 @@ MaybeBool v8__Object__HasIndex(const v8::Object& self,
       ptr_to_local(&self)->Has(ptr_to_local(&context), index));
 }
 
+MaybeBool v8__Object__HasOwnProperty(const v8::Object& self, const v8::Context& context,
+                          const v8::Value& key) {
+  return maybe_to_maybe_bool(
+      ptr_to_local(&self)->HasOwnProperty(ptr_to_local(&context), ptr_to_local(&key).As<v8::Name>()));
```
piscisaureus, line 1164: This cast is not safe I think.

src/binding.cc
```diff
@@ -1158,6 +1158,12 @@ MaybeBool v8__Object__HasIndex(const v8::Object& self,
       ptr_to_local(&self)->Has(ptr_to_local(&context), index));
 }
 
+MaybeBool v8__Object__HasOwnProperty(const v8::Object& self, const v8::Context& context,
+                          const v8::Value& key) {
```
piscisaureus, line 1162: Follow the C++ API unless there are overriding concerns.

```suggestion
                          const v8::Name& key) {
```

src/object.rs
```diff
@@ -99,6 +99,11 @@ extern "C" {
     context: *const Context,
     index: u32,
   ) -> MaybeBool;
+  fn v8__Object__HasOwnProperty(
+    this: *const Object,
+    context: *const Context,
+    key: *const Value,
```
piscisaureus, line 105:
```suggestion
    key: *const Name,
```

tests/test_api.rs
```diff
@@ -1550,6 +1550,7 @@ fn object() {
     assert_ne!(id, 0);
 
     assert!(object.has(scope, n1.into()).unwrap());
+    assert!(object.has_own_property(scope, n1.into()).unwrap());
```
piscisaureus, line 1553: Also add a test where `has_own_property()` returns false.

piscisaureus:
LGTM

piscisaureus (approved):

PR #898, opened 2022-02-11
Add v8::FunctionCallbackInfo::NewTarget bindings

littledivy:


src/function.rs
```diff
@@ -215,6 +218,13 @@ impl<'s> FunctionCallbackArguments<'s> {
         .unwrap()
     }
   }
+
+  /// For construct calls, this returns the "new.target" value.
```
piscisaureus, line 222: What value does it return for non-construct calls?

tests/test_api.rs
```diff
@@ -1949,6 +1949,18 @@ fn fn_callback(
   rv.set(s.into());
 }
 
+fn fn_callback_new(
+  scope: &mut v8::HandleScope,
+  args: v8::FunctionCallbackArguments,
+  mut rv: v8::ReturnValue,
+) {
+  assert_eq!(args.length(), 0);
+  assert!(!args.new_target().is_undefined());
```
piscisaureus, line 1958: Assert that it is an object.
I'd also try to set a property on the object, and then assert that the object returned by `new f2()` has that property.

piscisaureus:
LGTM but I'd like the test to be slightly more rigorous.

src/function.rs
littledivy: `undefined`

tests/test_api.rs
```diff
@@ -2070,6 +2085,16 @@ fn function() {
...
+
+    let function = v8::Function::builder(fn_callback_new).build(scope).unwrap();
+    let name = v8::String::new(scope, "f2").unwrap();
+    global.set(scope, name.into(), function.into()).unwrap();
+    let f2: v8::Local<v8::Object> =
+      eval(scope, "new f2()").unwrap().try_into().unwrap();
+    let key = v8::String::new(scope, "works").unwrap();
+    let value = f2.get(scope, key.into()).unwrap();
+    assert!(value.is_boolean());
+    assert_eq!(value.boolean_value(scope), true);
```
piscisaureus, line 2097:
```suggestion
    assert!(value.boolean_value(scope));
```

piscisaureus (approved):
LGTM

PR #901, opened 2022-02-16
Raw pointer in external references

littledivy:


tests/test_api.rs
```diff
@@ -2930,20 +2946,34 @@ fn snapshot_creator() {
...
+  let mut refs_mut = EXTERNAL_REFERENCES.lock().unwrap();
+  // Allocate externals for the test.
+  let external_ptr = Box::into_raw(vec![0_u8, 1, 2, 3, 4].into_boxed_slice())
+    as *mut [u8] as *mut c_void;
+  // Push them to the external reference table.
+  refs_mut.push(v8::ExternalReference {
+    pointer: external_ptr,
+  });
+  let refs: &'static v8::ExternalReferences =
+    Box::leak(Box::new(refs_mut.clone()));
```
piscisaureus, line 2972: Why does this need to be leaked?

src/external_references.rs
```diff
@@ -31,6 +33,14 @@ impl ExternalReferences {
   pub fn as_ptr(&self) -> *const intptr_t {
     self.null_terminated.as_ptr()
   }
+
+  pub fn push(&mut self, reference: ExternalReference) {
```
piscisaureus, line 37: Adding this API seems unnecessarily ad-hoc.
For testing purposes, I'd suggest to just build a `Vec<ExternalReference>` and convert that to a `ExternalReferences` object.

tests/test_api.rs
```diff
@@ -2930,20 +2946,34 @@ fn snapshot_creator() {
 }
 
 lazy_static! {
-  static ref EXTERNAL_REFERENCES: v8::ExternalReferences =
-    v8::ExternalReferences::new(&[v8::ExternalReference {
-      function: fn_callback.map_fn_to()
-    }]);
+  static ref EXTERNAL_REFERENCES: Mutex<v8::ExternalReferences> =
```
piscisaureus, line 2949: I don't think we need this Mutex. See below.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "v8"
-version = "0.39.0"
+version = "0.40.0"
```
piscisaureus, line 3: Unrelated?

tests/test_api.rs
littledivy: Because of the `'static` lifetime requirement on the [`Allocated`](https://github.com/denoland/rusty_v8/blob/main/src/support.rs#L375) type.

tests/test_api.rs
```diff
@@ -2930,20 +2946,34 @@ fn snapshot_creator() {
...
   let _setup_guard = setup();
+  let mut refs_mut = EXTERNAL_REFERENCES.lock().unwrap();
+  // Allocate externals for the test.
+  let external_ptr = Box::into_raw(vec![0_u8, 1, 2, 3, 4].into_boxed_slice())
+    as *mut [u8] as *mut c_void;
+  // Push them to the external reference table.
+  refs_mut.push(v8::ExternalReference {
+    pointer: external_ptr,
+  });
+  let refs: &'static v8::ExternalReferences =
```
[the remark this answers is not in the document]
littledivy: Sure, done. We still need to leak it to satisfy lifetime

piscisaureus (approved):

PR #902, opened 2022-02-16
feat: v8::Global::from_raw, v8::Global::into_raw

bartlomieju:


src/handle.rs
```diff
@@ -135,7 +135,9 @@ impl<T> Global<T> {
 
   /// Implementation helper function that contains the code that can be shared
   /// between `Global::new()` and `Global::clone()`.
-  unsafe fn new_raw(isolate: *mut Isolate, data: NonNull<T>) -> Self {
+  ///
+  /// Also allows to create a `Global` from raw pointer.
+  pub unsafe fn new_raw(isolate: *mut Isolate, data: NonNull<T>) -> Self {
```
piscisaureus, line 140: from_raw would be more idiomatic

piscisaureus (approved):
LGTM with one comment

src/handle.rs
bartlomieju: Agreed

andreubotella:
While looking into `Global::into_raw` and `from_raw` as part of #895, I noticed that `from_raw` doesn't restore the original global, but it creates a new one without resetting the previous one, just like `Global::new(isolate, &old_global)` does. This would cause memory leaks.

bartlomieju:
But the previous global is being consumed by `Global::into_raw()` call, why would that be a problem?

andreubotella:
```rust
let old_global: v8::Global<v8::Object> = something();
let new_global = v8::Global::new(isolate, old_global);
```

The code above has the `Global` constructor take ownership of the old global, create the new global, and it subsequently drops the old global. `Global::from_raw` doesn't reset the old pointer –in fact, that would be unsound if the `Global` constructor had been called with a reference. Instead, the old global is dropped in the body of `Global::new`, and that drop calls `v8__Global__Reset`.

If you call `Global::into_raw` and then pass the pointer to `Global::from_raw`, the old pointer never gets reset.

alshdavid:
@andreubotella. Sorry for reviving an old thread - just running into this and was a bit surprised. Is there guidance on how to get around this leak?

```rust
// Run this 10_000 times
fn run_in_loop() -> anyhow::Result<()> {
    let mut isolate = v8::Isolate::new(v8::CreateParams::default());

    {
        let handle_scope = &mut v8::HandleScope::new(&mut isolate);
        let context = v8::Context::new(handle_scope, Default::default());
        let context_scope = &mut v8::ContextScope::new(handle_scope, context);

        let global_raw = {
            let scope = &mut v8::HandleScope::new(context_scope);
            let local_str = v8::String::new(scope, "Hello world").unwrap();
            let global_str = v8::Global::new(scope, local_str);
            let global_raw = global_str.into_raw(); // Leaks value
            global_raw
        };

        {
            let scope = &mut v8::HandleScope::new(context_scope);
            let global_handle = unsafe { v8::Global::from_raw(scope, global_raw) };
            drop(global_handle) // previously allocated value is never GC'd
        }
    }

    isolate.request_garbage_collection_for_testing(v8::GarbageCollectionType::Full);
    Ok(())
}
```

Running this code in a loop, memory usage continuously increases as the value within `into_raw()` is never cleaned up by GC

PR #904, opened 2022-02-21
Add bindings for String::kMaxLength / TypedArray::kMaxLength

RomainMuller:
Those are useful for pre-checking buffer sizes before attempting to create
a new String or typed array (or ArrayBuffer). Also useful for pre-allocating
possibly large chunks of memory safely.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -457,6 +457,8 @@ bool v8__Data__IsFunctionTemplate(const v8::Data& self) {
   return self.IsFunctionTemplate();
 }
 
+size_t v8__TypedArray__kMaxLength() { return v8::TypedArray::kMaxLength; }
+
 bool v8__Value__IsUndefined(const v8::Value& self) {
   return self.IsUndefined();
 }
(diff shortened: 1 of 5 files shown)


RomainMuller:
Note: wanted to have a test case around attempting to create a `TypedArray`/`ArrayBuffer` that is too long, but unlike `v8::String` factories, which return `null`/`None` when the buffer's too large, the `ArrayBuffer` constructor raises a fatal in this situation.

piscisaureus:
> Note: wanted to have a test case around attempting to create a `TypedArray`/`ArrayBuffer` that is too long, but unlike `v8::String` factories, which return `null`/`None` when the buffer's too large, the `ArrayBuffer` constructor raises a fatal in this situation.

Can I suggest to add a smoke test for `TypedArray::max_lenght()` instead. Just assert that the returned value is somewhere in the range `2**30 <= max_length < 2**32`.

piscisaureus:
> Note: wanted to have a test case around attempting to create a `TypedArray`/`ArrayBuffer` that is too long, but unlike `v8::String` factories, which return `null`/`None` when the buffer's too large, the `ArrayBuffer` constructor raises a fatal in this situation.

Interestingly on the JS side of things, a non-fatal RangeError is thrown:

```
> new Uint8Array(2**33-1)
Uncaught RangeError: Invalid typed array length: 8589934591
    at new Uint8Array (<anonymous>)
    at <anonymous>:2:1
```

RomainMuller:
Added both:
- A range check test
- A JS test around the `RangeError` to work around the fatal

piscisaureus (approved):
LGTM, thanks @RomainMuller!

PR #906, opened 2022-02-23
fix: Global::from_raw() should not create a new Global

piscisaureus:


piscisaureus:
@andreubotella wdyt?

andreubotella:
> @andreubotella wdyt?

LGTM. I don't think this can be tested before landing #895, but with it, the following test fails on main and passes with this patch:

```rust
#[test]
fn global_from_into_raw() {
  let _setup_guard = setup();

  let context = v8::Context::new(scope);
  let scope = &mut v8::ContextScope::new(scope, context);

  let (raw, weak) = {
    let scope = &mut v8::HandleScope::new(scope);
    let local = v8::Object::new(scope);
    let global = v8::Global::new(scope, local);

    let weak = v8::Weak::new(scope, &global);
    let raw = global.into_raw();
    (raw, weak)
  };

  eval(scope, "gc()").unwrap();
  assert!(!weak.is_empty());

  {
    let reconstructed = unsafe { v8::Global::from_raw(scope, raw) };

    let global_from_weak = weak.to_global(scope).unwrap();
    assert_eq!(global_from_weak, reconstructed);
  }

  eval(scope, "gc()").unwrap();
  assert!(weak.is_empty());
}
```

Also, it might be worth mentioning in the rustdoc for `Global::from_raw` that the isolate passed must be the same in which the global was created, and that it can't be called twice on the same pointer.

bartlomieju (approved):
Good catch, LGTM

piscisaureus:
> LGTM. I don't think this can be tested before landing #895, but with it, the following test fails on main and passes with this patch:

Yes, let's actually land that test in or after #895.

PR #911, opened 2022-03-07
feat(isolate): expose get/set_data

AaronO:
Since that allows us to store data in the isolate without going through the Annex (less efficient) especially when in practice the cardinality is small (i.e: 2 in deno_core)


piscisaureus:
I think a range check would also be nice.

PR #917, opened 2022-03-09
feat: new_raw APIs avoiding MapFnTo path

AaronO:
Unblocks https://github.com/denoland/deno/pull/13861

The MapFnTo trait is quite contrived and can't be turned into a trait object thus blocking declarative ops

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -362,8 +362,12 @@ pub struct FunctionBuilder<'s, T> {
 impl<'s, T> FunctionBuilder<'s, T> {
   /// Create a new FunctionBuilder.
   pub fn new(callback: impl MapFnTo<FunctionCallback>) -> Self {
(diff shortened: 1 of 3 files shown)


littledivy (approved):
LGTM...just needs a test.

piscisaureus:
I see the point of this PR now. Yes we need a test - even if there isn't much to test.

AaronO:
> I see the point of this PR now. Yes we need a test - even if there isn't much to test.

Added some tests that are mostly copies of the non-raw tests, pretty straightforward but should at least catch regressions or breaking API changes

piscisaureus:
> Added some tests that are mostly copies of the non-raw tests, pretty straightforward but should at least catch regressions or breaking API changes

That's fine, nobody said that tests need to be creative.

piscisaureus (approved):
LGTM

PR #921, opened 2022-03-11
Revert "Upgrade to ICU 70 (#919)"

littledivy:
This reverts commit 184467a9ccb70c0f1ad55fadbe88d951da36dbc7.

crates.io size limit exceeded -> https://github.com/denoland/rusty_v8/runs/5503750428?check_suite_focus=true#step:19:13151


kitsonk:
We should get our limit raised, we have had it done before, versus reverting the commit, as we need to be able to update the ICU in the future.

lucacasonato:
#923 fixes this more elegantly :-)

PR #923, opened 2022-03-11
Ignore more unnecessary ICU data

lucacasonato:
The latest ICU update increased the file size of the crate to above 20 MB again. This commit removes an ICU data file from the package that is unneeded for the build.

I tested the build still works by doing:

```
$ export V8_FROM_SOURCE=1
$ cargo package
```

And it still works.


littledivy (approved):
LGTM

PR #924, opened 2022-03-15
Add feature flag for using custom libcxx

### LINKED ISSUE #920: Change pre-generated binaries to use system libcpp?
Hello maintainers :wave: First of all thank you for your work in these bindings, they are great.

I've hit a wall integrating v8 into a codebase with further cpp code, in my case Rocksdb. v8 and rocksdb don't play along due to how they want to use libcpp. V8 usually uses a custom libcpp by default, and that's what the Deno based V8 bindings use.
On linux environments using g++ Rocksdb can successfully use the libcpp that v8 brings along, but on MacOS that doesn't work and there's a segmentation fault when Rocksdb tries to use an unordered_map.
After some struggle I was able to make Rocksdb and V8 coexist on MacOS by forcing V8 to use the system libcpp. But maintaining that would require forking the Deno v8 bindings and all that entails.

I suppose my question is whether it would be acceptable (or better?) to switch the rusty_v8 default flags to be `use_custom_libcxx=false`?

arthurprs:
The feature `use_custom_libcxx` is enabled by default. If the feature `use_custom_libcxx` is disabled the crate will link the system libcxx instead.

Once this is merged and released, a followup PR will be created for:
* `deno_core`, so it re-exposes the feature flag. Thus allowing `deno_core` users to disable custom libcxx using `default_features=false`.
* `serde_v8`, so it depends on `rusty_v8` with `default_features=false`.

Closes #920

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -12,7 +12,6 @@ use which::which;
 
 fn main() {
   println!("cargo:rerun-if-changed=src/binding.cc");
-  println!("cargo:rerun-if-changed=build.rs");
 
   // These are all the environment variables that we check. This is
   // probably more than what is needed, but missing an important
--- a/build.rs
+++ b/build.rs
@@ -12,6 +12,7 @@ use which::which;
 
 fn main() {
   println!("cargo:rerun-if-changed=src/binding.cc");
+  println!("cargo:rerun-if-changed=build.rs");
 
   // These are all the environment variables that we check. This is
   // probably more than what is needed, but missing an important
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -12,6 +12,7 @@ use which::which;
 
 fn main() {
   println!("cargo:rerun-if-changed=src/binding.cc");
+  println!("cargo:rerun-if-changed=build.rs");
```
piscisaureus, line 15: This is unnecessary (build.rs is always taken into account).

piscisaureus, line 15:
```suggestion
```

piscisaureus (approved):
LGTM, thank you @arthurprs!

PR #925, opened 2022-03-16
perf: avoid double boxing when Arc/Box/Rc is stored in Isolate slot

piscisaureus:


bartlomieju (approved):
LGTM, two questions though:
1. is there any measurable performance gain from this change?
2. could you add `// SAFETY:` comments to the `unsafe` functions? I think we should enable `missing_safety_docs` lint similar to https://github.com/denoland/deno/pull/13870 but that would probably require a lot of work

src/isolate.rs
```diff
@@ -1030,3 +1042,128 @@ where
...
+  fn write(&mut self, _bytes: &[u8]) {
+    panic!("TypeIdHasher::write() called unexpectedly");
+  }
+
+  #[inline]
+  fn write_u64(&mut self, value: u64) {
+    // `TypeId` values are actually 64-bit values which themselves come out of
+    // some hash function, so it's unnecessary to shuffle their bits further.
+    assert_eq!(size_of::<TypeId>(), size_of::<u64>());
+    assert_eq!(align_of::<TypeId>(), size_of::<u64>());
```
littledivy, line 1063:
```rust

const _: () = {
   // `TypeId` values are actually 64-bit values which themselves come out of
   // some hash function, so it's unnecessary to shuffle their bits further.
   assert!(size_of::<TypeId>() == size_of::<u64>());
   assert!(align_of::<TypeId>() == size_of::<u64>());
};
```
This can be a constant assert.

littledivy (approved):

piscisaureus:
> is there any measurable performance gain from this change?

Not super significant. It shaved off 1 or 2 ns from the serde_v8 the fastest serde_v8 benchmarks `ser::de_bool_v8` IIRC.
Admittedly, that was was after I first made them perform significantly worse by putting a v8::String lookup cache in one of the slots.

It might be worthwhile to rethink the slots strategy. Despite the fact that the HashMap doesn't actually hash anything (the keys are pre-hashed), it nonetheless has measurable overhead.
Plus the ergonomics aren't that great either - it might be just as good to use a few thread local statics.

PR #926, opened 2022-03-17
feat: ArrayBuffer::new_backing_store_from_ptr

DjDeveloperr:
useful for https://github.com/denoland/deno/pull/13633
which currently uses a "hack" to create externally backed array buffers: https://github.com/littledivy/deno/blob/4fb9b90da6bcab9690ba2113083520f709a0b636/cli/napi/js_native_api.rs#L312
also needed to make zero-copy arraybuffers (returned from FFI. there is UnsafePointerView#getArrayBuffer, which currently copies)

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -513,14 +513,6 @@ fn get_isolate_from_handle() {
   check_eval(scope, Some(false), "3.3 / 3.3");
 }
 
-pub extern "C" fn backing_store_deleter_callback(
-  _data: *mut c_void,
-  _byte_length: usize,
-  _deleter_data: *mut c_void,
-) {
-  // no-op
-}
-
 #[test]
 fn array_buffer() {
   let _setup_guard = setup();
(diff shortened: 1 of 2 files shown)


tests/test_api.rs
```diff
@@ -513,6 +513,14 @@ fn get_isolate_from_handle() {
   check_eval(scope, Some(false), "3.3 / 3.3");
 }
 
+pub extern "C" fn backing_store_deleter_callback(
+  _data: *mut c_void,
+  _byte_length: usize,
+  _deleter_data: *mut c_void,
+) {
+  // no-op
```
aapoalas, line 521: Should part of the test be to assert based on the data here? Though, assertions here would try to unwind through C++ so that'd be pretty bad. Maybe cout based assertions could be utilized?

DjDeveloperr: Not sure I understand, why should we assert the data here? This finalized callback doesn't need to be utilized since the pointer is to slice owned by Rust, it will be freed when it goes out of scope - which is pretty bad though. Maybe I should use manually allocated memory for this test using `std::alloc`, such that I can assert data before freeing it?

aapoalas: Hmmm, well definitely the intended use case for this function is in general about memory more akin to `std::alloc`, ie. memory outside of Rust's borrow checker / memory management.

The current test asserts that the API can be called but not if it actually functionally works. That is, of course, more about V8 testing but one could argue that the bindings should still be tested with some code functionality as well.

So in short what I mean is that the callback could assert that the arguments are as expected: length is the same, data can be asserted to be equal to what was originally there, and that the `deleter_data` has passed through V8 unmolested (well, that is if we want to add some eg. `Box` as the data).

The eventual use-case for this API may well end up looking like this, a complete no-op as the deleter callback. But it might also be that some use-case will be found where a custom deleter callback is indeed needed. The callback argument assertions would be tests in preparation for that.

TimDiekmann:
We are highly interested in this change. Is there a way to help out to get this ready? What's needed to be able to merge this?

piscisaureus:
Before I accept this PR:

I really don't want to add yet another flavor of `ArrayBuffer::new_backing_store_that_does_something_very_specific()`.
We should make an attempt to add a general-purpose, safe, and rust-idiomatic `BackingStore` solution.

For example a trait that can turn any object into a backing store:

```rust
pub trait CustomBackingStore {
  // Returns the bytes that are available in v8.
  fn buffer(&mut self) -> &mut [u8];
}

impl ArrayBuffer {
  pub fn with_backing_store<'s, B: CustomBackingStore>(
    scope: &mut HandleScope<'s>,
    backing_store: B,
  ) -> Local<'s, ArrayBuffer> {
    // magic that makes it work.
  }
}

// Now you can go wild
struct CrazyBackingStore {
  stuff: Box<Box<Box<Box<[u8]>>>>;
  farewell: String
}
impl CustomBackingStore for CrazyBackingStore {
  fn buffer(&mut self) -> &mut [u8] {
    &mut (****self.stuff)[9..17]
  }
}
impl Drop for CrazyBackingStore {
  fn drop(&mut self) {
    eprinln!("See ya never! {}", &self.farewell);
  }
}

fn main() {
  // ...
  let fun = CrazyBackingStore { 
    stuff: Box::new(Box::new(Box::new(&b"long story this will be in the backing store"[..].into()))),
    farewell: "😂😂😂".to_owned()
  };

 // This works!
 let ab = v8::ArrayBuffer::with_backing_store(&mut scope, fun);
}
```

piscisaureus:
That said, I know n-api exposes a very low-level interface, exactly like this newly proposed API. 
If we want to emulate napi, we'll need this. So, LGTM

piscisaureus (approved):

PR #937, opened 2022-04-13
feat: `Context` slots

andreubotella:
Builds on top of #895.


src/context.rs
```diff
@@ -84,4 +105,149 @@ impl Context {
...
+      std::ptr::eq(isolate, unsafe { v8__Context__GetIsolate(self) }),
+      "attempted to use Context slots with the wrong Isolate"
+    );
+
+    let num_data_fields =
+      unsafe { v8__Context__GetNumberOfEmbedderDataFields(self) } as c_int;
+    if num_data_fields <= Self::ANNEX_SLOT {
+      let annex = Box::new(ContextAnnex {
+        slots: Default::default(),
+        // To be replaced
```
bartlomieju, line 123: What do you want to replace it with?

bartlomieju (commented):
@piscisaureus PTAL

src/context.rs
andreubotella: That comment wasn't a TODO, it was a note that `weak` gets set to a weak handle to the context further down the function. I'll reword it.

andreubotella:
While trying to implement multi-realm async ops based on this PR, I realized that this change keeps a `v8::Weak` alive since the moment a slot is created until the context is GC'd. But building a snapshot will crash if you have any global or weak handle alive, which is not a consequence you'd expect from using context slots.

Should we have a `clear_all_slots` method, with an explanation of why it's needed in the documentation?

bartlomieju:
> While trying to implement multi-realm async ops based on this PR, I realized that this change keeps a `v8::Weak` alive since the moment a slot is created until the context is GC'd. But building a snapshot will crash if you have any global or weak handle alive, which is not a consequence you'd expect from using context slots.
> 
> Should we have a `clear_all_slots` method, with an explanation of why it's needed in the documentation?

If we can't do this automatically, then `clear_all_slots` sounds fine.

andreubotella:
I've now implemented `clear_all_slots`. PTAL

PR #938, opened 2022-04-18
feat: add binding for Context::SetPromiseHooks

legendecas:
Add support for `Context::SetPromiseHooks` for follow-up works of https://github.com/denoland/deno/pull/8209.

### DIFF
--- a/.gn
+++ b/.gn
@@ -36,6 +36,7 @@ default_args = {
   clang_use_chrome_plugins = false
   v8_monolithic = false
   v8_enable_snapshot_compression = false
+  v8_enable_javascript_promise_hooks = true
   v8_use_external_startup_data = false
   v8_use_snapshot = true
   is_component_build = false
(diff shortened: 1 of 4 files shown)


legendecas:
As far as I could tell, add build flag `v8_enable_javascript_promise_hooks` should not break the `cached_data_version_tag` test. I can not reproduce it locally. 🤔

bradfrosty:
@legendecas nice work on this PR! I'm curious about the web compatibility here, and the ability to leverage this API to replace something like zone.js. I know when they tried to standardize zones it ultimately fizzled out. Do you know if there's any reason why this API couldn't be leveraged within the browser?

legendecas:
@bradfrosty for what I could tell, people at TC39 are concerned about the nature of the dynamic scope these APIs introducing to the language.

legendecas:
Looks like I still can not reproduce the problem locally, or on my own fork's GitHub Action. Rebasing and trying to see if the problem still exists.

legendecas:
anyone can help to approve the workflow to run? thanks

piscisaureus (approved):

PR #946, opened 2022-04-20
feat: v8::ArrayBuffer::new_backing_store_from_vec

AaronO:
So we can avoid `.into_boxed_slice()` calls which may require a (costly) memmove when instantiating from a `Vec`

## Notes

- In theory, dropping & deallocating a Vec depends on its layout, which is not fully restored in the deleter (since we simplify len=cap, which may be untrue)
- We could pass the cap in the deleter_data void ptr and transmute to `usize`
- But after double-checking rust's src the GlobalAllocator/SystemAllocator `dealloc` does not depend on layout and simply calls `free` (or `HeapFree`) as I expected
- So in practice this shouldn't be an issue

### DIFF
--- a/src/array_buffer.rs
+++ b/src/array_buffer.rs
@@ -459,7 +459,7 @@ impl ArrayBuffer {
         data_ptr,
         byte_length,
         vec_deleter_callback,
-        capacity as *c_void,
+        capacity as *mut c_void,
       ))
     }
   }
--- a/src/array_buffer.rs
+++ b/src/array_buffer.rs
@@ -247,7 +247,7 @@ pub unsafe extern "C" fn vec_deleter_callback(
   byte_length: usize,
   deleter_data: *mut c_void,
 ) {
-  let capacity: usize = std::mem::transmute(deleter_data);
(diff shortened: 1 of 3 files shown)


src/array_buffer.rs
```diff
@@ -242,6 +243,18 @@ pub unsafe extern "C" fn backing_store_deleter_callback(
...
 
+pub unsafe extern "C" fn vec_deleter_callback(
+  data: *mut c_void,
+  byte_length: usize,
+  _deleter_data: *mut c_void,
+) {
+  drop(Vec::from_raw_parts(
+    data as *mut u8,
+    byte_length,
+    byte_length,
```
andreubotella, line 254: One of the safety requirements of `Vec::from_raw_parts` is that the passed capacity must be the vector's capacity. The capacity should probably be leaked as a `Box<usize>` and passed as the deleter data.

ry:
We really need some sort of LSAN check in rusty_v8... it's not at all clear if this isn't causing a memory leak - we only know from inspection.

ry (approved):
LGTM

PR #952, opened 2022-04-26
add FunctionTemplate::instance_template binding

SpencerSharkey:
This adds bindings for the `FunctionTemplate->InstanceTemplate()` method, which may be used to get a handle to the object template used when the function is used as a constructor. 

I added a test to ensure you can set internal fields and read them from the constructed objects later.


bartlomieju (approved):
LGTM, thank you @SpencerSharkey

PR #953, opened 2022-04-28
feat: Add `v8::icu::get_language_tag()`

### LINKED ISSUE #797: Expose default locale
Per discussion on discord:

we need a rusty API for the following:
- get the default locale ID (`uprv_getDefaultLocaleID`)
- get language and script from that locale ID (https://unicode-org.github.io/icu/userguide/locale/#examples)

cc @lucacasonato

AnInternetTroll:
closes #797


src/binding.cc
```diff
@@ -3148,3 +3149,22 @@ void v8__CompiledWasmModule__DELETE(v8::CompiledWasmModule* self) {
...
+// icu
+
+extern "C" {
+
+void icu_get_default_locale(char* output) {
+  const icu_71::Locale& default_locale = icu::Locale::getDefault();
+  UErrorCode status = U_ZERO_ERROR;
+  auto tag = default_locale.toLanguageTag<std::string>(status);
+  std::copy(tag.begin(), tag.end(), output);
+  output[tag.size()] = '\0';
```
ry, line 3162: This seems a little dangerous. Can you add an assert that tag.length is less than 1024 - 1?

ry:
What do you need this for?

bartlomieju:
> What do you need this for?

Needed for https://github.com/denoland/deno/pull/12322

littledivy (commented):
`icu_get_default_locale` is doing 3 allocations! Is there a reason? You can just not copy at all:

```c
const char* icu_get_default_locale(size_t* len) {
   // ...
   *len = tag.size();
   return tag.data();
}
```

```rust
let mut len = 0;
let locale: *const u8 = unsafe { icu_get_default_locale(&mut len) };
let bytes = unsafe { std::slice::from_raw_parts(locale, len) };
unsafe { String::from_utf8_unchecked(bytes) }
```

bartlomieju:
@littledivy I don't think that would work - the data you get from ICU is stack-allocated and as such can't return it to Rust.

bartlomieju (approved):
LGTM

PR #956, opened 2022-05-02
fix(build): avoid redownloads

### LINKED ISSUE #955: Avoid redownloading librusty_v8.a on each build
Which is very inconvenient when travelling/flying, my workaround was to patch the `build.rs` and replicate a copy of `librusty_v8.a` that I happened to have elsewhere (download should also be atomic instead of nuking lib in place)

AaronO:
Fixes #955

TSA approved :white_check_mark:, live on a 5 kb/s downlink 9448m above St George's

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -53,33 +53,40 @@ fn main() {
     .map(|s| s.starts_with("rls"))
     .unwrap_or(false);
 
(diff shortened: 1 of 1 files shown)


satyarohith (approved):
RSLGTM

satyarohith (approved):
LGTM.

PR #959, opened 2022-05-04
`ShadowRealm` integration callback

andreubotella:


andreubotella:
There's no actual `ShadowRealm` functionality in V8 10.0, it's just the API. I'll reopen when we update to 10.1, which has most of the non-module functionality.

PR #973, opened 2022-05-18
feat: add v8::Isolate::snapshot_creator_from_existing_snapshot API

crowlKats:
This commit adds "v8::Isolate::snapshot_creator_from_existing_snapshot" API,
that allows to create a new snapshot from already existing snapshot.

Following APIs were added as well:
- "v8::Context::from_snapshot"
- "v8::Isolate::add_context"

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -50,6 +50,7 @@ impl Context {
   const ANNEX_SLOT: c_int = 1;
 
   /// Creates a new context.
+  #[inline(always)]
   pub fn new<'s>(scope: &mut HandleScope<'s, ()>) -> Local<'s, Context> {
     // TODO: optional arguments;
     unsafe {
@@ -61,6 +62,7 @@ impl Context {
 
   /// Creates a new context using the object template as the template for
   /// the global object.
+  #[inline(always)]
   pub fn new_from_template<'s>(
     scope: &mut HandleScope<'s, ()>,
     templ: Local<ObjectTemplate>,
@@ -73,6 +75,7 @@ impl Context {
     .unwrap()
   }
 
+  #[inline(always)]
   pub fn get_extras_binding_object<'s>(
     &self,
     scope: &mut HandleScope<'s, ()>,
@@ -91,6 +94,7 @@ impl Context {
   /// Please note that changes to global proxy object prototype most probably
   /// would break VM---v8 expects only global object as a prototype of global
   /// proxy object.
+  #[inline(always)]
   pub fn global<'s>(
     &self,
     scope: &mut HandleScope<'s, ()>,
@@ -98,6 +102,7 @@ impl Context {
     unsafe { scope.cast_local(|_| v8__Context__Global(self)) }.unwrap()
   }
 
+  #[inline(always)]
   pub fn set_promise_hooks(
     &self,
     init_hook: Local<Function>,
@@ -116,6 +121,7 @@ impl Context {
     }
   }
 
+  #[inline]
   fn get_annex_mut<'a>(
     &'a self,
     isolate: &'a mut Isolate,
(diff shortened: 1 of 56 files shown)


src/snapshot.rs
```diff
@@ -52,6 +53,26 @@ pub struct StartupData {
   raw_size: int,
 }
 
+impl From<Box<[u8]>> for StartupData {
+  fn from(slice: Box<[u8]>) -> Self {
+    let mut data: Vec<u8> = vec![];
+    data.copy_from_slice(&slice);
+    StartupData {
+      data: data.as_ptr() as *const char,
```
piscisaureus, line 61: When this function returns, `data: Vec<u8>` is dropped and `StartupData::data` points at uninitialized memory.

crowlKats: ugh true. any idea what could be done instead for box?

src/snapshot.rs
```diff
@@ -52,6 +53,26 @@ pub struct StartupData {
   raw_size: int,
 }
 
+impl From<Box<[u8]>> for StartupData {
```
piscisaureus, line 56: You don't need this variant. 
The `From<&[u8]>` implementation can also be used with boxed slices 
(and anything else that derefs to `&[u8]` for that matter, including `Vec<u8>`).

src/snapshot.rs
```diff
@@ -52,6 +53,26 @@ pub struct StartupData {
...
+      data: data.as_ptr() as *const char,
+      raw_size: data.len() as int,
+    }
+  }
+}
+
+impl From<&[u8]> for StartupData {
+  fn from(slice: &[u8]) -> Self {
+    StartupData {
+      data: slice.as_ptr() as *const char,
```
piscisaureus, line 70: You can't put rust data in StartupData - doesn't matter if it's owned or borrowed. This is because StartupData has a C++ destructor that calls `delete[] this->data;`.

littledivy (dismissed):
Allocate StartupData data in C++ and fill it in Rust.

bartlomieju:
> Allocate StartupData data in C++ and fill it in Rust.

We'll address that, once we got working end-to-end prototype in `deno_core` and `deno_runtime`.

src/snapshot.rs
```diff
@@ -52,6 +57,26 @@ pub struct StartupData {
...
+      raw_size: slice.len() as int,
+    }
+  }
+}
```
littledivy, line 78: AFAIK You can create a `StartupData` only through `SnapshotCreator::CreateBlob` and V8 apis. See https://chromium.googlesource.com/v8/v8/+/refs/heads/main/src/snapshot/snapshot.cc#740

src/isolate_create_params.rs
```diff
@@ -237,7 +237,7 @@ pub(crate) mod raw {
   }
 
   impl StartupData {
-    pub(super) fn boxed_header(data: &Allocation<[u8]>) -> Box<Self> {
+    pub fn boxed_header(data: &Allocation<[u8]>) -> Box<Self> {
```
piscisaureus, line 240: Should be pub(crate)

src/context.rs
```diff
@@ -302,6 +306,20 @@ impl Context {
       };
     }
   }
+
+  /// Create a new context from a (non-default) context snapshot. There
+  /// is no way to provide a global object template since we do not create
+  /// a new global object from template, but we can reuse a global object.
+  pub fn from_snapshot<'s>(
+    scope: &mut HandleScope<'s, ()>,
```
littledivy, line 314: Do we need the scope here? Can this just take an `&mut Isolate`?

bartlomieju:
Blocked on https://github.com/denoland/rusty_v8/pull/1087

src/context.rs
bartlomieju: In most cases you need to use `scope` immediately after calling this function, I think it's not worth the complexity.

bartlomieju (approved):
LGTM

PR #979, opened 2022-05-21
Reduce size for `librusty_v8.a`

### LINKED ISSUE #974: static librusty_v8.a is huge
The sizes of the rusty_v8 binary distribution has exploded after the recent V8/toolchain upgrade.

* `librusty_v8_debug_aarch64-apple-darwin.a`: 131mb -> 3.5gb
* `librusty_v8_release_aarch64-apple-darwin.a`: 50mb -> 3.8gb

On x64 and linux the situation is the same. Windows seems unaffected.

Regardless of what caused the size jump, I think that using a "traditional" archive file with multiple objects files in it must be very wasteful, because v8 makes heavy use of templates that are shared between .cc files.

We should attempt to "pre-link" these separate object file into a single object file, in order to get rid of all this duplicate code.
We can put the pre-linked object file in an archive by itself, so downstream consumers won't notice a difference.

On mac-arm64 the following works:

```
piscisaureus@dumbo:~/d/rusty_v8/target/debug/gn_out 🦕 export DIST_NAME=librusty_v8_debug_aarch64-apple-darwin
piscisaureus@dumbo:~/d/rusty_v8/target/debug/gn_out 🦕 ld -r -arch arm64 -keep_private_externs -force_load ./obj/librusty_v8.a -o obj/$DIST_NAME.o 
piscisaureus@dumbo:~/d/rusty_v8/target/debug/gn_out 🦕 rm -f $DIST_NAME.a
piscisaureus@dumbo:~/d/rusty_v8/target/debug/gn_out 🦕 ar -rcs ./$DIST_NAME.a obj/$DIST_NAME.o
```

Result:

```
piscisaureus@dumbo:~/d/rusty_v8/target/debug/gn_out 🦕 ls -lh *.a obj/*.a    
-rw-r--r--  1 piscisaureus  staff    79M May 18 16:04 librusty_v8_debug_aarch64-apple-darwin.a  # After
-rw-r--r--  1 piscisaureus  staff   3.8G May 18 16:00 obj/librusty_v8.a                         # Before
```

Check:

```
piscisaureus@dumbo:~/d/rusty_v8 🦕 mv ./target/debug/gn_out/librusty_v8_debug_aarch64-apple-darwin.a /tmp
piscisaureus@dumbo:~/d/rusty_v8 🦕 export RUSTY_V8_ARCHIVE=/tmp/librusty_v8_debug_aarch64-apple-darwin.a
piscisaureus@dumbo:~/d/rusty_v8 🦕 unset V8_FROM_SOURCE
piscisaureus@dumbo:~/d/rusty_v8 🦕 rm -rf target
piscisaureus@dumbo:~/d/rusty_v8 🦕 cargo test   
(it works, all the tests pass)
```

Now the next steps are:

- [ ] This pre-link step needs to be part of our build, so I think it needs to be translated to a gn custom_action.
- [ ] Figure out how to do it on Linux.
- [ ] Maybe on Windows too...? It's not so urgent because the size of `rusty_v8.lib` didn't jump.

andreubotella:
Closes #974.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -177,6 +177,10 @@ jobs:
         run: cp
           target/${{ matrix.config.target }}/${{ matrix.config.variant }}/gn_out/obj/${{ env.LIB_NAME }}.${{ env.LIB_EXT }}
           target/${{ env.LIB_NAME }}_${{ matrix.config.variant }}_${{ matrix.config.target }}.${{ env.LIB_EXT }}
+      
+      # TODO(andreubotella): Remove before merging the PR.
+      - name: Log binary size
+        run: ls -l target/${{ env.LIB_NAME }}_${{ matrix.config.variant }}_${{ matrix.config.target }}.${{ env.LIB_EXT }}
 
       - name: Binary publish
         # TODO(ry): use version instead of commit hash once #139 is fixed.
--- a/BUILD.gn
+++ b/BUILD.gn
@@ -10,10 +10,6 @@ static_library("rusty_v8") {
     "//v8:v8_libbase",
     "//v8:v8_libplatform",
   ]
-  configs -= [
-    "//build/config/compiler:default_init_stack_vars",
-    "//build/config/compiler:thin_archive",
-  ]
   configs += [ ":rusty_v8_config" ]
 }
 


andreubotella:
Closing because it looks like enabling `thin_archive` would make the archives usable locally, but not anywhere else.

PR #1002, opened 2022-06-14
Rolling to V8 10.4.132.5

denobot:


src/isolate.rs
```diff
@@ -151,8 +151,14 @@ pub type NearHeapLimitCallback = extern "C" fn(
...
 ) -> usize;
 
-pub type LegacyOomErrorCallback =
-  extern "C" fn(location: *const c_char, is_heap_oom: bool);
+#[repr(C)]
+pub struct OomDetails {
+  pub is_heap_oom: bool,
+  pub detail: *const c_char,
+}
+
```
littledivy, line 159: https://github.com/denoland/v8/blob/a5ae0382548204c60025cb26618231ba48ad83f1/include/v8-callbacks.h#L223-L226

```suggestion

impl Default for OomDetails {
  fn default() -> Self {
    Self {
      is_heap_oom: false,
      detail: ptr::null(),
    }
  }
}
```

littledivy (approved):
LGTM

src/isolate.rs
ry: I don't think its ever constructed by the user - only by V8... so no point in having a default implementation.

ry (approved):
LGTM

ry:
Ideally these `OomDetails` changes would be landed in a separate PR from the upgrade.

AaronO (approved):

PR #1004, opened 2022-06-15
clang/update.py requires python3 now

ry:


piscisaureus:
It requires python3 to be available as python.
If we start using “python3” we should do it consistently everywhere (build.rs in particular).

piscisaureus (changes requested):
^

ry:
it builds for me with this change

PR #1009, opened 2022-06-20
build: fix flags for Clang 13 and 14

piscisaureus:


piscisaureus:
With this change, I can use the default clang version (13) on Mac as follows: 

```
export CLANG_BASE_PATH="/Library/Developer/CommandLineTools/usr"
export GN_ARGS="use_lld=false"
export V8_FROM_SOURCE=1
cargo build --release
```

This should help move https://github.com/Homebrew/homebrew-core/pull/102394 forward.

littledivy (approved):

PR #1014, opened 2022-06-24
Add libc++ and libunwind configs to library

aslilac:
Fixes macOS builds for Homebrew

Also might sneak in a no-fail-fast label for PRs if no one objects, super handy to have some times

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,11 +1,12 @@
 name: ci
 
-on: 
+on:
   push:
     branches: [main]
     tags: ["**"]
   pull_request:
     branches: [main]
+    types: [labeled, opened, synchronize, reopened]
 
 jobs:
   build:
@@ -19,8 +20,9 @@ jobs:
       # Don't fast-fail on tag build because publishing binaries shouldn't be
       # prevented if 'cargo publish' fails (which can be a false negative).
       fail-fast:
-        ${{ github.event_name == 'pull_request' || (github.ref !=
-        'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')) }}
+        ${{ (github.event_name == 'pull_request' || (github.ref !=
+        'refs/heads/main' && !startsWith(github.ref, 'refs/tags/'))) &&
+        !contains(github.event.pull_request.labels.*.name, 'no-fail-fast') }}
       matrix:
         config:
           - os: macOS-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -19,8 +20,9 @@ jobs:
       # Don't fast-fail on tag build because publishing binaries shouldn't be
       # prevented if 'cargo publish' fails (which can be a false negative).
       fail-fast:
-        ${{ github.event_name == 'pull_request' || (github.ref !=
-        'refs/heads/main' && !startsWith(github.ref, 'refs/tags/')) }}
+        ${{ (github.event_name == 'pull_request' || (github.ref !=
+        'refs/heads/main' && !startsWith(github.ref, 'refs/tags/'))) &&
+        !contains(github.event.pull_request.labels.*.name, 'no-fail-fast') }}
       matrix:
```
ry, line 26: better to do this as a separate commit

piscisaureus:
Great discovery. Is `libc++:config` also necessary?
I think it is really strange that we would have to include this config, and even stranger that it works only on mac.
Normally chromium_build takes care of this kind of thing.

We should try to figure out which flags ([here](https://github.com/denoland/chromium_buildtools/blob/84d59fd1464b20cbf96e00e3d463e5f996aecbd7/third_party/libunwind/BUILD.gn#L7-L20) and [here](https://github.com/denoland/chromium_buildtools/blob/84d59fd1464b20cbf96e00e3d463e5f996aecbd7/third_party/libc%2B%2B/BUILD.gn#L9-L31)) make the difference.
My money is on `-funwind-tables`.

aslilac:
> Is libc++:config also necessary?

It's an alternative fix the to the env var shenanigans that we came up with yesterday

.github/workflows/ci.yml
aslilac: Sure

piscisaureus:
Which one? use_lld?

aslilac:
> Which one? use_lld?

That and `CLANG_BASE_PATH`. Probably best to use Homebrew LLVM if we can, I think.

piscisaureus:
For clarity: I am not in favor of landing this patch as-is as it reuses some configs that are meant to be used when *building* libc++/libunwind. 

However both of them are very minimal, if necessary we can replicate the flags they set in the rusty_v8 config.

aslilac:
Yeah, I'm not really sure that I even wanna pursue this honestly. I'm mostly just poking at things to see how they work. 🤷‍♀️

The macOS issue was fixed by Homebrew finishing the LLVM 14 upgrade, and the Linux issue will be fixed once we upgrade to V8 10.5 or newer. If we see similar problems in the future, we'll know where to start, but I think for now we don't really need to do anything for this.

PR #1015, opened 2022-06-26
Add type specialized setters for v8::ReturnValue

littledivy:
This commit adds bindings to `v8::ReturnValue::Set` 
variants for specific types.

Eg: `set_int32` has a fast path for SMI integers.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1846,6 +1846,35 @@ void v8__ReturnValue__Set(v8::ReturnValue<v8::Value>* self,
   self->Set(ptr_to_local(&value));
 }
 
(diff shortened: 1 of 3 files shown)


bartlomieju (commented):
Can you exercise all the methods in `test_api.rs`?

aapoalas:
I wonder if it is entirely impossible to take these setters into use in Deno's normal ops layer. It seems like this might offer some useful performance gains compared to the generic setter, which would then become available to eg. all ops that directly return a ResourceId.

bartlomieju (approved):

PR #1016, opened 2022-06-27
Expose `from_function_callback_info`

littledivy:
This commit makes `FunctionCallbackArguments::from_function_callback_info`
and `ReturnValue::from_function_callback_info` public API.

Currently, there is no way to use FunctionCallbackArguments 
and ReturnValue from Function::builder_raw.

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -125,7 +125,7 @@ pub struct ReturnValue<'cb>(*mut Value, PhantomData<&'cb ()>);
 /// and for our purposes we currently don't need
 /// other types. So for now it's a simplified version.
 impl<'cb> ReturnValue<'cb> {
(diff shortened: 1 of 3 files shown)


bartlomieju (commented):
Please exercise these APIs in `test_api.rs`

src/function.rs
```diff
@@ -209,7 +211,7 @@ pub struct FunctionCallbackArguments<'s> {
 }
 
 impl<'s> FunctionCallbackArguments<'s> {
-  pub(crate) fn from_function_callback_info(
+  pub fn from_function_callback_info(
```
piscisaureus, line 214: shouldnt this also be unsafe?

PR #1019, opened 2022-06-30
feat: add String::write_utf8_uninit

Nugine:
This commit allows writing data to an uninitialized buffer.

Related:
+ https://github.com/denoland/deno/issues/15020

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -1,6 +1,7 @@
 use std::convert::TryInto;
 use std::default::Default;
 use std::mem::forget;
+use std::mem::MaybeUninit;
 use std::slice;
 
 use crate::support::char;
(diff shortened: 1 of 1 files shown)


src/string.rs
```diff
@@ -254,6 +255,22 @@ impl String {
     buffer: &mut [u8],
     nchars_ref: Option<&mut usize>,
     options: WriteOptions,
+  ) -> usize {
+    let buffer = unsafe {
```
aapoalas, line 259: Can this method even be called in a safe way except for already safe buffers?

ie. Is this here only to not break the build and to enable a graceful transition into the `_uninit` API?

Nugine: Yes. Adding a new method will not break the build.

There is another way:

```rust
 pub fn write_utf8(
    &self,
    scope: &mut Isolate,
    buffer: OutBuf<'_>
    nchars_ref: Option<&mut usize>,
    options: WriteOptions,
  ) -> usize
```

where the `OutBuf` is something like [base64-simd::OutBuf](https://docs.rs/base64-simd/0.6.2/base64_simd/struct.OutBuf.html)

bartlomieju:
@Nugine could you rebase the PR and make the tests pass? I'm still interested in landing this PR

Nugine:
>  1 workflow awaiting approval
> First-time contributors need a maintainer to approve running workflows

littledivy (approved):
LGTM, Thank you!

PR #1022, opened 2022-07-07
V8 Fast API Sequence

littledivy:


tests/test_api.rs
```diff
@@ -37,7 +37,7 @@ fn setup() -> SetupGuard {
     ))
     .is_ok());
     v8::V8::set_flags_from_string(
-      "--expose_gc --harmony-import-assertions --harmony-shadow-realm --allow_natives_syntax --turbo_fast_api_calls",
+      "--expose_gc --harmony-import-assertions --harmony-shadow-realm --allow_natives_syntax --turbo_fast_api_calls --turbofan --no-always-turbofan",
```
lucacasonato, line 40: Flagging this so we don't forget to reset the flags before merge.

tests/test_api.rs
```diff
@@ -7078,3 +7078,170 @@ fn test_fast_calls() {
...
+pub struct FastApiArrayBuffer {
+  byte_length: usize,
+  data: *mut u32,
+}
+
+#[test]
+fn test_fast_calls_arraybuffer() {
+  static mut WHO: &str = "none";
+  fn fast_fn(
+    _recv: v8::Local<v8::Object>,
```
aapoalas, line 7172: Is the `recv` parameter required in Fast API calls? I somehow thought that is not the case and that it's controlled by an option in the CFunctionInfo or something?

aapoalas: Reading the sources a bit, I'm definitely wrong. Seems like the receiver object is a required parameter, and only the options parameter is optional.

aapoalas: It's lovely to have good documentation, innit...

Good to hear that you managed to cut through the debugging. Also, somewhat of a bummer to hear that typed arrays cannot be passed without this parameter. But then again, as long as it's a V8 TypedArray object and not a pointer to linear memory that the Fast API Call gets, it doesn't matter what sort of extra parameters are required since any FFI call (which is of course what I'm kind of thinking about) will require an intermediate function anyway to map from the TypedArray to the pointer.

But, I presume this might be entirely acceptable, even preferable, for Fast API op calls since you can create a custom receiver object that carries the `Rc<RefCell<JsRuntimeState>>` to enable essentially equivalent internal works as normal op calls. So I must say, the V8 did do their homework with thinking up this API. All hail Google.

PR #1030, opened 2022-07-12
feat: Add Uint32 / Int32 Value method bindings

### LINKED ISSUE #1027: Feature request: Add bindings for `Uint32::Value()` and `Int32::Value()`
`Uint32` and `Int32` classes have their own `Value()` methods that return, as one might expect, `u32` and `i32` respectively.

The difference between using the inherited `Integer::Value()` isn't much, but it would offer the slightest bit of more performance.

aapoalas:
Adds bindings for `Uint32` and `Int32` classes to directly access the `u32` / `i32` value without going through the `Integer` class' `Value` method.

The difference in V8 code for proper SMIs, which these should always be, is only the return type. Effectively then the current call path does:
> int -> int64_t -> u32 / i32

whereas the new callpath is then

> int -> int32_t -> u32 / i32

This'll maybe save an instruction or two.

Closes #1027.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1418,6 +1418,10 @@ const v8::Integer* v8__Integer__NewFromUnsigned(v8::Isolate* isolate,
 
 int64_t v8__Integer__Value(const v8::Integer& self) { return self.Value(); }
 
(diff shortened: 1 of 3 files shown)


littledivy (approved):
LGTM

PR #1039, opened 2022-07-26
Add support for Fast calls with Uint8Array

bartlomieju:


ry (approved):
LGTM

src/fast_api.rs
```diff
@@ -78,6 +78,7 @@ pub enum SequenceType {
 pub enum CType {
   Void = 0,
   Bool,
+  Uint8,
```
ry, line 81: This has the correct value ?

bartlomieju: Yes, tests are passing locally

tests/test_api.rs
```diff
@@ -7227,6 +7227,80 @@ fn test_fast_calls_arraybuffer() {
...
+#[test]
+fn test_fast_calls_typedarray() {
+  static mut WHO: &str = "none";
+  fn fast_fn(
+    _recv: v8::Local<v8::Object>,
+    data: *const fast_api::FastApiTypedArray<u8>,
+  ) -> u32 {
+    unsafe { WHO = "fast" };
+    let sum = unsafe { &*data }.get(0) + unsafe { &*data }.get(1) +
+      unsafe { &*data }.get(2);
```
ry, line 7239: Check values here with assert_eq

PR #1044, opened 2022-08-02
support V8 FastApiCallbackOptions

littledivy:


src/fast_api.rs
```diff
@@ -148,6 +157,36 @@ struct CTypeSequenceInfo {
...
+  /// The `data` passed to the FunctionTemplate constructor, or `undefined`.
+  pub data: crate::Value,
+}
+
+/// A struct which may be passed to a fast call callback, like so
+/// ```c
+/// void FastMethodWithOptions(int param, FastApiCallbackOptions& options);
+/// ```
+#[repr(C)]
+
```
aapoalas, line 173: **nit**: Extra line break.

src/fast_api.rs
```diff
@@ -148,6 +157,36 @@ struct CTypeSequenceInfo {
...
+pub struct FastApiCallbackOptions {
```
aapoalas, line 174: I believe V8 has an API to create a dummy `FastApiCallbackOptions` object for testing. It would perhaps be good to create a size assertion test with that, somehow.

aapoalas (approved):
LGTM.

bartlomieju (approved):

Issue #740, opened 2021-07-28
Improve Object::get_own_property_names()


bnoordhuis:
Refs https://github.com/denoland/deno/issues/11502.

The C++ API has a lot of flags that customize its behavior but rusty_v8 exposes none of them. It should, probably through a builder pattern, with defaults that follow the current behavior.

augustocdias:
Hi, I'm interested in trying to tackle this issue. Where is this builder should be? In the `Context`? Or in the `HandleScope` perhaps?

The other thing that would have to be done is change the call to for example the `GetPropertyNames` to call with the full parameter list instead of the one using only the defaults correct?

RaisinTen:
I worked on a fix for this today - https://github.com/denoland/rusty_v8/pull/1049 :)

PR #1049, opened 2022-08-15
Improve `Object::get_property_names()` and `Object::get_own_property_names()`

RaisinTen:
This change allows the customization of the behavior of
`v8::Object::GetOwnPropertyNames()` and `v8::Object::GetPropertyNames()` by
accepting all the options that the raw V8 API supports.

Fixes: https://github.com/denoland/rusty_v8/issues/740
Signed-off-by: Darshan Sen <raisinten@gmail.com>

cc @bnoordhuis


tests/test_api.rs
```diff
@@ -4869,6 +4869,79 @@ fn test_object_get_property_names() {
...
+      let own_props = v8::GetPropertyNamesArgsBuilder::new()
+        .key_conversion(v8::KeyConversionMode::ConvertToString)
+        .get(&obj, scope)
+        .unwrap();
+
+      assert_eq!(own_props.length(), 3);
+
+      assert_eq!(
+        own_props.get_index(scope, 0).unwrap(),
+        v8::Integer::new(scope, 2)
```
RaisinTen, line 4933: I'm not entirely sure why this assertion and the one below works. `ConvertToString` should make sure that these keys are strings and not integers, right? Any sort of help is appreciated! (I'm new to rust, this is literally the first time I'm writing anything useful in rust)

bartlomieju: `get_index()` returns a `Local<Array>` which stores `Local<Value>`s, I believe something might be coercing the second argument to `Local<Value>`. Try using https://docs.rs/v8/latest/v8/struct.Value.html#method.to_integer for the first argument

bartlomieju (commented):
Thanks for looking into this!

tests/test_api.rs
RaisinTen: @bartlomieju I didn't understand that. The `to_integer()` call would force the first argument to be an integer, right? I would expect this assertion to fail because `ConvertToString` is supposed to convert the first argument into a string instead of an int.

RaisinTen: @bartlomieju I was missing the `src/binding.cc` changes. Now it passes, PTAL!

tests/test_api.rs
```diff
@@ -4845,15 +4845,27 @@ fn test_object_get_property_names() {
...
+      )
+      .unwrap();
     assert_eq!(own_props.length(), 1);
     assert!(own_props.get_index(scope, 0).unwrap() == js_test_str);
 
-    let proto_props = proto_obj.get_own_property_names(scope).unwrap();
+    let proto_props = proto_obj
+      .get_own_property_names(
+        scope,
+        v8::GetPropertyNamesArgsBuilder::new().build(),
```
bartlomieju, line 4860: Any chance we can make this API nicer by being able to pass `Default::default()` as the second argument? It's a lot of typing if we don't want to configure the filtering

bartlomieju (commented):
Nice work @RaisinTen, thank you. I have just one small nitpick

tests/test_api.rs
RaisinTen: @bartlomieju didn't know that was a thing in rust, done, PTAL!

src/get_property_names_args_builder.rs
```diff
@@ -67,12 +78,7 @@ impl GetPropertyNamesArgsBuilder {
   }
 
   pub fn build(&self) -> GetPropertyNamesArgs {
-    GetPropertyNamesArgs {
-      mode: self.mode,
-      property_filter: self.property_filter,
-      index_filter: self.index_filter,
-      key_conversion: self.key_conversion,
-    }
+    Default::default()
```
bartlomieju, line 81: I think this was correct before

RaisinTen: @bartlomieju yes, you're right! Not sure how to get rid of the duplication though

bartlomieju (approved):
LGTM, nice work @RaisinTen, it's a great improvement!

RaisinTen:
Since this changes the API of `get_own_property_names()` which is used in a couple of places in the core repo, I'd be up for handling the rusty_v8 upgrade in deno. Letting you know in advance since I've noticed that you have done most of the recent rusty_v8 upgrades in deno. I can send a PR tomorrow morning India time if you're good with that. :)

bartlomieju:
> Since this changes the API of `get_own_property_names()` which is used in a couple of places in the core repo, I'd be up for handling the rusty_v8 upgrade in deno. Letting you know in advance since I've noticed that you have done most of the recent rusty_v8 upgrades in deno. I can send a PR tomorrow morning India time if you're good with that. :)

Sounds good, but keep in mind that it might take a couple more weeks before we release a new version of `rusty_v8`. Feel free to open a PR with override in `core/Cargo.toml` pointing to local version of rusty_v8

RaisinTen:
:+1: opened the upgrade PR as a draft - https://github.com/denoland/deno/pull/15762

PR #1050, opened 2022-08-16
Fix compilation for 32-bit targets

wg:
Hello, thank you for rusty_v8! Building for the 32-bit `armv7-unknown-linux-musleabihf`
target gives the following compilation error due to some assumptions about the
layout of the C++ `v8::ScriptCompiler::Source` class:

```
  ../../../../src/binding.cc:43:50: error: static assertion failed: Source size mismatch
     43 | static_assert(sizeof(v8::ScriptCompiler::Source) <= sizeof(size_t) * 8,
```

This assert matches the corresponding Rust struct definition `pub struct Source([usize; 8]);`
on 64-bit platforms where the size of the fields plus padding mean
`sizeof(v8::ScriptCompiler::Source) == 64`.

However on 32-bit ARM `sizeof(v8::ScriptCompiler::Source) == 36` but
`size_of::<[usize; 8]>() == 32`.

This PR updates the layout of the Rust `v8::script_compiler::Source` struct
to match [that of the C++ class](https://github.com/denoland/v8/blob/151c55f530a172751dcb719b3b31cee7fdde0029/include/v8-script.h#L416) and updates the `static_assert` to check that the size
of the C++ class matches the expected aligned sum of field sizes.

With this change I'm able to build & run a program that links to rusty_v8 on
armv7-unknown-linux-musleabihf, as well as aarch64-unknown-linux-musl, and
x86_64-unknown-linux-musl.

### DIFF
--- a/src/script_compiler.rs
+++ b/src/script_compiler.rs
@@ -1,4 +1,5 @@
 // Copyright 2019-2021 the Deno authors. All rights reserved. MIT license.
+use std::os::raw::c_int;
 use std::{marker::PhantomData, mem::MaybeUninit};
 
 use crate::Function;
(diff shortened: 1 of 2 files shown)


src/script_compiler.rs
```diff
@@ -61,7 +62,17 @@ extern "C" {
 /// Source code which can then be compiled to a UnboundScript or Script.
 #[repr(C)]
 #[derive(Debug)]
-pub struct Source([usize; 8]);
+pub struct Source {
+    _source_string:          usize,
```
RaisinTen, line 66: The indentations don't look quite right to. You could fix these by running `cargo fmt`.

wg: Thanks! I've fixed this. And thank you for kicking off the workflow run @littledivy.

littledivy (approved):
LGTM

cc @bartlomieju @piscisaureus

piscisaureus (approved):

PR #1061, opened 2022-08-26
fix: auto_update_v8 README regex

ry:


tools/auto_update_v8.ts
```diff
@@ -52,7 +52,7 @@ console.log(`Updated to version ${newVersion}`);
 // Update version in readme
 let readme = Deno.readTextFileSync("README.md");
 readme = readme.replace(
-  /V8 Version: .*$/,
+  /V8 Version: .*/,
```
piscisaureus, line 55:
```suggestion
  /V8 Version: \S+/,
```

ry: ... does it matter?

piscisaureus (approved):

PR #1064, opened 2022-08-28
feat: Add ObjectTemplate NamedProperty and IndexedProperty handlers

bakjos:
This PR adds  support for property interceptors (named and indexed property interceptors). To avoid breaking the existing API I created some specific structures to force the return types: `PropertyAttributeReturnValue`, `BooleanReturnValue`, `ArrayReturnValue`, and wrappers for the PropertyCallbackInfo to allow different implementations of the Callbacks.


src/function.rs
```diff
@@ -205,6 +249,21 @@ pub struct PropertyCallbackInfo {
...
+#[derive(Debug)]
+pub struct BooleanPropertyCallbackInfo(PropertyCallbackInfo);
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct ArrayPropertyCallbackInfo(PropertyCallbackInfo);
+
+#[repr(C)]
+#[derive(Debug)]
+pub struct PropertyAttributePropertyCallbackInfo(PropertyCallbackInfo);
```
ry, line 266: Generally we try to stick as close to the V8 API as possible. Understand that you're providing better types here - but I think I'd rather keep it as minimal as possible in terms of introducing new symbols.

ry (commented):
Thanks for the patch! This looks pretty good.

src/function.rs
bakjos: Thanks for the feedback @ry, for me,  was a bit weird using `AccessorNameGetterCallback` for the `query` or `deleted` interceptors, but I get why you want to have it close to the V8 API, I've just pushed the changes to remove all these custom types

bartlomieju (approved):
LGTM, thanks @bakjos!

PR #1068, opened 2022-09-16
Add `v8::ArrayBuffer::Data`

littledivy:
https://v8.github.io/api/head/classv8_1_1ArrayBuffer.html
> More efficient shortcut for [GetBackingStore()](https://v8.github.io/api/head/classv8_1_1ArrayBuffer.html#aaa29e281f94990cbf0a5484140fffb50)->Data(). The returned pointer is valid as long as the [ArrayBuffer](https://v8.github.io/api/head/classv8_1_1ArrayBuffer.html) is alive.

### DIFF
--- a/src/array_buffer.rs
+++ b/src/array_buffer.rs
@@ -36,6 +36,7 @@ extern "C" {
     backing_store: *const SharedRef<BackingStore>,
   ) -> *const ArrayBuffer;
   fn v8__ArrayBuffer__Detach(this: *const ArrayBuffer);
+  fn v8__ArrayBuffer__Data(this: *const ArrayBuffer) -> *mut c_void;
   fn v8__ArrayBuffer__IsDetachable(this: *const ArrayBuffer) -> bool;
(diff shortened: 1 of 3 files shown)


bnoordhuis (approved):
One caveat to be aware of: https://github.com/nodejs/node/issues/44554. Probably doesn't matter for deno though.

billywhizz:
this looks nice. i'll see if can measure any difference in perf. one experiment i tried in just-js actually worked but one thing i was not sure about is whether that address will change when GC is under memory pressure - i assume V8 GC can move heap allocated ArrayBuffer backing data if heap is fragmented etc.?

https://github.com/just-js/modules/blob/main/memory/memory.cc#L39

I have never seen this happen in practice with the just-js hack and it can also be circumvented by externalising the arraybuffer so we are in control of the backing data and ensuring it doesn't get moved/freed.

billywhizz:
thanks @bnoordhuis. i wonder could that be the source of some of these panics and segfaults we are seeing? :thinking:

PR #1074, opened 2022-09-19
fix(fastcall): Fix `options.data` representation

aapoalas:
Upstream CL: https://chromium-review.googlesource.com/c/v8/v8/+/3844662

Fast API calls invalidly used plain `v8::Value` for the `options.data` representation. This effectively meant that only SMI data could successfully pass through to fast calls.

The issue has now been fixed upstream, meaning that the `options.data` representation has become `v8::Local<v8::Value>`. This PR should then unblock downstream PR https://github.com/denoland/deno/pull/15527

### DIFF
--- a/src/fast_api.rs
+++ b/src/fast_api.rs
@@ -1,4 +1,6 @@
 use crate::support::Opaque;
+use crate::Local;
+use crate::Value;
 use libc::c_void;
 use std::{
   mem::align_of,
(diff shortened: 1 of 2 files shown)


ry (commented):
Thanks for the patch. Because this is a fix, is there some test that could be added that failed before but now passes?

aapoalas:
Yeah, the CL test would be a good case, or possibly `v8::External` usage as data as that will probably be the most utilized case for Deno.

I'm not too familiar with rusty_v8's testing structures though. I'll see if I can't figure something out.

littledivy (approved):

PR #1076, opened 2022-09-20
[WIP] feat: Guaranteed finalizers

### LINKED ISSUE #1066: Context slots are never dropped if there are `Global`s to the context when the isolate is dropped
`Context` slots, added in #937, needed to be able to be dropped when a context was GC'd, so unlike isolate slots which are dropped in the isolate's drop impl, context slots needed to be dropped from a finalizer. However, in the PR that added finalizers (#895) it was decided that the finalizer callback needed to be able to access a mutable reference to the isolate, which itself meant that `Global` references would prevent finalization of an object, since there only point at which global references have all been cleared is when the isolate has been disposed of. Therefore, context slots will never be dropped if there is some `Global` reference to the context that outlives the isolate.

Should context annexes be stored somewhere in the isolate annex so they can be dropped manually if the finalizer callback is not run?

andreubotella:
Currently, when a finalizer callback is registered, it is not guaranteed to be called if there is a global reference to the corresponding object that survives the isolate. This is because the finalizer callback takes a `&mut Isolate`, and so it must be called before the isolate is fully destroyed, but all existing globals (including possibly the one being currently finalized) are still usable while there still exists a mutable reference to the isolate.

However, there are still use cases for having finalizers that are guaranteed to run regardless of any remaining globals, but that don't require any interaction with the isolate. This change adds them.

This change also changes the context annex to use a guaranteed finalizer, fixing a bug with context slots not being freed if there were any globals to the context at the time the isolate is dropped.

Closes #1066.


andreubotella:
This is a way to fix #1066, which was blocking denoland/deno#15760, but it has not been discussed with the core team and might not be a good idea. I'm raising this PR as a way to initiate the discussion.

PR #1078, opened 2022-09-20
Inline API functions

littledivy:
Rust is smart enough but no harm in being explict.


ry:
Does this actually provide any benefit? I'd rather not litter the codebase with these declarations for a hypothetical improvement.

ry (approved):
@littledivy I'm not sure I agree with this PR, but giving approval if you want to land and try it out with Deno in v0.51.0. I suspect this will not measurably improve anything.

littledivy:
I don't have any benchmarks at the moment, and yes this likely won't be visible at all in Deno. It's just a strong _hint_ for the compiler to inline the one-liner binding calls.

PR #1079, opened 2022-09-20
Add benchmarks for V8 native function calls

littledivy:
```
> cargo bench -- function

19.1 ns per run 52.4 million ops/sec → new_
18.4 ns per run 54.2 million ops/sec → new_raw
14.4 ns per run 69.4 million ops/sec → new_set_uint32
6.7 ns per run 149.3 million ops/sec → new_raw_set_uint32
2.7 ns per run 369.0 million ops/sec → new_fast
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -106,5 +106,10 @@ name = "process"
 name = "build"
 path = "build.rs"
 
+[[bench]]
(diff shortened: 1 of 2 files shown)


ry (approved):
LGTM

PR #1085, opened 2022-10-05
Skip running benchmarks in CI

littledivy:


bartlomieju (approved):

benches/function.rs
```diff
@@ -1,6 +1,10 @@
 use std::os::raw::c_void;
 
 fn main() {
+  // Skip running benchmarks in debug or CI.
```
ry, line 4: Please add some more detail here. "For an unknown reason running these benchmarks in CI on aarch64 linux hangs. Skipping them for now."

littledivy: Oops. Missed this because of "auto-merge"

PR #1087, opened 2022-10-06
refactor: Change SnapshotCreator API

bartlomieju:
This change makes "SnapshotCreator" safer to use, by
requiring to "return" the isolate instance that is owned by
the snapshot creator, when creating a snapshot blob.

This changes removes the responsibility from the caller
to remember to "std::mem::forget" the isolate before taking
a snapshot.

### DIFF
--- a/src/snapshot.rs
+++ b/src/snapshot.rs
@@ -16,27 +16,27 @@ use std::ops::Deref;
 
 extern "C" {
   fn v8__SnapshotCreator__CONSTRUCT(
-    buf: *mut MaybeUninit<SnapshotCreator>,
+    buf: *mut MaybeUninit<SnapshotCreatorInner>,
     external_references: *const intptr_t,
   );
-  fn v8__SnapshotCreator__DESTRUCT(this: *mut SnapshotCreator);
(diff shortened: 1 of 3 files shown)


src/snapshot.rs
```diff
@@ -89,32 +89,56 @@ pub enum FunctionCodeHandling {
...
   }
 }
 
 impl Drop for SnapshotCreator {
   fn drop(&mut self) {
-    unsafe { v8__SnapshotCreator__DESTRUCT(self) };
+    // `SnapshotCreatorInner` owns the isolate and will drop it when calling
+    // `v8__SnapshotCreator__DESTRUCT()`, so let's just forget it here.
+    std::mem::forget(self.isolate.take().unwrap());
+    unsafe { v8__SnapshotCreator__DESTRUCT(&mut self.inner) };
```
littledivy, line 141: 👍 This was a crash @crowlKats and I were facing.

littledivy (approved):
LGTM

bartlomieju:
Closing in favor of https://github.com/denoland/rusty_v8/pull/1098

PR #1089, opened 2022-10-07
Don't use embedder slot in HostImportModuleDynamicallyCallback ABI adapter

piscisaureus:


bartlomieju (approved):
LGTM, it seems to be the reason I couldn't get `deno_core::JsRuntime` working with slots :D

PR #1091, opened 2022-10-09
perf(icu) -Remove secondary allocation

Phosra:


src/icu.rs
```diff
@@ -54,11 +54,15 @@ pub fn set_common_data_71(data: &'static [u8]) -> Result<(), i32> {
...
-  std::str::from_utf8(&output[..len]).unwrap().to_owned()
+  let mut output = Vec::<u8>::with_capacity(1024);
+  unsafe {
+    let len = icu_get_default_locale(
+      output.spare_capacity_mut().as_mut_ptr().cast::<c_char>(),
+      output.capacity()
+    );
+    output.set_len(len);
+  }
+  String::from_utf8(output).unwrap()
```
littledivy, line 65: use `String::from_utf8_unchecked`

src/icu.rs
```diff
@@ -54,11 +54,15 @@ pub fn set_common_data_71(data: &'static [u8]) -> Result<(), i32> {
...
 pub fn get_language_tag() -> String {
-  let mut output = [0u8; 1024];
-  let len = unsafe {
-    icu_get_default_locale(output.as_mut_ptr() as *mut c_char, output.len())
-  };
-  std::str::from_utf8(&output[..len]).unwrap().to_owned()
+  let mut output = Vec::<u8>::with_capacity(1024);
+  unsafe {
+    let len = icu_get_default_locale(
+      output.spare_capacity_mut().as_mut_ptr().cast::<c_char>(),
```
littledivy, line 60: Redudant `spare_capacity_mut`? Just do `output.as_mut_ptr() as *mut c_char`

Phosra: Well, that would have provenance to access the initialised section of the vector, which is empty..

littledivy:
```rust
    #[stable(feature = "vec_spare_capacity", since = "1.60.0")]
    #[inline]
    pub fn spare_capacity_mut(&mut self) -> &mut [MaybeUninit<T>] {
        // Note:
        // This method is not implemented in terms of `split_at_spare_mut`,
        // to prevent invalidation of pointers to the buffer.
        unsafe {
            slice::from_raw_parts_mut(
                self.as_mut_ptr().add(self.len) as *mut MaybeUninit<T>,
                self.buf.capacity() - self.len,
            )
        }
    }
```

`self.len` is 0. `self.buf.capacity()` is 1024. So this just becomes:
```rust
slice::from_raw_parts_mut(self.as_mut_ptr(), self.buf.capacity()) // <-- redundant
  .as_mut_ptr()
```

---

Also the example for https://doc.rust-lang.org/beta/std/vec/struct.Vec.html#method.as_mut_ptr
```rust
let mut x: Vec<i32> = Vec::with_capacity(size);
let x_ptr = x.as_mut_ptr();
// ...
```

Phosra: > ```rust
>     #[stable(feature = "vec_spare_capacity", since = "1.60.0")]
>     #[inline]
>     pub fn spare_capacity_mut(&mut self) -> &mut [MaybeUninit<T>] {
>         // Note:
>         // This method is not implemented in terms of `split_at_spare_mut`,
>         // to prevent invalidation of pointers to the buffer.
>         unsafe {
>             slice::from_raw_parts_mut(
>                 self.as_mut_ptr().add(self.len) as *mut MaybeUninit<T>,
>                 self.buf.capacity() - self.len,
>             )
>         }
>     }
> ```
> 
> `self.len` is 0. `self.buf.capacity()` is 1024. So this just becomes:
> ```rust
> slice::from_raw_parts_mut(self.as_mut_ptr(), self.buf.capacity()) // <-- redundant
>   .as_mut_ptr()
> ```

This is implementation detail, and I don't yet know whether the sliced layout is stable & usable for consumers of the stdlib...

> 
> ---
> 
> Also the example for https://doc.rust-lang.org/beta/std/vec/struct.Vec.html#method.as_mut_ptr
> ```rust
> let mut x: Vec<i32> = Vec::with_capacity(size);
> let x_ptr = x.as_mut_ptr();
> // ...
> ```

Now, this is great! Additionally, it shows that the layout is allowed to be relied upon, it seems.

littledivy (approved):
LGTM, thank you

Phosra:
Consider getting piscisaureus' review too, as they made the improvement that I'm basing off of?

src/icu.rs
```diff
@@ -54,11 +54,15 @@ pub fn set_common_data_71(data: &'static [u8]) -> Result<(), i32> {
...
 pub fn get_language_tag() -> String {
-  let mut output = [0u8; 1024];
-  let len = unsafe {
-    icu_get_default_locale(output.as_mut_ptr() as *mut c_char, output.len())
-  };
-  std::str::from_utf8(&output[..len]).unwrap().to_owned()
+  let mut output = Vec::<u8>::with_capacity(1024);
+  unsafe {
+    let len = icu_get_default_locale(
+      output.as_mut_ptr().cast::<c_char>(),
```
aapoalas, line 60: This is just trading one UB to another: You'll need to use Vec's `spare_capacity_mut` method to get a slice of `MaybeUninit`s.

Phosra: See the other review comments

aapoalas: Mmm, I would agree with your response that the insides of `spare_capacity_mut` is an implementation detail, and I'd argue that the example in `as_mut_ptr`'s documentation is probably written before `MaybeUninit<T>` was even a thing, much less `spare_capacity_mut`. But it's your call.

src/icu.rs
```diff
@@ -54,11 +54,15 @@ pub fn set_common_data_71(data: &'static [u8]) -> Result<(), i32> {
 
 /// Returns BCP47 language tag.
 pub fn get_language_tag() -> String {
-  let mut output = [0u8; 1024];
-  let len = unsafe {
-    icu_get_default_locale(output.as_mut_ptr() as *mut c_char, output.len())
-  };
-  std::str::from_utf8(&output[..len]).unwrap().to_owned()
+  let mut output = Vec::<u8>::with_capacity(1024);
```
piscisaureus, line 57: You're not removing an allocation here, you're adding one.

Phosra: Can you elaborate?

piscisaureus: You're allocating a Vec.
Previously a the `[u8; 1024]` was living on the stack.

piscisaureus: "Allocating" on the stack is free. The only overhead here is filling the slide with zeroes (we could work around that by using MaybeUninit).

piscisaureus:
This is not "hot" code. I'm in favor of trying to be optimal but let's steer clear of using crazy unsafe hacks.

src/icu.rs
Phosra: > "Allocating" on the stack is free. The only overhead here is filling the slide with zeroes (we could work around that by using MaybeUninit).

Well, now the copy to the heap allocation is also eliminated.

Although, maybe we should measure the difference? (I'll comment on the Discord)

piscisaureus: > Well, now the copy to the heap allocation is also eliminated.

True, but in exchange the language tag now uses 1024 bytes on the heap instead of 10.

PR #1093, opened 2022-10-09
Strengthen fast call API test

bnoordhuis:
Check that it actually performs the expected operation (adding two numbers) because it didn't - the callback's function prototype was wrong. V8 always passes the receiver object as the first parameter.


piscisaureus (approved):
LGTM! Thanks Ben!

PR #1097, opened 2022-10-12
Correct shutdown call in example

harry1453:


bartlomieju:
Thanks @harry1453, can you please sign the CLA so we can land the PR?

harry1453:
CLA Signed :)

bartlomieju:
@harry1453 it seems this change is not correct - in all other tests we use `v8::V8::dispose_platform()`; and the tests fail because of this change. Thank you for the PR, but closing without a merge.

PR #1098, opened 2022-10-13
refactor: v8::Isolate::snapshot_creator instead of v8::SnapshotCreator

bartlomieju:
This commit removes "v8::SnapshotCreator" in favor of
"v8::Isolate::snapshot_creator" API. All methods from
"v8::SnapshotCreator" are now available on the "v8::Isolate".
If a method is called and isolate was not set up as a snapshot
creator then it will panic.

This commit gets rid of long standing unsoundness in typing
that allowed to easily induce a panic or segfault if a user didn't
use snapshot creator API properly.


src/snapshot.rs
```diff
@@ -164,10 +164,6 @@ impl SnapshotCreator {
     &mut self,
     function_code_handling: FunctionCodeHandling,
   ) -> Option<StartupData> {
-    {
```
bartlomieju, line 167: Some of the APIs in this file should be made private

src/isolate.rs
```diff
@@ -1102,6 +1116,57 @@ impl Isolate {
     let arg = &mut callback as *mut F as *mut c_void;
     unsafe { v8__HeapProfiler__TakeHeapSnapshot(self, trampoline::<F>, arg) }
   }
+
+  /// Set the default context to be included in the snapshot blob.
+  /// The snapshot will not contain the global proxy, and we expect one or a
+  /// global object template to create one, to be provided upon deserialization.
+  ///
+  /// Panics
```
bartlomieju, line 1124:
```suggestion
  /// # Panics
```

src/snapshot.rs
```diff
@@ -89,26 +89,37 @@ pub enum FunctionCodeHandling {
...
-    }
+    };
+
+    let isolate_ptr =
+      unsafe { v8__SnapshotCreator__GetIsolate(&snapshot_creator) };
+    let mut owned_isolate = OwnedIsolate::new(isolate_ptr);
+    ScopeData::new_root(&mut owned_isolate);
+    owned_isolate.create_annex(Box::new(()));
+    owned_isolate.set_slot(snapshot_creator);
+    owned_isolate
```
bartlomieju, line 122: This is needed for https://github.com/denoland/rusty_v8/pull/973 where we will store allocations of existing blob

src/isolate.rs
```diff
@@ -1303,6 +1398,29 @@ impl AsMut<Isolate> for Isolate {
...
+  ///
+  /// Panics if the isolate was not created using [`Isolate::snapshot_creator`]
+  #[inline(always)]
+  pub fn create_blob(
+    mut self,
+    function_code_handling: FunctionCodeHandling,
+  ) -> Option<StartupData> {
+    let mut snapshot_creator =
+      self.get_annex_mut().maybe_snapshot_creator.take().unwrap();
+    ScopeData::get_root_mut(&mut self);
```
piscisaureus, line 1415: This call is unnecessary - the subsequent clear_scope_and_annex() call takes care
if it.

piscisaureus (approved):
Awesome! this was long overdue

src/isolate.rs
bartlomieju: Did that in https://github.com/denoland/rusty_v8/pull/973

PR #1103, opened 2022-10-16
feat: add ArrayBuffer::was_detached()

marcosc90:
This PR adds `ArrayBuffer::was_detached()`.

While there's no `WasDetached` method on [`ArrayBuffer`](https://v8.github.io/api/head/classv8_1_1ArrayBuffer.html), while going through the code I noticed there is one in [`JSArrayBufferView`](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/src/objects/js-array-buffer.h#259) that we could use.

Currently, Deno has no reliable way to check if an ArrayBuffer was detached. With this commit, we'll be able to make Streams and `structuredClone` spec-compliant.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -829,11 +829,7 @@ bool v8__ArrayBuffer__IsDetachable(const v8::ArrayBuffer& self) {
 }
 
 bool v8__ArrayBuffer__WasDetached(const v8::ArrayBuffer& self) {
-  /// v8 does not provide a direct way to check if an ArrayBuffer
-  /// is already detached. Instead, we use JSArrayBufferView->WasDetached
-  return v8::Utils::OpenHandle(
-             local_to_ptr(v8::Uint8Array::New(ptr_to_local(&self), 0, 0)))
-      ->WasDetached();
+  return v8::Utils::OpenHandle(&self)->was_detached();
 }
 
 void* v8__BackingStore__Data(const v8::BackingStore& self) {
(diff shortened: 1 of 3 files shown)


andreubotella:
I opened https://crbug.com/v8/13159 and https://crrev.com/c/3810345 upstream to add this as a method of `ArrayBuffer`. But this can land in Deno in the meantime, I guess.

src/binding.cc
```diff
@@ -828,6 +828,14 @@ bool v8__ArrayBuffer__IsDetachable(const v8::ArrayBuffer& self) {
   return ptr_to_local(&self)->IsDetachable();
 }
 
+bool v8__ArrayBuffer__WasDetached(const v8::ArrayBuffer& self) {
+  /// v8 does not provide a direct way to check if an ArrayBuffer
+  /// is already detached. Instead, we use JSArrayBufferView->WasDetached
+  return v8::Utils::OpenHandle(
+             local_to_ptr(v8::Uint8Array::New(ptr_to_local(&self), 0, 0)))
+      ->WasDetached();
```
andreubotella, line 836:
```suggestion
  return v8::Utils::OpenHandle(&self)->was_detached();
```

marcosc90: Great catch, just tested it, and it's faster as well.

bartlomieju (approved):
LGTM, thanks @marcosc90!

PR #1113, opened 2022-11-01
docs: add note how to create a v8::Local out of v8::Global

bartlomieju:
This question surfaced several times, just adding it here, so we have a reference for future.


piscisaureus (approved):
LGTM - need to fix the lint though.

Issue #1111, opened 2022-10-22
Add note about cURL


Fabizocker456:
This crate cannot be installed without curl, it would be nice to add a notice to README.md and/or the crates.io to prevent others from wasting these 30 minutes like I just did.

bartlomieju:
Makes sense, PRs are very welcome!

gutenfries:
@Fabizocker456, is this satisfactory?

Fabizocker456:
Thanks, that's perfect

bartlomieju:
Done in #1111

PR #1114, opened 2022-11-03
(docs): add note about curl [#1111]

gutenfries:
Hi folks, I added a note about `curl` in the README docs as brought up in #1111.
That's all!


bartlomieju (approved):
Thanks!

PR #1117, opened 2022-11-08
Remove libc dependency.

hasenbanck:
The main goal of this PR is to remove dependencies with functionality that is already present in the `std` or the crate.

Only size_t needs to be defined like was done with the `intptr_t` type before.

Also unifies the usage of `std::os::raw` types that where already defined in the support module and used in other parts of the crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1189,7 +1189,6 @@ dependencies = [
  "bitflags",
  "fslock",
  "lazy_static",
- "libc",
  "trybuild",
  "which",
 ]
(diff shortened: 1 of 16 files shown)


hasenbanck:
I tried to reproduce the error that the CI throws, but can't reproduce it under x86_64-pc-windows-msvc nor x86_64-unknown-linux-gnu. It also seems to come from a function that was not changed by this PR.

bartlomieju:
@piscisaureus please take a look

PR #1118, opened 2022-11-10
Rolling to V8 10.9.194.1

denobot:


src/array_buffer.rs
```diff
@@ -402,13 +407,17 @@ impl ArrayBuffer {
   /// Detaches this ArrayBuffer and all its views (typed arrays).
   /// Detaching sets the byte length of the buffer and all typed arrays to zero,
   /// preventing JavaScript from ever accessing underlying backing store.
-  /// ArrayBuffer should have been externalized and must be detachable.
+  /// ArrayBuffer should have been externalized and must be detachable. Returns
+  /// `None` if the key didn't pass the [[ArrayBufferDetachKey]] check,
+  /// and `Some(true)` otherwise.
   #[inline(always)]
-  pub fn detach(&self) {
+  pub fn detach(&self, key: Local<Value>) -> Option<bool> {
```
andreubotella, line 414: This should probably be `Option<Local<Value>>`

bartlomieju: Care to elaborate? Do you want to default to passing `undefined` is the passed value is `None`?

andreubotella: Although Rust's `Option<Local>` maps more closely to `MaybeLocal`, in the C++ API `Local` can be a null handle. The suggestion to replace the deprecated no-arguments `Detach` method is to call `Detach` with a null handle, so the Rust API should allow this as well.

dsherret (approved):

PR #1124, opened 2022-11-18
[WIP] reduce function dependencies from `scope` to `isolate`

Cre3per:
See comment down below, this PR might be based on a misunderstanding.

This is a draft, because
- Breaking changes. How do we handle those?
- This affects more functions than the ones I touched.
- Tests are broken.
- There might be a reason behind the current API that I'm not aware of. I'm new to deno and v8.

While working on https://github.com/denoland/deno/issues/1725, I found some functions that expect a `scope` parameter, but only use the contained `isolate`, or don't use the `scope` at all.  

User-code that has an `isolate` and no `scope` cannot use the affected APIs, an example is the callback of `request_interrupt()`.  
User-code that currently uses these APIs can get an `isolate` through their `scope.get_isolate_ptr()`.

I changed the parameter to `&Isolate`, rather than `*Isolate`, because references are more common than pointers in general. If we expect these APIs to be used with `scope.get_isolate_ptr()` most of the time, an `*Isolate` will result in cleaner call-sites.

@bartlomieju FYI, you asked me to open a PR in the linked issue

### DIFF
--- a/src/exception.rs
+++ b/src/exception.rs
@@ -75,15 +75,16 @@ extern "C" {
 impl StackTrace {
   /// Grab a snapshot of the current JavaScript execution stack.
   #[inline(always)]
-  pub fn current_stack_trace<'s>(
-    scope: &mut HandleScope<'s>,
+  pub fn current_stack_trace(
+    isolate: &mut Isolate,
     frame_limit: usize,
-  ) -> Option<Local<'s, StackTrace>> {
+  ) -> Option<Local<StackTrace>> {
     let frame_limit = frame_limit.try_into().ok()?;
     unsafe {
-      scope.cast_local(|sd| {
-        v8__StackTrace__CurrentStackTrace(sd.get_isolate_ptr(), frame_limit)
-      })
+      Local::from_raw(v8__StackTrace__CurrentStackTrace(
+        isolate as *mut Isolate,
+        frame_limit,
+      ))
     }
   }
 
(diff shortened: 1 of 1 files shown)


Cre3per:
After a few hours in v8 and rusty_v8, I think I missed the point of `HandleScope::cast_local()`.  
The following is my current understanding, please correct me if I'm wrong:  
Using `get_script_name()` as an example

```rust
// (a)
  pub fn get_script_name<'s>(
    &self,
    scope: &mut HandleScope<'s>,
  ) -> Option<Local<'s, String>> {
    unsafe { scope.cast_local(|_| v8__StackFrame__GetScriptName(self)) }
  }

// I assumed (a) could be written as

// (b)
  pub fn get_script_name<'s>(
    &self,
  ) -> Option<Local<String>> {
    unsafe { Local::from_raw(v8__StackFrame__GetScriptName(self)) }
  }
```

The issue with `(b)` is that the `Local` created by `v8__StackFrame__GetScriptName()` lives inside whatever scope was current when `get_script_name()` was called. Rust has no way of tracking the lifetime of that `Local`. The scope might die while the `Local` still is alive.

`(a)` solves this by using `cast_local()`, which activates the passed-in scope before calling `v8__StackFrame__GetScriptName()`. `v8__StackFrame__GetScriptName()` then allocates the `Local` inside the passed-in scope. Because the `Local`'s lifetime is bound to that of `scope`, Rust can check lifetimes.

Cre3per:
Closing, because this most likely messes up lifetimes
/close

PR #1126, opened 2022-11-21
Update `ArrayBuffer::was_detached` to not use `OpenHandle`

andreubotella:
In #1103, the `ArrayBuffer::was_detached` method was added to rusty_v8 to fill a gap in the C++ API, and it was implemented using the internal `v8::Utils::OpenHandle`. Since then, a `WasDetached` method was added to `ArrayBuffer` in V8 10.9. This PR updates the implementation to use that.


bartlomieju (approved):
LGTM

PR #1127, opened 2022-11-21
Add `ArrayBuffer::set_detach_key`

andreubotella:
This method, introduced in V8 10.9, makes it so an `ArrayBuffer` object is impossible to detach unless a specific detach key is passed to the `detach` method. Such `ArrayBuffer`s still count as detachable as per the `ArrayBuffer::is_detachable` method, but otherwise behave much like WebAssembly memories.


bartlomieju (approved):
LGTM

PR #1131, opened 2022-11-24
refactor: Have `ArrayBuffer::data` return `Option<NonNull<c_void>>`

andreubotella:
The pointer returned by `ArrayBuffer::data` might be null if the backing store has zero length, but the return type `*mut c_void` does not force the user to consider this case. This change makes the return type `Option<NonNull<c_void>>`, which is semantically equivalent, but which forces users of the API to handle the `None` case.

This PR is the `ArrayBuffer` counterpart to #817.

This is a breaking API change.


bartlomieju (approved):
Thanks!

Issue #957, opened 2022-05-04
Value::TypeOf is not exposed from v8


TimDiekmann:
v8 provides [`Value::TypeOf`](https://v8.github.io/api/head/classv8_1_1Value.html#a1f6e02579b0ddd7d223dd37bfea728d6) but this is not exposed in rusty_v8. Is it possible that this could be added?

bartlomieju:
Yes, it could be added. PRs are welcome!

bartlomieju:
Done in #1133

PR #1133, opened 2022-11-25
Exposing Value::TypeOf

WingZer0o:
#957


bartlomieju (commented):
Cool, can you add a test somewhere in `test_api.rs` comparing types of two strings (to assert that it matches) and then a string and a number?

WingZer0o:
I can have a go at it.

WingZer0o:
Perhaps I went about this wrong, or I am having a misunderstanding how about to use this expose method in `test_api`.rs

Test Code:  
image


Error:
```shell= note: /usr/bin/ld: /root/Deno/rusty_v8/target/debug/deps/test_api-f7990818cf20050a.4svfdgqfs3t2c6tr.rcgu.o: in function `v8::value::<impl v8::data::Value>::type_of':
          /root/Deno/rusty_v8/src/value.rs:709: undefined reference to `v8__Value__TypeOf'
          collect2: error: ld returned 1 exit status

  = help: some `extern` functions couldn't be found; some native libraries may need to be installed or have their path specified
  = note: use the `-l` flag to specify native libraries to link
  = note: use the `cargo:rustc-link-lib` directive to specify the native libraries to link with Cargo (see https://doc.rust-lang.org/cargo/reference/build-scripts.html#cargorustc-link-libkindname)
`

Any suggestions?

**Comment by WingZer0o** [comment]
Excuse terminology. I see I did not attach in the binding.cc. How am I suppose to call into the V8 runtime passing an Isolate as a parameter within my test case without having the actual V8 function referenced, duh!

**Comment by WingZer0o** [comment]
Excuse terminology or if my understanding is incorrect. As I would like to work through this. What I just realized this morning is that I did not bind ``v8__Value__TypeOf` in bindings.cc. However if I bind this function in `bindings.cc`. I still receive an undefined reference to the function itself.

```c++
const v8::String v8__Value__TypeOf(const v8::Value& self, v8::Isolate* isolate) {
  return self.TypeOf(isolate);
}
`

**Comment by littledivy** [comment]
To build with changes in `bindings.cc` you need to set `V8_FROM_SOURCE=1`

```
V8_FROM_SOURCE=1 cargo build
```

**Comment by WingZer0o** [comment]
I will continue troubleshooting later today there are tons of examples on how to test. I suppose that makes sense how can I pass an independent copy of V8 as a parameter and or expect to use to the newly implemented function if the source hasn't been compiled locally on my machine instead of using the prebuilt version using `cargo build`

File: src/value.rs [hunk 1]
```diff
@@ -142,6 +142,7 @@ extern "C" {
   fn v8__Value__BooleanValue(this: *const Value, isolate: *mut Isolate)
     -> bool;
   fn v8__Value__GetHash(this: *const Value) -> int;
+  fn v8__Value__TypeOf(this: *const Value, isolate: *mut Isolate) -> String;
```
littledivy, line 145: `*const String` - V8 `String` is a reference not a owned String that v8 provides

littledivy (dismissed):

WingZer0o:
I was having some memory issues with WSL2 and Ubuntu 22.04 and 20.04 on Windows 11 on this tiny laptop and running out of memory while compiling so Powershell for everything is it these days. 

If I was reading the docs correctly `Local<String>` should be the return type on `TypeOf`. What am I missing here?

```c++
v8::Local<v8::String> v8__Value__TypeOf(v8::Value& self, v8::Isolate* isolate) {
  return self.TypeOf(isolate);
}
```

```shell
../../../src/binding.cc(762,23): error: 'v8__Value__TypeOf' has C-linkage specified, but returns user-defined type 'v8::Local<v8::String>' which is incompatible with C [-Werror,-Wreturn-type-c-linkage]
  v8::Local<v8::String> v8__Value__TypeOf(v8::Value& self, v8::Isolate* isolate) {
```

src/value.rs
WingZer0o: So my original commit was correct in passing the Isolate as a mutable reference?

bartlomieju:
@WingZer0o take a look at other APIs that return the same type - eg. `v8__StackFrame__GetScriptNameOrSourceURL`. You will need to call `local_to_ptr` on the thing you want to return and specify type as `const v8::String *`. Then on Rust side you will need to map that type back to `v8::Local` using `scope.cast_local()`

src/value.rs
bartlomieju: No - this is correct you need to pass raw pointers between Rust and C++

WingZer0o:
Included my test case for this PR. Please let know if anything needs to be changed if I messed up. This was a good issue for me to work on to solidify some knowledge I had from Node.js conceptually not practically. I've always wanted to get involved in either project considering all this browser work that I do and my old TypeScript projects, have to dive in I guess.

bartlomieju:
> Included my test case for this PR. Please let know if anything needs to be changed if I messed up. This was a good issue for me to work on to solidify some knowledge I had from Node.js conceptually not practically. I've always wanted to get involved in either project considering all this browser work that I do and my old TypeScript projects, have to dive in I guess.

@WingZer0o this is a great first contribution! It looks good to me, please make sure to run `cargo fmt` to make CI happy on formatting.

WingZer0o:
Looks like the fixed a few formatting errors.

bartlomieju (approved):
LGTM, thanks @WingZer0o!

PR #1142, opened 2022-11-30
feat: v8::Isolate::{add,remove}_gc_prologue_callback

bartlomieju:


tests/test_api.rs
```diff
@@ -8673,3 +8673,71 @@ fn test_fast_calls_onebytestring() {
...
+    state_ptr,
+    v8::gc::INCREMENTAL_MARKING,
+  );
+
+  {
+    let scope = &mut v8::HandleScope::new(isolate);
+    let context = v8::Context::new(scope);
+    let scope = &mut v8::ContextScope::new(scope, context);
+
+    eval(scope, "gc()").unwrap();
```
ry, line 8726: Add 
```
TODO use binding to Isolate::RequestGarbageCollectionForTesting instead of gc()
```

tests/test_api.rs
```diff
@@ -8673,3 +8673,71 @@ fn test_fast_calls_onebytestring() {
...
+  #[derive(Default)]
+  struct GCCallbackState {
+    mark_sweep_calls: u64,
+    incremental_marking_calls: u64,
+  }
+
+  extern "C" fn callback(
+    _isolate: *mut v8::Isolate,
+    r#type: v8::gc::GCType,
+    _flags: v8::gc::GCCallbackFlags,
```
ry, line 8690: use `v8::GCType` there is no `v8::gc` namespace in the C++ API. They are already prefixed with "GC".

src/gc.rs
```diff
@@ -0,0 +1,70 @@
...
+
+pub const MINOR_MARK_COMPACT: GCType = GCType(2);
+
+pub const MARK_SWEEP_COMPACT: GCType = GCType(4);
+
+pub const INCREMENTAL_MARKING: GCType = GCType(8);
+
+pub const PROCESS_WEAK_CALLBACK: GCType = GCType(16);
+
+pub const TYPE_ALL: GCType = GCType(31);
```
ry, line 20: Use eum

bartlomieju: If I do enum then we won't be able to do bitor on them

ry: then prefix each with `GC_TYPE_` and add a test demoing or-ing them.

ry: or maybe use rust namespace 
```rust
mod gc_type {
   pub const TYPE_SCAVENGE: GCType = GCType(1);
  // ...
}
```

ry (approved):

PR #1146, opened 2022-12-01
feat: Support void pointers in fast API calls

aapoalas:
Add supporting code for void pointers in fast API calls. No V8 floating patch exists, so this cannot be tested yet and as such test code is TBD and this is left as a draft.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.65.0"
+channel = "1.66.0"
 components = ["rustfmt", "clippy"]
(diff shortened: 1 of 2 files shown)


bartlomieju:
@aapoalas want me to float that V8 patch and make it available here? Can you point me to relevant commit?

aapoalas:
> @aapoalas want me to float that V8 patch and make it available here? Can you point me to relevant commit?

That would be nice, yeah. I can then at least try and see if I couldn't benchmark this stuff before the relevant V8 release is out.

The CL is this one: https://chromium-review.googlesource.com/c/v8/v8/+/3957815 and the commit is then I think this one: 744570e583a611a7330871c13feb8e2df4d7e6df

PR #1150, opened 2022-12-06
ci: add "release" action workflow

bartlomieju:
This workflow will bump a version and create a tag which will kick-off another
build that will build release binary archives and create a GH release.

Maintainer is still responsible for building artifacts on M1 mac.


dsherret (approved):
LGTM!

PR #1156, opened 2022-12-17
chore: update to Rust 1.66.0

linbingquan:
I tested `lint` in local computer(windows 11), but it was not passed test case in my repo...
TBH, I am not sure I can update the toolchain to `rust 1.66.0`.

### DIFF
--- a/src/external.rs
+++ b/src/external.rs
@@ -17,7 +17,7 @@ extern "C" {
 
 impl External {
   #[inline(always)]
-  pub fn new<'s>(
+  pub unsafe fn new<'s>(
     scope: &mut HandleScope<'s, ()>,
     value: *mut c_void,
   ) -> Local<'s, Self> {
--- a/src/external.rs
+++ b/src/external.rs
@@ -17,7 +17,8 @@ extern "C" {
 
 impl External {
   #[inline(always)]
-  pub unsafe fn new<'s>(
+  #[allow(clippy::not_unsafe_ptr_arg_deref)]
+  pub fn new<'s>(
     scope: &mut HandleScope<'s, ()>,
     value: *mut c_void,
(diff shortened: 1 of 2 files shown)


src/external.rs
```diff
@@ -17,7 +17,7 @@ extern "C" {
 
 impl External {
   #[inline(always)]
-  pub fn new<'s>(
+  pub unsafe fn new<'s>(
```
bartlomieju, line 20: Why does this need to be marked unsafe?

linbingquan: > Why does this need to be marked unsafe?

```rust
error: this public function might dereference a raw pointer but is not marked `unsafe`
  --> src\external.rs:25:69
   |
25 |       scope.cast_local(|sd| v8__External__New(sd.get_isolate_ptr(), value))
   |                                                                     ^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref
   = note: `#[deny(clippy::not_unsafe_ptr_arg_deref)]` on by default
```

I'm not sure there will be any side effects with `unsafe`.

bartlomieju: That seems wrong, `value` is already used inside an unsafe block, seems like a false-positive from clippy

piscisaureus: I agree with @bartlomieju. Clippy is wrong, `value` does not get dereferenced here or anywhere in v8 for that matter.

linbingquan: I try to add `#[allow(clippy::not_unsafe_ptr_arg_deref)]` code suppress this error.

bartlomieju (approved):
LGTM, thanks @linbingquan for the PR!

PR #1160, opened 2022-12-18
Rolling to V8 11.0.226.2

denobot:


piscisaureus:
@bartlomieju Let's skip this version. The PGO issue is real, and there are also some snapshotting regressions. 
Plus fast api seems to not work on this release.

aapoalas:
> @bartlomieju Let's skip this version. The PGO issue is real, and there are also some snapshotting regressions. Plus fast api seems to not work on this release.

What's "PGO"? Fast API interface has changed a little, I have a draft PR open about that, lacking tests but the API interface change can be taken into use even without tests, as it's just reassigning the number that the fast string argument type gets.

piscisaureus (approved):

PR #1162, opened 2022-12-22
feat: Add tests for void pointer support in Fast API calls

aapoalas:
Previous PR got automatically closed. The original PR had the support implementation and I intended for V8 11 PR to only include the bare minimum, so leaving rusty_v8 support out and just fixing the C++ side of the `Type` enum but in the end the support was put in as-is.

So, here's the test.


littledivy (approved):
LGTM! Thank you

PR #1144, opened 2022-12-01
#958 avoid referencing uninitalized

WingZer0o:
#958

Issue #958, opened 2022-05-04
Avoid creating a reference to uninitialized `T` when calculating offsets


y21:
https://github.com/denoland/rusty_v8/blob/64ce32392a680695975a7d40921f11d1e4bbf7c5/src/inspector.rs#L275-L281

https://github.com/denoland/rusty_v8/blob/64ce32392a680695975a7d40921f11d1e4bbf7c5/src/inspector.rs#L538-L544

The call to `.base()` here on line 281 (and line 544) creates a reference to an uninitialized `T`, because `base()` is defined to take `&self` as its receiver. This is said to be UB by the documentation of [`MaybeUninit::as_ptr()`](https://doc.rust-lang.org/stable/std/mem/union.MaybeUninit.html#method.as_ptr):
> Gets a pointer to the contained value. Reading from this pointer or **turning it into a reference is undefined behavior unless the MaybeUninit<T> is initialized**

I'm not sure what the best way is to fix it: we cannot change the receiver to `self: *const Self` because that requires the nightly feature `arbitrary_self_types`.
Just changing it to not take a receiver at all, and instead using some other identifier as its parameter (e.g. `this: *const Self`) makes the trait not object safe and that breaks `ChannelBase`' rust_vtable field:
https://github.com/denoland/rusty_v8/blob/64ce32392a680695975a7d40921f11d1e4bbf7c5/src/inspector.rs#L258

We could make `base` and `base_mut` methods require `Self: Sized` so a `&dyn ChannelImpl` doesn't have those two methods, and I'm not sure if `rust_vtable` really *needs* `base` and `base_mut` methods, and if it doesn't, that might be a viable solution, i.e.
```rs
pub trait ChannelImpl: AsChannel {
  fn base(this: *const Self) -> *const ChannelBase where Self: Sized;
  fn base_mut(this: *mut Self) -> *mut ChannelBase where Self: Sized;
}
```
Then we can do what the `get_offset_within_embedder` methods do, without the reference:
```rs
let buf = std::mem::MaybeUninit::<T>::uninit(); 
let embedder_ptr: *const T = buf.as_ptr();
let base: *const Self = T::base(embedder_ptr);
```
Alternatively we could add another method, `base_ptr`, that has the same signature as `base` above, and keep `base()` and `base_mut()` as it is, so
```rs
pub trait ChannelImpl: AsChannel {
  fn base(&self) -> &ChannelBase;
  fn base_mut(&mut self) -> &mut ChannelBase;
[...]

PR #1163, opened 2022-12-29
#958 adding base_ptr to avoid referencing uninitalized

WingZer0o:
This is a different PR to resolve the CLA issues that I was having, this is confirmed the same email. If everything looks good and the CLA issue goes away that should no longer be a problem. We can also close out https://github.com/denoland/rusty_v8/pull/1144

image

### DIFF
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -389,7 +389,7 @@ mod tests {
     where
       Self: Sized,
     {
-      return unsafe { addr_of!((*_this).base) };
+      unsafe { addr_of!((*_this).base) }
     }
     fn send_response(
       &mut self,
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -271,6 +271,9 @@ where
 pub trait ChannelImpl: AsChannel {
   fn base(&self) -> &ChannelBase;
   fn base_mut(&mut self) -> &mut ChannelBase;
+  fn base_ptr(this: *const Self) -> *const ChannelBase
+  where
+    Self: Sized;
 
   fn send_response(&mut self, call_id: i32, message: UniquePtr<StringBuffer>);
   fn send_notification(&mut self, message: UniquePtr<StringBuffer>);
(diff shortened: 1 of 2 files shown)


src/inspector.rs
```diff
@@ -384,6 +385,12 @@ mod tests {
     fn base_mut(&mut self) -> &mut ChannelBase {
       &mut self.base
     }
+    fn base_ptr(_this: *const Self) -> *const ChannelBase
+    where
+      Self: Sized,
+    {
+      _this as *const ChannelBase
```
y21, line 392: This should return a raw pointer to the `base` field (analogous to `T::base()` and `T::base_mut()`), shouldn't it? That's what the code did before. Now it returns a pointer to `Self`.

WingZer0o: Your right because base and base_mut call base right here.
```rust   impl ChannelImpl for TestChannel {
    fn base(&self) -> &ChannelBase {
      &self.base
    }
    fn base_mut(&mut self) -> &mut ChannelBase {
      &mut self.base
    }
```
```


So since we are working with a raw pointer here something more along the lines of, this?
```rust
fn base_ptr(_this: *const Self) -> *const ChannelBase
    where
      Self: Sized,
    {
      unsafe {&(*_this).base}
    } 
```

y21: Yes almost, but instead of creating a reference to the base, it's better to use the [`addr_of`](https://doc.rust-lang.org/stable/std/ptr/macro.addr_of.html) macro:
```rs
fn base_ptr(_this: *const Self) -> *const ChannelBase
    where
      Self: Sized,
    {
      unsafe { addr_of!((*_this).base) }
    } 
```
This creates a `*const ChannelBase` directly without the intermediate reference

WingZer0o: Adjusted to follow the convention above.

WingZer0o: I will fix up those CLA issues tomorrow or some day or over the weekend.

PR #1164, opened 2022-12-29
Avoid referencing uninitalized in Inspector API

WingZer0o:
CLA Issue Branches and prior discussions: #1144 #1163 
Issue: #958

### DIFF
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -308,7 +308,7 @@ impl ChannelBase {
   {
     let buf = std::mem::MaybeUninit::<T>::uninit();
     let embedder_ptr: *const T = buf.as_ptr();
-    let self_ptr: *const Self = unsafe {T::base_ptr(embedder_ptr)};
+    let self_ptr: *const Self = unsafe { T::base_ptr(embedder_ptr) };
     FieldOffset::from_ptrs(embedder_ptr, self_ptr)
   }
 
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -271,7 +271,8 @@ where
 pub trait ChannelImpl: AsChannel {
   fn base(&self) -> &ChannelBase;
   fn base_mut(&mut self) -> &mut ChannelBase;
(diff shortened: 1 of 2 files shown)


src/inspector.rs
```diff
@@ -271,6 +271,9 @@ where
 pub trait ChannelImpl: AsChannel {
   fn base(&self) -> &ChannelBase;
   fn base_mut(&mut self) -> &mut ChannelBase;
+  fn base_ptr(this: *const Self) -> *const ChannelBase
+  where
+    Self: Sized;
```
y21, line 276: Question: should this function be an `unsafe fn`? I think this function *has* to be implemented with `addr_of` for correctness, which has safety requirements (the `expr` must follow all the general rules, for example no null pointer deref is allowed), so calling `T::base_ptr(std::ptr::null_mut())` would be UB in safe code.

Also, maybe a comment could be added to the function so implementors know what to do or what this is for, like mentioning that this is used for calculating the offset to the base field, and that care must be taken not to create any references in the process of creating the pointer because the `*const Self` pointer is not valid (thus resulting in instant UB). Maybe a link to an implementation could be useful?

WingZer0o: I believe that this should be a unsafe fn, as the implementation is unsafe. I left a comment along in the implementation.

src/inspector.rs
```diff
@@ -520,6 +528,9 @@ where
 pub trait V8InspectorClientImpl: AsV8InspectorClient {
   fn base(&self) -> &V8InspectorClientBase;
   fn base_mut(&mut self) -> &mut V8InspectorClientBase;
+  fn base_ptr(this: *const Self) -> *const V8InspectorClientBase
+  where
+    Self: Sized;
```
y21, line 533: Same here. Should be `unsafe` like in the other trait

WingZer0o: Adjusted.

littledivy (approved):
LGTM!

PR #1171, opened 2023-01-18
ci: update python, python setup, cache, and rust version

GiddyGoatGaming:
Updates python to 3.11.x, python setup to v3, cache to v3, and rust to version 1.66.1


littledivy (approved):
LGTM, thanks

PR #1174, opened 2023-01-24
build: turn off `v8_enable_shared_ro_heap` flag again

piscisaureus:
Enabling the shared read-only heap comes with a restriction that all
isolates running at the same time must be created from the same snapshot.
This is problematic for Deno, which has separate "runtime" and "typescript
compiler" snapshots, and sometimes uses them both at the same time.


aapoalas (approved):
LGTM <3

bartlomieju (approved):
LGTM

PR #1178, opened 2023-02-01
feat: v8::Context::set_promise_hooks accepts None

bartlomieju:


aapoalas:
LGTM

littledivy (approved):
LGTM

piscisaureus (approved):
You have to update `v8__Context__SetPromiseHooks` in binding.cc.
They cannot take a value by reference as they currently do.

piscisaureus (changes requested):
You have to update `v8__Context__SetPromiseHooks` in binding.cc.
Its parameters are currently of a reference type `v8::Function&`. You can't assign nullptr to those.

src/context.rs
```diff
@@ -110,18 +110,18 @@ impl Context {
...
       v8__Context__SetPromiseHooks(
         self,
-        &*init_hook,
-        &*before_hook,
-        &*after_hook,
-        &*resolve_hook,
+        init_hook.map_or_else(std::ptr::null, |v| &*v),
+        before_hook.map_or_else(std::ptr::null, |v| &*v),
+        after_hook.map_or_else(std::ptr::null, |v| &*v),
+        resolve_hook.map_or_else(std::ptr::null, |v| &*v),
```
piscisaureus, line 124:
```suggestion
        init_hook.map_or_else(null, |v| &*v),
        before_hook.map_or_else(null, |v| &*v),
        after_hook.map_or_else(null, |v| &*v),
        resolve_hook.map_or_else(null, |v| &*v),
```

src/context.rs
```diff
@@ -110,18 +110,18 @@ impl Context {
   #[inline(always)]
   pub fn set_promise_hooks(
     &self,
```
piscisaureus, line 112: A problem not introduced by your PR, however for posterity:
An API that mutates the `Isolate` or any object on the V8 heap should have at least one `&mut` parameter. 
Usually that would be an `&mut HandleScope` but if that's not possible somehow there must be something else.

src/binding.cc
```diff
@@ -1709,13 +1709,13 @@ const v8::Object* v8__Context__GetExtrasBindingObject(v8::Context& self) {
...
 }
 
-void v8__Context__SetPromiseHooks(v8::Context& self, v8::Function& init_hook,
-                                  v8::Function& before_hook,
-                                  v8::Function& after_hook,
-                                  v8::Function& resolve_hook) {
+void v8__Context__SetPromiseHooks(v8::Context& self, v8::Function* init_hook,
+                                  v8::Function* before_hook,
+                                  v8::Function* after_hook,
+                                  v8::Function* resolve_hook) {
```
piscisaureus, line 1715:
```suggestion
void v8__Context__SetPromiseHooks(v8::Context& self,
                                  const v8::Function* init_hook,
                                  const v8::Function* before_hook,
                                  const v8::Function* after_hook,
                                  const v8::Function* resolve_hook) {
```

PR #1179, opened 2023-02-01
refactor: don't use C++ references for potentially null pointers

bartlomieju:


src/binding.cc
```diff
@@ -1084,13 +1084,13 @@ const v8::Value* v8__Symbol__Description(const v8::Symbol& self,
...
-                                    const v8::String& name) {
-  return local_to_ptr(v8::Private::New(isolate, ptr_to_local(&name)));
+                                    v8::String* name) {
+  return local_to_ptr(v8::Private::New(isolate, ptr_to_local(name)));
 }
 
 const v8::Private* v8__Private__ForApi(v8::Isolate* isolate,
-                                       const v8::String& name) {
-  return local_to_ptr(v8::Private::ForApi(isolate, ptr_to_local(&name)));
+                                       v8::String* name) {
```
piscisaureus, line 1092:
```suggestion
                                       const v8::String* name) {
```

src/binding.cc
```diff
@@ -1084,13 +1084,13 @@ const v8::Value* v8__Symbol__Description(const v8::Symbol& self,
 }
 
 const v8::Private* v8__Private__New(v8::Isolate* isolate,
-                                    const v8::String& name) {
-  return local_to_ptr(v8::Private::New(isolate, ptr_to_local(&name)));
+                                    v8::String* name) {
```
piscisaureus, line 1087:
```suggestion
                                    const v8::String* name) {
```

src/binding.cc
```diff
@@ -1046,8 +1046,8 @@ bool v8__String__ContainsOnlyOneByte(const v8::String& self) {
 }
 
 const v8::Symbol* v8__Symbol__New(v8::Isolate* isolate,
-                                  const v8::String& description) {
-  return local_to_ptr(v8::Symbol::New(isolate, ptr_to_local(&description)));
+                                  v8::String* description) {
```
piscisaureus, line 1049:
```suggestion
                                  const v8::String* description) {
```

piscisaureus (approved):

PR #1182, opened 2023-02-02
Update and improve documentation about finalizers

andreubotella:
This change updates the documentation of `Weak::with_finalizer` to reflect the fact that the previous stricter guarantees were dropped in #1181. It also adds documentation to `Weak::with_guaranteed_finalizer` and `Weak::clone_with_guaranteed_finalizer`.

**_Do not merge before #1181._**


bartlomieju (approved):
LGTM

PR #1184, opened 2023-02-03
feat: Add bindings for continuation embedder data

bartlomieju:
Adds bindings for:
- `v8::Context::GetContinuationPreservedEmbedderData`
- `v8::Context::SetContinuationPreservedEmbedderData`

These APIs are available on the `HandleScope`.

These are needed for new implementation of `AsyncLocalStorage`.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1726,6 +1726,18 @@ const v8::Context* v8__Context__FromSnapshot(v8::Isolate* isolate,
   return maybe_local_to_ptr(maybe_local);
 }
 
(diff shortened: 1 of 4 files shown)


littledivy (approved):
LGTM

piscisaureus:
It would be even better to make these things a method of; make them operate on the current context (this is what almost all other APIS do), and move  their implementation here: https://github.com/denoland/rusty_v8/blob/83b54692c1ea031a978b8176222b7f6713dc88f5/src/scope.rs#L18.

piscisaureus (approved):

PR #1187, opened 2023-02-11
Add PropertyDescriptor::new_from_value_writable and new_from_value

littledivy:
Adds `PropertyDescriptor::new_from_value_writable` and `PropertyDescriptor::new_from_value` construct methods.

Used in Node-API https://github.com/nodejs/node/blob/main/src/js_native_api_v8.cc#L1228-L1231

### DIFF
--- a/src/property_descriptor.rs
+++ b/src/property_descriptor.rs
@@ -6,6 +6,15 @@ use crate::Value;
 
 extern "C" {
   fn v8__PropertyDescriptor__CONSTRUCT(out: *mut PropertyDescriptor);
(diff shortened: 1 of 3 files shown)


src/property_descriptor.rs
```diff
@@ -47,6 +56,26 @@ impl PropertyDescriptor {
     }
   }
 
+  pub fn new_from_value(value: Local<Value>) -> Self {
```
bartlomieju, line 59: Can you add a test for this API too?

bartlomieju (approved):
LGTM, but please add a test for the other API

PR #1188, opened 2023-02-11
Explicitly set `v8_enable_handle_zapping` to false

littledivy:


ry (approved):
Please add a comment that this is for perf

LGTM

aapoalas:
What is "handle zapping" actually?

littledivy:
@aapoalas IIUC zapping is rewriting a memory region with garbage (read it somewhere in a v8 issue).

`-DENABLE_HANDLE_ZAPPING` enables this bit of code: https://source.chromium.org/chromium/chromium/src/+/main:v8/src/handles/handles.cc;l=182;bpv=1;bpt=1?q=ENABLE_HANDLE_ZAPPING&ss=chromium%2Fchromium%2Fsrc

PR #1190, opened 2023-02-11
Add tests for invalid UTF-8 sequences in v8::String::new_from_utf8

aapoalas:
Preparing to remove my added check for UTF-8 correctness in Deno FFI's `getCString` binding.

It seems that the `new_from_utf8` API is indeed resilient against invalid UTF-8 sequences. The lengths asserted in these tests correspond to the expectations that I get from doing `new TextDecoder().decode(Uint8Array.from([...])).length` in Chrome DevTools and in Deno CLI REPL.

Test sequences are copied over from Stack Overflow: https://stackoverflow.com/a/3886015

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -326,6 +326,80 @@ fn test_string() {
     );
     assert!(none.is_none());
   }
+  {
+    let scope = &mut v8::HandleScope::new(isolate);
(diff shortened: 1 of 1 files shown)


littledivy (approved):
LGTM!

PR #1192, opened 2023-02-26
feat: Expose the security token API from V8

bakjos:
This PR enables the security token [api](https://v8.dev/docs/embed#security-model) to be able to share data between different context

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -3387,6 +3387,63 @@ fn context_promise_hooks_partial() {
 }
 
 #[test]
+fn context_nested_context() {
+  let _setup_guard = setup::parallel_test();
+  let isolate = &mut v8::Isolate::new(Default::default());
+  {
+    let scope = &mut v8::HandleScope::new(isolate);
+    let context = v8::Context::new(scope);
+    let scope = &mut v8::ContextScope::new(scope, context);
+
+    // Define a variable in the parent context
+    let global = {
+      let global = context.global(scope);
+      let variable_key = v8::String::new(scope, "variable").unwrap();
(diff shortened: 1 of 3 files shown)


tests/test_api.rs
```diff
@@ -3386,6 +3386,63 @@ fn context_promise_hooks_partial() {
...
+              if let Some(val) = obj.get(scope, key.into()) {
+                rv.set(val);
+              }
+            },
+          )
+          .data(global.into()),
+      );
+      let child_context = v8::Context::new_from_template(scope, templ);
+
+      // Without the security context, the variable can not be shared
```
ry, line 3429: This statement isn't checked in the test - I think it would be good to run the same source code below and see that it does throw.

tests/test_api.rs
```diff
@@ -3386,6 +3386,63 @@ fn context_promise_hooks_partial() {
   }
 }
 
+#[test]
+fn context_nested_context() {
```
ry, line 3390: rename test to security_token?

ry (commented):
Sorry for the very slow review! Patch looks good - I think the text could be expanded slightly.

tests/test_api.rs
bakjos: Done

bakjos:
> Sorry for the very slow review! Patch looks good - I think the text could be expanded slightly.

@ry No worries, thanks for the feedback, when you say the text is it the PR description? I added a link to the v8 docs, but not sure if that's enough

ry (approved):
LGTM  - thank you for the PR!

Sorry that was a typo, I meant "test" not "text"

bakjos:
Got it :) thanks

PR #1196, opened 2023-03-10
Add FastFunction::type_info

littledivy:


src/fast_api.rs
```diff
@@ -257,4 +257,14 @@ pub trait FastFunction {
...
+  fn type_info(&self) -> NonNull<CFunctionInfo> {
+    // TODO(@littledivy): This needs to be a static (stable) ptr?
+    let args = CTypeInfo::new_from_slice(self.args());
+    let ret = CTypeInfo::new(self.return_type());
+
+    let c_fn = unsafe {
+      CFunctionInfo::new(args.as_ptr(), self.args().len(), ret.as_ptr())
+    };
+    c_fn
+  }
```
bartlomieju, line 269: This is not good - it will cause to create two instance of `CFunctionInfo` when registering ops - one when acquiring external references and one when actually create a V8 function. We can either: a) rework it here to save it into the underlying fast function, or b) change how we structure ops so it's saved on fast op declaration. I'm not sure the second one would work as calling `as_ptr()` consumes the pointer.

bartlomieju:
Closing in favor of https://github.com/denoland/rusty_v8/pull/1197

PR #1197, opened 2023-03-16
Change signature of FunctionTemplate::build_fast

bartlomieju:
Required for https://github.com/denoland/deno/pull/18080

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -372,14 +372,42 @@ impl<'s> FunctionBuilder<'s, FunctionTemplate> {
     .unwrap()
   }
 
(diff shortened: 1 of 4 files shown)


src/template.rs
```diff
@@ -376,23 +376,11 @@ impl<'s> FunctionBuilder<'s, FunctionTemplate> {
     self,
     scope: &mut HandleScope<'s, ()>,
     overload1: &dyn FastFunction,
+    c_fn1: *const CFunctionInfo,
```
bartlomieju, line 379: I think this could be `Option<*const CFunctionInfo>` - if it's not present it will be automatically created for the caller

littledivy (approved):
LGTM

PR #1199, opened 2023-03-16
Inline hint for default FastFunction trait methods

littledivy:


src/fast_api.rs
```diff
@@ -250,9 +250,11 @@ impl<T: Default> FastApiTypedArray<T> {
...
 
 pub trait FastFunction {
+  #[inline(always)]
   fn args(&self) -> &'static [Type] {
     &[]
   }
+  #[inline(always)]
   fn return_type(&self) -> CType {
     CType::Void
   }
```
bartlomieju, line 260: The one below?

bartlomieju (approved):

src/fast_api.rs
littledivy: It does not have a default impl, just type.

PR #1202, opened 2023-03-21
build: add PYTHON envvar to specify your python binary

levex:
macOS no longer ships with `python` symlinked, so let's add
an environment variable so that people can easily build rusty_v8.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -34,6 +34,7 @@ fn main() {
     "SCCACHE",
     "V8_FORCE_DEBUG",
     "V8_FROM_SOURCE",
+    "PYTHON",
   ];
   for env in envs {
     println!("cargo:rerun-if-env-changed={}", env);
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
LGTM

build.rs
```diff
@@ -638,6 +640,14 @@ fn gn() -> String {
   env::var("GN").unwrap_or_else(|_| "gn".to_owned())
 }
 
+/*
+ * Get the system's python binary - specified via the PYTHON environment
+ * variable or defaulting to `python3`.
+ */
+fn python() -> String {
+  env::var("PYTHON").unwrap_or_else(|_| "python3".to_owned())
```
avindra, line 648: Thanks, this allows me to remove a hack I was using:

```sh
# workaround to use python3
# where "python" is invoked
mkdir -p "$(pwd)/bin"
ln -sf %{_bindir}/python3 "$(pwd)/bin/python"
export PATH="$PATH:$(pwd)/bin"
```

levex: Happy to hear that!

PR #1206, opened 2023-03-31
FastApiOneByteString is not always utf-8

littledivy:
This commit removes the `as_str()` method and adds the `as_bytes` method to access the Latin-1 data. The utf8 assumption is wrong, the application should validate before converting to Rust string.


aapoalas (approved):
LGTM.

lucacasonato (approved):

PR #1209, opened 2023-04-10
Support for unconventional builds

hack3ric:
We may build rusty_v8 on platforms other than the common ones (for example, linux/riscv64), though they are not officially supported and prebuilt binaries are not available. This patch adds several environment variables:

- `DISABLE_CLANG`: use `is_clang=false` and disables Clang-specific options for building with GCC, etc.
- `EXTRA_GN_ARGS`: allow to pass extra GN arguments, such as custom toolchain
- `NO_PRINT_GN_ARGS`: Disables printing GN arguments to stdout. GN currently segfaults on riscv64 when printing arguments. Since it is trivial, we can have an option to simply disable it.

Also added fallback `arch` and `os` declaration in `platform()` to remove compile-time error.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -35,6 +35,9 @@ fn main() {
     "V8_FORCE_DEBUG",
     "V8_FROM_SOURCE",
     "PYTHON",
+    "DISABLE_CLANG",
+    "EXTRA_GN_ARGS",
+    "NO_PRINT_GN_ARGS",
   ];
   for env in envs {
     println!("cargo:rerun-if-env-changed={}", env);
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -127,7 +130,11 @@ fn build_v8() {
     gn_args.push("host_cpu=\"arm64\"".to_string())
   }
 
-  if let Some(clang_base_path) = find_compatible_system_clang() {
+  if env::var_os("DISABLE_CLANG").is_some() {
+    gn_args.push("is_clang=false".into());
+    // -gline-tables-only is Clang-only
+    gn_args.push("line_tables_only=false".into());
```
ry, line 136: This could be covered by EXTRA_GN_ARGS, right?

ry (approved):
LGTM - thanks @hack3ric, sorry for the slow review

PR #1211, opened 2023-04-21
feat: Add v8::StackTrace::current_script_name_or_source_url

bartlomieju:


littledivy (approved):
LGTM

lucacasonato (commented):
Bit contrived, but we don't actually ensure in the test that the fn is called

levex (approved):

PR #1222, opened 2023-04-27
feat: Add the option to disable generation from strings

bakjos:
It also includes the IsSetGeneratorObject option


bartlomieju (approved):
LGTM, thanks

Issue #1226, opened 2023-05-08
Exceptions thrown across boundaries behave differently on Win/Linux


futscdav:
Consider this simple example of having a rust print function that calls a `repr` function on passed objects. As `repr` can be arbitrary Js code, it could throw an exception. Since the exception occurs inside the Rust -> Js -> Rust -> Js chain, the only plausible course of action is to return control to the Js engine via normal means and hope that it escapes from there into the intended `TryCatch` handler.

```rust
use v8;

fn call_object_property<'s>(
        scope: &mut v8::HandleScope<'s>, 
        value: v8::Local<v8::Value>,
        property: &str) -> Option<v8::Local<'s, v8::Value>>  {
    let object: v8::Local<v8::Object> = value.try_into().unwrap();
    let ident = v8::String::new(scope, property).unwrap().into();
    let prop = object.get(scope, ident).unwrap();
    let func: v8::Local<v8::Function> = prop.try_into().unwrap();
    let recv = scope.get_current_context().global(scope).into();

    let retval = func.call(scope, recv, &[]);
    return retval
}

fn print_fn(scope: &mut v8::HandleScope,
    args: v8::FunctionCallbackArguments,
    _retval: v8::ReturnValue) {
    let local_arg = args.get(0);
    let print_str = if local_arg.is_string() {
        local_arg.to_rust_string_lossy(scope)
    } else if local_arg.is_object() {
        let obj_repr = call_object_property(scope, local_arg, "repr");
        if obj_repr.is_none() {
            return;
        }
        "[".to_owned() + &obj_repr.unwrap().to_rust_string_lossy(scope) + "]"
    } else {
        "Unknown type".to_owned()
    };
    println!("{print_str}");
}

fn main() {
    let platform = v8::new_default_platform(0, false).make_shared();
    v8::V8::initialize_platform(platform);
    v8::V8::initialize();
    
    let mut isolate = v8::Isolate::new(v8::CreateParams::default());
    let scope = &mut v8::HandleScope::new(&mut isolate);
    let local_context = v8::Context::new(scope);
    let scope = &mut v8::ContextScope::new(scope, local_context);

    let global_proxy = scope.get_current_context().global(scope);

    let identifier = v8::String::new(scope, "print").unwrap();
    let value = v8::FunctionTemplate::new(scope, print_fn).get_function(scope).unwrap();
    global_proxy.set(scope, identifier.into(), value.into());

    let code = r#"
    let object_with_ok_repr = {
[...]

PR #1227, opened 2023-05-10
test: Add tests showing incorrect exception behavior in debug build

bartlomieju:
Reproduces https://github.com/denoland/rusty_v8/issues/1226 and https://github.com/denoland/deno/issues/19021

```
// Fails
$ V8_FROM_SOURCE=1 cargo test exception
// Passes
$ V8_FROM_SOURCE=1 cargo test --release exception
```

We bisected this and this problem first appeared with V8 v11.2 upgrade. After further
bisects we established that https://github.com/v8/v8/commit/1f349da554800d9278cd1472850ca90a1fd7dc1e#diff-de75fc3e7b84373d94e18b4531827e8b749f9bbe05b59707e894e4e0ce2a1535
is the first V8 commit that causes this failure. However after more investigation we can't
find any reason why that particular commit might cause this problem.

It is only reproducible in debug build, but not release build. Current working theory
is that it is a Rust compiler bug as changing the optimization level for this code
makes the bug go away. This commit should be considered a band-aid that works
around the problem, but doesn't fix it completely. We are gonna go with it as it
unblocks un on day-to-day work, but ultimately we should track it down (or wait
for another Rust upgrade which might fix it?).

### DIFF
--- a/src/script.rs
+++ b/src/script.rs
@@ -80,7 +80,7 @@ impl Script {
   /// Runs the script returning the resulting value. It will be run in the
   /// context in which it was created (ScriptCompiler::CompileBound or
   /// UnboundScript::BindToCurrentContext()).
-  #[inline(always)]
+  #[inline]
   pub fn run<'s>(
     &self,
     scope: &mut HandleScope<'s>,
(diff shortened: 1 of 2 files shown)


piscisaureus (approved):
Nice tests! 
I would suggest updating the commit message, but LGTM otherwise.

PR #1229, opened 2023-05-12
Fix exception not caught in debug build for v8::Function::call

bartlomieju:
Follow up to https://github.com/denoland/rusty_v8/commit/6ac0fabb56ea20a4ec778d014eedb7f8a3bc0ec7 which
exhibits the same bug in `v8::Function::call`.

This will allow to fix https://github.com/denoland/deno/issues/19021 (checked in local build).

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -9745,7 +9745,8 @@ fn bubbling_up_exception_in_function_call() {
 
   let scope = &mut v8::TryCatch::new(scope);
   let call_boom_fn_val = script.run(scope).unwrap();
-  let call_boom_fn = v8::Local::<v8::Function>::try_from(call_boom_fn_val).unwrap();
+  let call_boom_fn =
+    v8::Local::<v8::Function>::try_from(call_boom_fn_val).unwrap();
 
   let scope = &mut v8::TryCatch::new(scope);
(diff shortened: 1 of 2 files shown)


dsherret (approved):
LGTM

tests/test_api.rs
```diff
@@ -9702,6 +9702,60 @@ try {
...
+  let code = r#"
+(function () {
+  return function callBoom() {
+    try {
+        boom()
+    } catch (e) {
+        //
+    }
+  }
+})();
```
ry, line 9740: I think you can just do
```js
 function callBoom() {
    try {
        boom()
    } catch (e) {
        //
    }
  }
 ```

bartlomieju: No, it doesn't work

piscisaureus: However, this *would* work:

```js
(function callBoom() {
    try {
        boom()
    } catch (e) {
        //
    }
  })
  ```
   
That's because, to avoid ambiguity between function expressions and function declarations, [an expression cannot start with the keyword `function`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/function#sect1). However it can start with a `(`, obviously!

PR #1237, opened 2023-05-23
Improved ObjectTemplate::set_*_handlers

lucacasonato:
Prior to this commit, `v8::NamedPropertyHandlerConfiguration`
and `v8::IndexedPropertyHandlerConfiguration` did not expose the
`definer` hook, or `flags`.

This commit adds these options. In the process of doing this a couple of
other changes were made:

- Bitflag enum consts are now member consts of the related struct.
  This is done because PropertyHandlerFlags has conflicts with
  PropertyAttribute.
- PropertyDescriptor gets all C++ introspection methods exposed to Rust.
- NamedPropertyHandlerConfiguration callback types get rustdoc comments.
- IndexedPropertyHandlerConfiguration callback types get rustdoc
  comments.
- GenericNamedPropertySetterCallback gets a ReturnValue parameter, to
  signal trap passthrough.


src/gc.rs
```diff
@@ -7,17 +7,19 @@
 #[derive(Debug, Eq, PartialEq, Clone, Copy)]
 pub struct GCType(u32);
 
-pub const GC_TYPE_TYPE_SCAVENGE: GCType = GCType(1);
+impl GCType {
```
lucacasonato, line 10: Enum members are now accessed as:

```
v8::GcType::MINOR_MARK_COMPACT
```

rather than

```
v8::GC_TYPE_MINOR_MARK_COMPACT
```

src/gc.rs
```diff
@@ -7,17 +7,19 @@
...
+  pub const MINOR_MARK_COMPACT: Self = Self(1 << 1);
 
-pub const GC_TYPE_MARK_SWEEP_COMPACT: GCType = GCType(4);
+  pub const MARK_SWEEP_COMPACT: Self = Self(1 << 2);
 
-pub const GC_TYPE_INCREMENTAL_MARKING: GCType = GCType(8);
+  pub const INCREMENTAL_MARKING: Self = Self(1 << 3);
 
-pub const GC_TYPE_PROCESS_WEAK_CALLBACK: GCType = GCType(16);
+  pub const PROCESS_WEAK_CALLBACKS: Self = Self(1 << 4);
```
ry, line 20: for these flags is there a way we could assert they're the same as in v8?

lucacasonato: Not sure - what do you suggest?

These are tested in test_api.rs in the sense that they are used and we check they behave correctly. Is that enough?

ry: nevermind - I can't think of a way to assert them being the same as in C

src/external_references.rs
```diff
@@ -1,14 +1,26 @@
+use crate::IndexedDefinerCallback;
+use crate::IndexedGetterCallback;
+use crate::IndexedSetterCallback;
+use crate::NamedDefinerCallback;
+use crate::NamedSetterCallback;
+use crate::PropertyEnumeratorCallback;
```
bartlomieju, line 6: Can you move these under the copyright?

src/property_handler_flags.rs
```diff
@@ -0,0 +1,134 @@
+#[repr(C)]
```
bartlomieju, line 1: Add copyright

bartlomieju (approved):
LGTM, nice work. Do you want me to cut a rusty_v8 release after this lands?

PR #1238, opened 2023-05-23
feat(fastcall): Int64Representation

aapoalas:
The new V8 supports fast calls returning 64-bit integers. With the default setting of `Int64Representation::kNumber` this just means that return values can be lossy 64-bit integers. Any data loss is, as they say, your problem.

With `Int64Representation::kBigInt` all 64-bit integers both on the parameter and return value side will be handled as BigInts: Calling `fast_call(32234646)` will never enter the fast call path with this representation but instead `fast_call(3234226n)` needs to be used. With this representation, any 64-bit integer return values are of course also returned as BigInts.

Performance measurements performed by a V8 intern gave numbers like this:

Microbenchmark | Operand range | As Number (us) | BigInt Only (us)
-- | -- | -- | --
Call-and-store | Smi (31-bit signed) | 12603.28 | 13188.74
Call-and-store | Safe integer (53-bit) | 15899.5 | 12929.22
Call-and-store | Small bigint (64-bit signed) | 15138.96 | 12953.86
Call-and-mul | Smi (31-bit signed) | 11710.8 | 13918.4
Call-and-mul | Safe integer (53-bit) | 14769.72 | 14115.7
Call-and-mul | Small bigint (64-bit signed) | 14739.52 | 13910.16
Call-and-mul-and-compare | Smi (31-bit signed) | 12045.62 | 13666.82
Call-and-mul-and-compare | Safe integer (53-bit) | 14082.04 | 13799.14
Call-and-mul-and-compare | Small bigint (64-bit signed) | 14005.88 | 13947.42

Effectively: Returning 64-bit integers as numbers is faster when the numbers are in the Smi range (31-bit signed) but even then are not much faster than returning BigInts. BigInts are faster in other cases.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1979,9 +1979,10 @@ v8::CTypeInfo* v8__CTypeInfo__New__From__Slice(unsigned int len,
 
 v8::CFunctionInfo* v8__CFunctionInfo__New(const v8::CTypeInfo& return_info,
                                           unsigned int args_len,
-                                          v8::CTypeInfo* args_info) {
+                                          v8::CTypeInfo* args_info,
+                                          Int64Representation repr) {
   std::unique_ptr<v8::CFunctionInfo> info = std::make_unique<v8::CFunctionInfo>(
-      v8::CFunctionInfo(return_info, args_len, args_info));
+      v8::CFunctionInfo(return_info, args_len, args_info, repr));
   return info.release();
 }
 
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1980,7 +1980,7 @@ v8::CTypeInfo* v8__CTypeInfo__New__From__Slice(unsigned int len,
 v8::CFunctionInfo* v8__CFunctionInfo__New(const v8::CTypeInfo& return_info,
                                           unsigned int args_len,
                                           v8::CTypeInfo* args_info,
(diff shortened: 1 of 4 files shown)


aapoalas:
I don't feel great about adding a `new_with_bigint` API for `FastFunction` for two reasons:
1. Bikeshedding wise it feels horrible.
2. The representation would probably be better as a parameter, but that'd be an API break for what is, in the end, a feature that most users (deno fast ops) will probably want to keep the same as it was for at least the immediate future.

bartlomieju:
> I don't feel great about adding a `new_with_bigint` API for `FastFunction` for two reasons:
> 
> 1. Bikeshedding wise it feels horrible.
> 2. The representation would probably be better as a parameter, but that'd be an API break for what is, in the end, a feature that most users (deno fast ops) will probably want to keep the same as it was for at least the immediate future.

I think it's fine - if it unblocks us on supporting it in CLI then we can land it. Another option would be to change the signature altogether and require caller to pass an enum defining if we want to use number or int64. I'm gonna approve this PR, feel free to land as is or change to use an enum.

bartlomieju (approved):
LGTM

aapoalas:
I don't actually have merge rights here (good and proper restriction, IMO).

Please merge.

PR #1240, opened 2023-05-25
Add more Object methods

lucacasonato:
Specifically Object::get_own_property_descriptor
and Object::get_property_attributes

### DIFF
--- a/src/object.rs
+++ b/src/object.rs
@@ -768,10 +768,11 @@ impl Object {
     .into()
   }
 
-  /// Gets the property attributes of a property which can be [None] or any
-  /// combination of [PropertyAttribute::READ_ONLY],
-  /// [PropertyAttribute::DONT_ENUM] and [PropertyAttribute::DONT_DELETE].
-  /// Returns [None] when the property doesn't exist.
+  /// Gets the property attributes of a property which can be
+  /// [PropertyAttribute::NONE] or any combination of
+  /// [PropertyAttribute::READ_ONLY], [PropertyAttribute::DONT_ENUM] and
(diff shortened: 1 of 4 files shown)


src/object.rs
```diff
@@ -755,6 +767,45 @@ impl Object {
...
+        &*key,
+        &mut out,
+      )
+    };
+    out.into()
+  }
+
+  /// Implements Object.getOwnPropertyDescriptor(O, P), see
+  /// https://tc39.es/ecma262/#sec-object.getownpropertydescriptor.
+  pub fn get_own_property_descriptor<'s>(
```
bartlomieju, line 794: Any chance you could test a code path that returns `None`?

src/object.rs
```diff
@@ -755,6 +767,45 @@ impl Object {
     }
     .into()
   }
+
+  /// Gets the property attributes of a property which can be [None] or any
+  /// combination of [PropertyAttribute::READ_ONLY],
+  /// [PropertyAttribute::DONT_ENUM] and [PropertyAttribute::DONT_DELETE].
+  /// Returns [None] when the property doesn't exist.
+  pub fn get_property_attributes(
```
bartlomieju, line 775: Ditto

lucacasonato: Done.

src/object.rs
lucacasonato: V8 seems to never return an Empty MaybeLocal here. They also don't test this in their test_api.cc.

bartlomieju (approved):
LGTM

PR #1244, opened 2023-06-04
Don't preserve V8 archive mode and ownership on build

fracek:
By default, copying a file will preserve its mode and ownership attributes.
This is an issue when copying the V8 archive from a read-only filesystem since the archive file also becomes read-only, so subsequent builds will fail.

We now create a new destination file and copy the content of the archive to it, this ensures the destination file has the default attributes.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -3,6 +3,7 @@ use fslock::LockFile;
 use std::collections::HashSet;
 use std::env;
 use std::fs;
+use std::io;
 use std::path::Path;
 use std::path::PathBuf;
 use std::process::exit;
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -383,7 +384,10 @@ fn build_dir() -> PathBuf {
 
 fn download_file(url: String, filename: PathBuf) {
   if !url.starts_with("http:") && !url.starts_with("https:") {
-    fs::copy(&url, filename).unwrap();
+    let mut src = fs::File::open(&url).unwrap();
+    // create new file if not exists, or overwrite existing
+    let mut dst = fs::File::create(&filename).unwrap();
+    io::copy(&mut src, &mut dst).unwrap();
```
ry, line 390: Please create a utility function and add a comment about its purpose.

ry (approved):
lgtm - thanks!

PR #1250, opened 2023-06-21
Add `Function::{ScriptId, GetScriptOrigin}, ScriptOrigin::Get{ScriptId, ResourceName, SourceMapUrl}` bindings

rakeeb-hossain:
I added the bindings in the title to use in an application! Let me know if I'm missing anything.

### DIFF
--- a/src/function.rs
+++ b/src/function.rs
@@ -10,7 +10,6 @@ use crate::support::MapFnTo;
 use crate::support::ToCFn;
 use crate::support::UnitType;
 use crate::support::{int, Opaque};
-use crate::undefined;
 use crate::Context;
 use crate::Function;
 use crate::HandleScope;
@@ -23,6 +22,7 @@ use crate::Signature;
 use crate::String;
 use crate::UniqueRef;
 use crate::Value;
+use crate::{undefined, ScriptOrigin};
 
 extern "C" {
   fn v8__Function__New(
@@ -50,6 +50,10 @@ extern "C" {
   fn v8__Function__SetName(this: *const Function, name: *const String);
   fn v8__Function__GetScriptColumnNumber(this: *const Function) -> int;
   fn v8__Function__GetScriptLineNumber(this: *const Function) -> int;
+  fn v8__Function__ScriptId(this: *const Function) -> int;
+  fn v8__Function__GetScriptOrigin(
+    this: *const Function,
+  ) -> *const ScriptOrigin<'static>;
 
   fn v8__Function__CreateCodeCache(
     script: *const Function,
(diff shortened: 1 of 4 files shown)


src/function.rs
```diff
@@ -903,6 +907,24 @@ impl Function {
...
+  pub fn get_script_origin(&self) -> &ScriptOrigin {
+    unsafe {
+      let ptr = v8__Function__GetScriptOrigin(self);
+      &*ptr
+    }
+  }
+
+  /// Returns scriptId.
+  #[inline(always)]
+  pub fn get_script_id(&self) -> Option<u32> {
```
mmastrac, line 920: Should this be i32?

rakeeb-hossain: Was unsure about this. V8 docs are sparse on when a script ID is < 0, but the only cases I could find in the source were 1) uninitialized ScriptOrigin classes (value is -1) and 2) kTemporaryScriptId = -2. Due to uncertainty on when these cases actually occur, I just treated these as None values.

I can change this to i32 instead if you think that would make more sense.

rakeeb-hossain: @mmastrac Would you suggest i32 instead?

bartlomieju: Yes, we should use `i32` here. Also please remove `get_` prefix - in `v8/include/v8-function.h` this function is named `ScriptId()`

src/script.rs
```diff
@@ -125,4 +132,28 @@ impl<'s> ScriptOrigin<'s> {
...
+  #[inline(always)]
+  pub fn get_resource_name(&self) -> Option<Local<'s, Value>> {
+    unsafe {
+      let ptr = v8__ScriptOrigin__GetResourceName(self);
+      Local::from_raw(ptr)
+    }
+  }
+
+  #[inline(always)]
+  pub fn get_source_map_url(&self) -> Option<Local<'s, Value>> {
```
bartlomieju, line 153: Please remove `get_` prefixes from these functions to match V8 API

src/script.rs
```diff
@@ -125,4 +132,28 @@ impl<'s> ScriptOrigin<'s> {
       buf.assume_init()
     }
   }
+
+  #[inline(always)]
+  pub fn get_script_id(&self) -> Option<u32> {
```
bartlomieju, line 137: Use `i32` here

bartlomieju (commented):
Almost there, just a few nitpicks to match V8 API

src/function.rs
rakeeb-hossain: Done @bartlomieju @mmastrac

bartlomieju (approved):
LGTM, thank you!

PR #1255, opened 2023-06-26
feat: Fix is_onebyte and add an uninit write for onebyte

mmastrac:
Some fixes around one-byte strings:

 - `is_onebyte` was calling the wrong v8 API.
 - We didn't have a way to write one-byte strings with uninitialized buffers
 - (bonus) The test_string method was quite slow making testing a bit of a pain


aapoalas (approved):
Nice catch!

bartlomieju (approved):
LGTM

PR #1256, opened 2023-06-26
feat: Use MaybeUninit for to_rust_string_lossy and add to_rust_cow_lossy

mmastrac:
Use allocator APIs for `to_rust_string_lossy`, and add a `to_rust_cow_lossy` method that allows callers to pass a `MaybeUninit` buffer for (possibly) malloc-free Rust strings.


bartlomieju (approved):
LGTM, do you expect having two separate paths for `to_rust_string_lossy()` will be faster?

mmastrac:
I think so as write_one_byte_uninit is basically a memcpy with no checks for UTF-8-ness. If we can fall into one-byte paths things become much faster.

PR #1269, opened 2023-07-07
silence warning for unused 'must use'

graham:
this warning only occurs when using the library as a dependency.

```
warning: unused return value of `Box::<T>::from_raw` that must be used
    --> /rusty_v8/src/scope.rs:1092:16
     |
1092 |       unsafe { Box::from_raw(root) };
     |                ^^^^^^^^^^^^^^^^^^^
     |
   = note: call `drop(Box::from_raw(ptr))` if you intend to drop the `Box`
   = note: `#[warn(unused_must_use)]` on by default
```


aapoalas (approved):
Nice catch, thank you!

mmastrac (approved):
LGTM

PR #1270, opened 2023-07-07
WIP: experiment: const static external strings

aapoalas:


src/string.rs
```diff
@@ -337,6 +371,35 @@ impl String {
     Self::new_from_utf8(scope, value.as_ref(), NewStringType::Normal)
   }
 
+  // Allocates a static ExternalOneByteConst and leaks it.
+  // The allocated ExternalOneByteConst can never be deallocated.
+  #[inline(always)]
+  pub const fn create_external_onebyte_const(
+    buffer: &'static [u8],
+  ) -> ExternalOneByteConst {
+    ExternalOneByteConst {
```
mmastrac, line 380: We have a const is_ascii in deno_core we can steal for this:

```
  /// Compile-time function to determine if a string is ASCII. Note that UTF-8 chars
  /// longer than one byte have the high-bit set and thus, are not ASCII.
  const fn is_ascii(s: &'static [u8]) -> bool {
    let mut i = 0;
    while i < s.len() {
      if !s[i].is_ascii() {
        return false;
      }
      i += 1;
    }
    true
  }
```

src/string.rs
```diff
@@ -88,6 +94,34 @@ extern "C" {
   #[allow(dead_code)]
   fn v8__String__IsOneByte(this: *const String) -> bool;
   fn v8__String__ContainsOnlyOneByte(this: *const String) -> bool;
+
+  #[link_name = "_ZTV34ExternalConstOneByteStringResource"]
```
mmastrac, line 98: Since we can't make this `const`-compatible, I wonder if we should return the vtable pointer itself from C++ code which should be compatible with pretty much any gcc/clang platform

mmastrac:
I love this concept and I am pretty sure we can make it work.

src/string.rs
aapoalas: That's definitely possible. The only downside is that we'll need to bind a static (maybe AtomicPtr again) that gets initialized by calling into C++, and that call needs to allocate an instance to get the vtable pointer out of it, then delete and return.

One alternative would be to have the `init` call actually call into C++ for in-place building if vtable pointer is not in place yet.

src/string.rs
aapoalas: Copied this over.

src/string.rs
aapoalas: Change to calling into C++ for in-place construction.

PR #1272, opened 2023-07-08
chore: Add Debug for ExternalReference

mmastrac:


bartlomieju (commented):
Sorry I missed this PR. Could you exercise this codepath in tests just to make sure it doesn't lead to a crash?

bartlomieju (approved):
Thank you!

PR #1273, opened 2023-07-09
Add v8::NamedPropertyHandlerConfiguration::*_raw methods

bartlomieju:
Needed for https://github.com/denoland/deno/pull/19307 due to possible Rust compiler bug.

This commit adds methods for setting "raw" callback that need to be `.map_fn_to()` manually
by a user.


piscisaureus (approved):
Straightforward implementation + test. LGTM

PR #1275, opened 2023-07-10
feat: Implement Rust-side const ExternalOneByteStringResource subclass

aapoalas:
MSVC and Itanium C++ ABIs agree that for simple inheritance the basic structure of a vtable contains metadata fields at a "negative offset" from the vtable pointer, and at zero or positive offsets come the virtual function pointers in the order of declaration. The only difference between the two is that MSVC only places the virtual deleting destructor in the vtable while Itanium ABI places both the deleting and the complete object destructors in it, leading to a vtable that is one pointer larger in Itanium / on Linux. Also MSVC only has a single metadata field instead of two for Itanium. Itanium inlines the base offset into the vtable while MSVC keeps it in what is essentially the entry point into the type info data.

Since the two are so similar, creating a custom vtable on Rust-side is pretty easy and can be done entirely at compile-time, meaning that instances of the class can also be created entirely at compile time. This leads to fully const external strings being possible.

The only thing that would be nice to add is some C++ side assertions for the vtable offsets but while member pointers to member functions are a thing, getting their offset is non-trivial. It's also a bit unclear to me if it's even possible to perform offset calculations with those.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1002,6 +1002,8 @@ int v8__String__WriteUtf8(const v8::String& self, v8::Isolate* isolate,
   return self.WriteUtf8(isolate, buffer, length, nchars_ref, options);
 }
 
+// NOTE: This class is never used and only serves as a reference for
+// the OneByteConst struct created on Rust-side.
 class ExternalStaticOneByteStringResource
     : public v8::String::ExternalOneByteStringResource {
  public:
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1002,8 +1002,6 @@ int v8__String__WriteUtf8(const v8::String& self, v8::Isolate* isolate,
   return self.WriteUtf8(isolate, buffer, length, nchars_ref, options);
 }
 
-// NOTE: This class is never used and only serves as a reference for
-// the OneByteConst struct created on Rust-side.
 class ExternalStaticOneByteStringResource
     : public v8::String::ExternalOneByteStringResource {
  public:
@@ -1017,6 +1015,8 @@ class ExternalStaticOneByteStringResource
   const int _length;
 };
 
+// NOTE: This class is never used and only serves as a reference for
+// the OneByteConst struct created on Rust-side.
 class ExternalConstOneByteStringResource
     : public v8::String::ExternalOneByteStringResource {
  public:
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1018,8 +1018,8 @@ class ExternalStaticOneByteStringResource
 class ExternalConstOneByteStringResource
     : public v8::String::ExternalOneByteStringResource {
  public:
-  ExternalConstOneByteStringResource(const char* data, int length)
-      : _data(data), _length(length) {
+  ExternalConstOneByteStringResource(int length)
+      : _length(length) {
     static_assert(offsetof(ExternalConstOneByteStringResource, _length) == 16,
                   "ExternalConstOneByteStringResource's length was not at offset 16");
     static_assert(sizeof(ExternalConstOneByteStringResource) == 24,
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1159,15 +1159,15 @@ void v8__ObjectTemplate__SetInternalFieldCount(const v8::ObjectTemplate& self,
   ptr_to_local(&self)->SetInternalFieldCount(value);
 }
 
-void v8__ObjectTemplate__SetAccessor(const v8::ObjectTemplate& self,
-                                     const v8::Name& key,
-                                     v8::AccessorNameGetterCallback getter,
-                                     v8::AccessorNameSetterCallback setter,
-                                     const v8::Value* data_or_null,
-                                     v8::PropertyAttribute attr) {
-  ptr_to_local(&self)->SetAccessor(ptr_to_local(&key), getter, setter,
-                                   ptr_to_local(data_or_null),
-                                   v8::AccessControl::DEFAULT, attr);
+void v8__ObjectTemplate__SetAccessor(
+    const v8::ObjectTemplate& self, const v8::Name& key,
+    v8::AccessorNameGetterCallback getter,
+    v8::AccessorNameSetterCallback setter,
+    const v8::Value* data_or_null,
+    v8::PropertyAttribute attr) {
+  ptr_to_local(&self)->SetAccessor(
+    ptr_to_local(&key), getter, setter,  ptr_to_local(data_or_null), v8::AccessControl::DEFAULT,
+    attr);
 }
 
 void v8__ObjectTemplate__SetNamedPropertyHandler(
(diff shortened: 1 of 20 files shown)


src/string.rs
```diff
@@ -431,7 +419,7 @@ impl String {
   ) -> OneByteConst {
     is_ascii(buffer);
     OneByteConst {
-      vtable: &ONE_BYTE_CONST_VTABLE.2,
+      vtable: &ONE_BYTE_CONST_VTABLE.delete1,
```
mmastrac, line 422: This is interesting -- you can take an address of a const?

aapoalas: > This is interesting -- you can take an address of a const?

Seems so. It was originally a static but clippy told me to make it a const.

aapoalas: image

src/string.rs
```diff
@@ -90,6 +95,75 @@ extern "C" {
...
+#[repr(C)]
+pub struct OneByteConst {
+  vtable: *const OneByteConstNoOp,
+  cached_data: *const char,
+  length: int,
+}
+
+unsafe impl Sync for OneByteConst {}
+
+extern "C" fn one_byte_const_no_op(_this: &'static OneByteConst) {}
```
mmastrac, line 107: I suspect these should be `*const OneByteConst`

src/string.rs
```diff
@@ -90,6 +95,75 @@ extern "C" {
...
+  unlock: OneByteConstNoOp,
+  data: OneByteConstData,
+  length: OneByteConstLength,
+}
+
+const ONE_BYTE_CONST_VTABLE: OneByteConstVtable = OneByteConstVtable {
+  _typeinfo: std::ptr::null(),
+  _base_offset: 0,
+  delete1: one_byte_const_no_op,
+  #[cfg(not(target_family = "windows"))]
```
mmastrac, line 144: Worth a comment here and above on what `delete2` represent for Windows.

src/string.rs
```diff
@@ -337,6 +411,34 @@ impl String {
...
+  ) -> OneByteConst {
+    is_ascii(buffer);
+    OneByteConst {
+      vtable: &ONE_BYTE_CONST_VTABLE.delete1,
+      cached_data: buffer.as_ptr() as *const char,
+      length: buffer.len() as i32,
+    }
+  }
+
+  // Creates a v8::String from a `&'static [u8]`,
```
mmastrac, line 428: Replace w/:

from a `&'static OneByteConst` which is guaranteed to be Latin-1 or ASCII.

src/binding.cc
```diff
@@ -1013,6 +1015,31 @@ class ExternalStaticOneByteStringResource
   const int _length;
 };
 
+class ExternalConstOneByteStringResource
```
mmastrac, line 1018: Maybe worth adding a note that this class is never actually used -- it only serves as a c++ reference to the one created on the Rust side.

mmastrac (approved):
LGTM. @bartlomieju I think we should land and do a small experiment in Deno to confirm that it works at scale (ideally in a pretty heavily used codepath).

bartlomieju:
Feel free to do it!

src/binding.cc
```diff
@@ -1013,6 +1015,33 @@ class ExternalStaticOneByteStringResource
...
+    : public v8::String::ExternalOneByteStringResource {
+ public:
+  ExternalConstOneByteStringResource(int length)
+      : _length(length) {
+    static_assert(offsetof(ExternalConstOneByteStringResource, _length) == 16,
+                  "ExternalConstOneByteStringResource's length was not at offset 16");
+    static_assert(sizeof(ExternalConstOneByteStringResource) == 24,
+                  "ExternalConstOneByteStringResource size was not 24");
+    static_assert(alignof(ExternalConstOneByteStringResource) == 8,
+                  "ExternalConstOneByteStringResource align was not 8");
```
avindra, line 1030: FWIW, the 32-bit build started failing here:

```
   ../../../src/binding.cc:68:1: error: static assertion failed due to requirement 'sizeof(v8::Isolate::DisallowJavascriptExecutionScope) == sizeof(unsigned int) * 2': DisallowJavascriptExecutionScope size mismatch
   static_assert(sizeof(v8::Isolate::DisallowJavascriptExecutionScope) ==
   ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   ../../../src/binding.cc:68:69: note: expression evaluates to '12 == 8'
   static_assert(sizeof(v8::Isolate::DisallowJavascriptExecutionScope) ==
                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
   ../../../src/binding.cc:1033:5: error: static assertion failed due to requirement '__builtin_offsetof(ExternalConstOneByteStringResource, _length) == 16': ExternalConstOneByteStringResource's length was not at offset 16
       static_assert(offsetof(ExternalConstOneByteStringResource, _length) == 16,
       ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   ../../../src/binding.cc:1033:73: note: expression evaluates to '8 == 16'
       static_assert(offsetof(ExternalConstOneByteStringResource, _length) == 16,
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
   ../../../src/binding.cc:1035:5: error: static assertion failed due to requirement 'sizeof(ExternalConstOneByteStringResource) == 24': ExternalConstOneByteStringResource size was not 24
       static_assert(sizeof(ExternalConstOneByteStringResource) == 24,
       ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   ../../../src/binding.cc:1035:62: note: expression evaluates to '12 == 24'
       static_assert(sizeof(ExternalConstOneByteStringResource) == 24,
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~
   ../../../src/binding.cc:1037:5: error: static assertion failed due to requirement 'alignof(ExternalConstOneByteStringResource) == 8': ExternalConstOneByteStringResource align was not 8
       static_assert(alignof(ExternalConstOneByteStringResource) == 8,
       ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   ../../../src/binding.cc:1037:63: note: expression evaluates to '4 == 8'
       static_assert(alignof(ExternalConstOneByteStringResource) == 8,
                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~
```

aapoalas: @avindra Thanks for letting us know and sorry about the break! Can you check if this PR fixes the build on 32-bit architectures? https://github.com/denoland/rusty_v8/pull/1289

avindra: > Can you check if this PR fixes the build on 32-bit architectures? #1289

Sorry for the late reply. The 32 bit build passes now!

PR #1280, opened 2023-07-11
feat(v8): Add a new_empty_backing_store method allowing us to create cheap and empty backing stores with no allocation

mmastrac:


src/array_buffer.rs
```diff
@@ -544,4 +544,27 @@ impl ArrayBuffer {
...
   }
+
+  #[inline(always)]
+  /// Create a cheap, zero-sized backing store.
+  pub fn new_empty_backing_store() -> UniqueRef<BackingStore> {
+    // A zero-sized-but-valid, mutable buffer
+    static mut BUFFER: [u8; 0] = [];
+    // A function that does nothing
+    extern "C" fn empty_deleter(
+      _ptr: *mut c_void,
```
aapoalas, line 555: **suggestion:** Expose the [EmptyDeleter](https://v8.github.io/api/head/classv8_1_1BackingStore.html#acb89ca23fdc981dd2b55cc6725ea4722) from V8.

mmastrac: Ah yeah, so apparently this is actually an optimization that V8 uses internally. We should be using this static value. But also, we should be exposing `EmptyBackingStore` itself.

bartlomieju:
@mmastrac did you have a specific use case in mind for this API? Can we finish the PR?

mmastrac:
@bartlomieju  I believe this was going to be useful for certain op2 optimizations where we are returning empty buffers, but it wasn't a critical change and should probably have us expose `EmptyBackingStore` (https://chromium.googlesource.com/v8/v8/+/833b3c96a6f5a345b44aeca9e0034c2f97b43d5f/src/objects/backing-store.cc#810) directly instead.

PR #1282, opened 2023-07-13
fix(OneByteConst): missing ASCII assertion, assert length

aapoalas:
We forgot to actually assert the ASCII-ness instead of just checking it and ignoring the result. While here it's best to also assert that the buffer doesn't overflow V8's String max length. Finally: We do not need to store the buffer length as an `i32` since it both comes in and is returned as a `usize`: We can just use a `usize` directly.

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -100,7 +100,7 @@ extern "C" {
 pub struct OneByteConst {
   vtable: *const OneByteConstNoOp,
   cached_data: *const char,
-  length: int,
+  length: usize,
 }
 
 // SAFETY: The vtable for OneByteConst is an immutable static and all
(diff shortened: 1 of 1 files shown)


mmastrac (approved):
LGTM. I initially wondered about the fact that we're mostly going to store zeroes there, but Rust was just padding this struct to 8-byte alignment anyways.

src/string.rs
```diff
@@ -121,7 +121,7 @@ extern "C" fn one_byte_const_data(this: *const OneByteConst) -> *const char {
 }
 extern "C" fn one_byte_const_length(this: *const OneByteConst) -> usize {
```
mmastrac, line 122: Just to sanity check -- should this return `int`? I checked the v8 ref and it's a 4-byte return value. While this _shouldn't_ matter in most cases, should we make this int as well?

PR #1289, opened 2023-07-26
fix: 32-bit build fails on non-size_t based size assertions

aapoalas:
Recent static assert additions did not either use size_t based assertions or would not function with them because of padding required. This PR tries to fix the issue.


bartlomieju (approved):

avindra:
Confirming the 32-bit build passes.

https://github.com/denoland/rusty_v8/pull/1275#discussion_r1287424278

PR #1297, opened 2023-08-09
Remove a redundant and confusing sentence from FAQ

troyjfarrell:
"In the limit…" should probably read "In the end…" or "In the future…", but the last sentence makes the first redundant, so this PR removes the confusing sentence.


mmastrac (approved):
LGTM

PR #1308, opened 2023-08-28
feat: support iOS build

leanmendoza:
This enables at least the development (V8_FROM_SOURCE) on iOS.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -181,6 +181,20 @@ fn build_v8() {
       maybe_install_sysroot("amd64");
     };
 
+    if target_triple == "aarch64-apple-ios" {
+      gn_args.push(r#"v8_target_cpu="arm64""#.to_string());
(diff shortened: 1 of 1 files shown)


bartlomieju:
Thanks for the original prototype here, and sorry it sat open so long.

rusty_v8 has since grown iOS support in `build.rs` that covers this and more: it keys off `target_os == "ios"` (so it handles the simulator and non-aarch64 targets too, not just `aarch64-apple-ios`), sets `target_environment` for device vs. simulator, uses a current `ios_deployment_target`, and makes device builds jitless with Sparkplug/Maglev/Turbofan and WebAssembly disabled — which iOS requires for non-WebKit apps. It also respects an explicit `target_os="ios"` in `GN_ARGS` as an override.

Closing as superseded. Please reopen an issue if the current support doesn't work for your setup.

PR #1312, opened 2023-09-07
fix: fix make_pod static asserts

CGQAQ:
* std::is_pod is deprecated in c++20 changed to std::is_standard_layout && std::is_trivial
* V shoud be trivially_copyable in order to copy from

### DIFF
--- a/src/support.h
+++ b/src/support.h
@@ -78,7 +78,7 @@ struct make_pod {
   template <class V>
   union helper {
     static_assert(std::is_trivially_copyable<V>::value, "type V must be trivially copyable");
-    static_assert(std::is_standard_layout<P>::value && std::is_trivially_copyable<P>::value, "type P must a pod type");
+    static_assert(std::is_standard_layout<P>::value && std::is_trivial<P>::value, "type P must a pod type");
     static_assert(sizeof(V) == sizeof(P), "type P must be same size as type V");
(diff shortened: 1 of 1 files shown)


CGQAQ:
Seems like we are memcpying std::shared_ptr in the test

bartlomieju:
Thanks, and sorry for the long delay here.

The `std::is_pod<P>` -> `std::is_trivial<P> && std::is_standard_layout<P>` half of this landed independently in ffd2b29 (#1755), so that part is already in `src/support.h`.

The remaining part — `static_assert(std::is_trivially_copyable<V>::value)` — unfortunately can't be applied: `V` is the source type, and every caller passes something that is deliberately *not* trivially copyable (`std::shared_ptr<v8::BackingStore>`, `std::shared_ptr<v8::Platform>`, `std::string`, `v8::Global`). That's the whole purpose of `make_pod`: move a non-POD value into a union whose destructor is a no-op, memcpy the bytes out, and thereby transfer ownership across the FFI boundary into Rust. With that assert, the class would have nothing left to do and the build would fail at every call site.

Your follow-up comment about memcpying a `shared_ptr` is a fair point — it isn't strictly standard-conforming — but fixing that properly would mean reworking the ownership transfer at all the call sites rather than adding the assert. Closing for now; happy to look at a proposal along those lines if you're interested.

PR #1313, opened 2023-09-07
fix windows builds

CGQAQ:
Windows don't have `python3`

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -679,7 +679,13 @@ fn gn() -> String {
  * variable or defaulting to `python3`.
  */
 fn python() -> String {
-  env::var("PYTHON").unwrap_or_else(|_| "python3".to_owned())
+  env::var("PYTHON").unwrap_or_else(|_| {
+    if cfg!(target_os = "windows") {
+      "python".to_owned()
+    } else {
+      "python3".to_owned()
+    }
+  })
 }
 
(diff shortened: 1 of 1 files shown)


linrongbin16:
> Windows don't have `python3`

It depends on different ways of installing python.

Official python window installer indeed doesn't provide a `python3.exe`.

While others for example scoope may provide `python3.exe` instead of `python.exe`.

Maybe keep the code simple and stupid, don't involve too much logics here. Anyway you can use the `PYTHON` env var to customize.

bartlomieju:
Thanks for this, and apologies for the very long delay.

We're going to leave the default as `python3`. Swapping it to `python` on Windows can regress in a worse way: a bare `python` frequently resolves to the Microsoft Store app-execution-alias stub, which silently opens the Store instead of running Python, and on older setups it can still be Python 2. As @linrongbin16 noted, whether `python3.exe` exists depends on how Python was installed — scoop/winget do provide it, and `actions/setup-python` does too, which is why our from-source Windows CI builds pass.

If your Windows install only has `python.exe`, set `PYTHON=python` — that env var exists exactly for this and is already wired into `rerun-if-env-changed`.

If this keeps biting people, the right fix is probably to probe (try `python3`, fall back to `python` when it isn't found) rather than to guess from the host OS. Closing for now.

PR #1318, opened 2023-09-12
fix: crash on x86_64 systems that support memory protection keys

piscisaureus:
This fixes the issue.

It needs some comments explaining why one would need `new_unprotected_default_platform()` (see TODOs).
@mmastrac Can you help out with that?

### DIFF
--- a/src/platform.rs
+++ b/src/platform.rs
@@ -14,6 +14,10 @@ extern "C" {
     thread_pool_size: int,
     idle_task_support: bool,
   ) -> *mut Platform;
+  fn v8__Platform__NewUnprotectedDefaultPlatform(
+    thread_pool_size: int,
+    idle_task_support: bool,
+  ) -> *mut Platform;
   fn v8__Platform__NewSingleThreadedDefaultPlatform(
     idle_task_support: bool,
   ) -> *mut Platform;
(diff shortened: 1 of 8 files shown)


src/platform.rs
```diff
@@ -58,6 +62,10 @@ pub struct Platform(Opaque);
 /// If |idle_task_support| is enabled then the platform will accept idle
 /// tasks (IdleTasksEnabled will return true) and will rely on the embedder
 /// calling v8::platform::RunIdleTasks to process the idle tasks.
+///
+/// TODO: reference the threading restrictions that apply when V8 is initialized
+/// with the default platform, the implications for the Cargo test harness, and
+/// when to use new_unprotected_default_platform().
```
mmastrac, line 68:
```suggestion
///
/// The default platform for v8 may include restrictions and caveats on thread creation
/// and initialization. This platform should only be used in cases where v8 can be
/// reliably initialized on the application's main thread, or the parent thread to all 
/// threads in the system that will use v8.
///
/// One example of a restriction is the use of Memory Protection Keys (pkeys) on
/// modern Linux systems using modern Intel/AMD processors. This particular technology
/// requires that all threads using v8 are created as descendent threads of the thread
/// that called `v8::Initialize`. 
```

src/platform.rs
```diff
@@ -101,6 +122,20 @@ impl Platform {
     }
   }
 
+  /// TODO: doc comment
+  #[inline(always)]
```
mmastrac, line 126:
```suggestion
/// Creates a platform that is identical to the default platform, but does not enforce
/// thread-isolated allocations. This may reduce security in some cases, so this method
/// should be used with caution in cases where the threading guarantees of
/// `new_default_platform` cannot be upheld (generally for tests).
#[inline(always)]
```

src/platform.rs
```diff
@@ -66,6 +74,15 @@ pub fn new_default_platform(
   Platform::new(thread_pool_size, idle_task_support)
 }
 
+/// TODO: doc comment
+#[inline(always)]
```
mmastrac, line 78:
```suggestion
/// Creates a platform that is identical to the default platform, but does not enforce
/// thread-isolated allocations. This may reduce security in some cases, so this method
/// should be used with caution in cases where the threading guarantees of
/// `new_default_platform` cannot be upheld (generally for tests).
#[inline(always)]
```

src/platform.rs
```diff
@@ -88,6 +105,10 @@ impl Platform {
   /// If |idle_task_support| is enabled then the platform will accept idle
   /// tasks (IdleTasksEnabled will return true) and will rely on the embedder
   /// calling v8::platform::RunIdleTasks to process the idle tasks.
+  ///
+  /// TODO: reference the threading restrictions that apply when V8 is
+  /// initialized with the default platform, the implications for the Cargo
+  /// test harness, and when to use new_unprotected().
```
mmastrac, line 111:
```suggestion
/// The default platform for v8 may include restrictions and caveats on thread creation
/// and initialization. This platform should only be used in cases where v8 can be
/// reliably initialized on the application's main thread, or the parent thread to all 
/// threads in the system that will use v8.
///
/// One example of a restriction is the use of Memory Protection Keys (pkeys) on
/// modern Linux systems using modern Intel/AMD processors. This particular technology
/// requires that all threads using v8 are created as descendent threads of the thread
/// that called `v8::Initialize`.
```

mmastrac (approved):
LGTM!

PR #1322, opened 2023-09-16
feat: Support custom host objects in ValueSerializer

lrowe:
Add v8::ValueSerializerImpl::{has_custom_host_object,is_host_object} equivalents for v8::ValueSerializer::Delegate::{HasCustomHostObject,IsCustomHostObject}.

This enables serializing custom host objects without embedder fields.

* V8 commit: v8/v8@cf13b9b46572a9824d2d632abdd48c56161ace02
* V8 bug: https://bugs.chromium.org/p/v8/issues/detail?id=11927
* Deno bug: denoland/deno#12067

### DIFF
--- a/src/value_serializer.rs
+++ b/src/value_serializer.rs
@@ -46,6 +46,34 @@ pub unsafe extern "C" fn v8__ValueSerializer__Delegate__ThrowDataCloneError(
 }
 
 #[no_mangle]
+pub unsafe extern "C" fn v8__ValueSerializer__Delegate__HasCustomHostObject(
+  this: &mut CxxValueSerializerDelegate,
+  _isolate: *mut Isolate,
+) -> bool {
+  let value_serializer_heap = ValueSerializerHeap::dispatch_mut(this);
+  let scope =
+    &mut crate::scope::CallbackScope::new(value_serializer_heap.context);
+  value_serializer_heap
+    .value_serializer_impl
+    .as_mut()
+    .has_custom_host_object(scope)
+}
+
+#[no_mangle]
+pub unsafe extern "C" fn v8__ValueSerializer__Delegate__IsHostObject(
+  this: &mut CxxValueSerializerDelegate,
+  _isolate: *mut Isolate,
+  object: Local<Object>,
+) -> MaybeBool {
+  let value_serializer_heap = ValueSerializerHeap::dispatch_mut(this);
+  let scope =
+    &mut crate::scope::CallbackScope::new(value_serializer_heap.context);
+  let value_serializer_impl =
+    value_serializer_heap.value_serializer_impl.as_mut();
+  MaybeBool::from(value_serializer_impl.is_host_object(scope, object))
+}
+
+#[no_mangle]
 pub unsafe extern "C" fn v8__ValueSerializer__Delegate__WriteHostObject(
   this: &mut CxxValueSerializerDelegate,
   _isolate: *mut Isolate,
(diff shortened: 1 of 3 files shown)


lrowe:
Note. While I am fairly confident the changes in this PR are correct, I am still working through the corresponding changes in deno_core required to make use of this functionality.

src/value_serializer.rs
```diff
@@ -45,6 +45,34 @@ pub unsafe extern "C" fn v8__ValueSerializer__Delegate__ThrowDataCloneError(
...
+  this: &mut CxxValueSerializerDelegate,
+  _isolate: *mut Isolate,
+) -> bool {
+  let value_serializer_heap = ValueSerializerHeap::dispatch_mut(this);
+  let scope =
+    &mut crate::scope::CallbackScope::new(value_serializer_heap.context);
+  value_serializer_heap
+    .value_serializer_impl
+    .as_mut()
+    .has_custom_host_object(scope)
```
legendecas, line 59: Is HandleScope needed in `HasCustomHostObject`? The original signature doesn't expect this method to call into JavaScript.

lrowe: Probably not. I simply added this for consistency with the other `ValueSerializerImpl` trait methods that pass `HandleScope` where the C++ `v8::ValueSerializer::Delegate` methods pass the `Isolate`.

Digging back through git blame the reasoning for that divergence seems to be that using `Isolate` requires unsafe code whereas `HandleScope` does not. (Discussed from here: https://github.com/denoland/rusty_v8/pull/442#issuecomment-682495244.)

As I understand it, the goal of these bindings is to match the original API as closely as possible and to provide safe abstractions where possible. The original signature for `HasCustomHostObejct` includes the Isolate. I'm not sure what  type (if any) that should translate to.

I see that you were the author of the V8 commit, did you have any uses in mind for it or was it for consistency with the other methods? It is presumably cost free to do so there whereas I assume creating a HandleScope is not.

https://github.com/denoland/rusty_v8/blob/2545a6159f7ef174538c84c5a0c0e0949222835c/src/value_serializer.rs#L242-L250

https://github.com/v8/v8/blob/7fdc2728f2114b083ebbd24710a32fe4574c0c57/include/v8-value-serializer.h#L78-L90
```c++
    /**
     * The embedder overrides this method to enable custom host object filter
     * with Delegate::IsHostObject.
     *
     * This method is called at most once per serializer.
     */
    virtual bool HasCustomHostObject(Isolate* isolate);


    /**
     * The embedder overrides this method to determine if an JS object is a
     * host object and needs to be serialized by the host.
     */
    virtual Maybe<bool> IsHostObject(Isolate* isolate, Local<Object> object);
```

lrowe: Following the suggestion on Discord I've updated to pass `&mut Isolate` to `has_custom_host_object`.

legendecas: It was intentionally constrained to accessing `Isolate` only, indicating that the method was not expected to call into JavaScript, so that the embedder should return a constant value (or a value read from `Isolate` data). This method is called on every serialization to determine if the custom host object serialization should be enabled, so it is expected to be free from costly operations.

lrowe: @legendecas does the change in [82b38ed](https://github.com/denoland/rusty_v8/pull/1322/commits/82b38ed65ea3a5f8f094555c7c66ca137f586d64) resolve this satisfactorily?

legendecas: Yeah, thanks, it looks great!

lrowe:
With the change to has_custom_host_object I think this is ready for merging. @legendecas could you approve your review unless you have other concerns?

mmastrac (approved):
LGTM. @bartlomieju can you also take a look?

bartlomieju:
Yup, I'll review tonight.

bartlomieju (approved):
LGTM too, no comments. Thank you @lrowe!

PR #1324, opened 2023-09-17
fix(Symbol): deprecate `for_global` in favour of `for_key` and `for_api`

### LINKED ISSUE #1323: Symbol::for_global incorrectly using ForApi, returns symbols from separate registry
v8 provides two methods on Symbol for returning global symbols, one of which uses a separate registry, see: https://v8.github.io/api/head/classv8_1_1Symbol.html:

> #### For()
> `static Local<Symbol> v8::Symbol::For(Isolate * isolate, Local< String > description)`
> Access global symbol registry. Note that symbols created this way are never collected, so they should only be used for statically fixed properties. Also, there is only one global name space for the descriptions used as keys. To minimize the potential for clashes, use qualified names as keys.
>
> #### ForApi()
> `static Local<Symbol> v8::Symbol::ForApi(Isolate * isolate, Local< String > description)`
> Retrieve a global symbol. Similar to |For|, but using a separate registry that is not accessible by (and cannot clash with) JavaScript code.

In rusty_v8 for_global is described as being equivalent to For(), see: https://docs.rs/v8/0.76.0/v8/struct.Symbol.html#method.for_global

> #### for_global()
> `pub fn for_global<'s>(scope: &mut HandleScope<'s, ()>, description: Local<String>) -> Local<'s, Symbol>`
> Access global symbol registry. Note that symbols created this way are never collected, so they should only be used for statically fixed properties. Also, there is only one global description space for the descriptions used as keys. To minimize the potential for clashes, use qualified descriptions as keys. Corresponds to v8::Symbol::For() in C++.

But in the code we can see it uses `v8__Symbol__ForApi`:
https://github.com/denoland/rusty_v8/blob/0f0697039ebfa6c1e8fc2a6db5d7d76063d658fd/src/symbol.rs#L60-L71

I imagine rusty_v8 should expose both APIs. I think it would make sense to expose `for_api` with the current implementation and fix `for_global` to match its description and call `v8__Symbol__For`.

A GitHub search of the denoland org does not show any current use of `for_global` other than the rusty_v8 tests.

lrowe:
`for_global` was documented as `for_key` but implemented as `for_api`.

Closes #1323

### DIFF
--- a/src/symbol.rs
+++ b/src/symbol.rs
@@ -10,6 +10,10 @@ extern "C" {
     isolate: *mut Isolate,
     description: *const String,
   ) -> *const Symbol;
+  fn v8__Symbol__For(
+    isolate: *mut Isolate,
+    description: *const String,
+  ) -> *const Symbol;
   fn v8__Symbol__ForApi(
     isolate: *mut Isolate,
     description: *const String,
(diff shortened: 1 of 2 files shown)


mmastrac (approved):
LGTM

PR #1326, opened 2023-09-20
fix: never create a null slice

mmastrac:
V8 may pass null pointers to fast API targets when lengths are zero. These are invalid to pass to Rust's slice methods:

> data must be non-null and aligned even for zero-length slices. One reason for this is that enum layout optimizations may rely on references (including slices of any length) being aligned and non-null to distinguish them from other data. You can obtain a pointer that is usable as data for zero-length slices using [NonNull::dangling()](https://doc.rust-lang.org/std/ptr/struct.NonNull.html#method.dangling).

### DIFF
--- a/src/fast_api.rs
+++ b/src/fast_api.rs
@@ -228,6 +228,12 @@ pub struct FastApiOneByteString {
 impl FastApiOneByteString {
   #[inline(always)]
   pub fn as_bytes(&self) -> &[u8] {
+    // Ensure that we never create a null-ptr slice (even a zero-length null-ptr slice
+    // is invalid because of Rust's niche packing).
+    if self.data.is_null() {
+      return &mut [];
+    }
+
     // SAFETY: The data is guaranteed to be valid for the length of the string.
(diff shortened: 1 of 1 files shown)


bartlomieju:
Is there a way to create a test to validate this behavior?

mmastrac:
@bartlomieju I put the test here: https://github.com/denoland/deno_core/pull/220

src/fast_api.rs
```diff
@@ -236,14 +242,37 @@ impl FastApiOneByteString {
...
+  #[inline(always)]
+  pub unsafe fn get_storage_from_pointer_if_aligned<'a>(
+    ptr: *mut Self,
+  ) -> Option<&'a mut [T]> {
+    debug_assert!(!ptr.is_null());
+    let self_ref = ptr.as_mut().unwrap_unchecked();
+    self_ref.get_storage_if_aligned()
+  }
+
+  /// Returns a slace pointing to the underlying data if safe to do so.
```
aapoalas, line 267: **nitpick:** "slice" instead of "slave"

PR #1334, opened 2023-09-26
feat: new_backing_store_from_bytes and empty for ArrayBuffer and SharedArrayBuffer

mmastrac:
Allow creation of `v8::ArrayBuffer` and `v8::SharedArrayBuffer` from things that `DerefMut` to `&mut [u8]`. We stash the underlying object in a `Box` and then keep its slice around in the form of a buffer until we destroy the buffer. There's no way to take the object out of the buffer for now.

This is useful for when we want to return a `BytesMut` to v8.

We also expose an `empty` method for both buffers so we can avoid the overhead when creating empty buffers of any type.

### DIFF
--- a/src/array_buffer.rs
+++ b/src/array_buffer.rs
@@ -61,9 +61,7 @@ extern "C" {
     deleter: BackingStoreDeleterCallback,
     deleter_data: *mut c_void,
   ) -> *mut BackingStore;
-  fn v8__BackingStore__EmptyBackingStore(
-    shared: bool,
-  ) -> *mut BackingStore;
+  fn v8__BackingStore__EmptyBackingStore(shared: bool) -> *mut BackingStore;
 
   fn v8__BackingStore__Data(this: *const BackingStore) -> *mut c_void;
   fn v8__BackingStore__ByteLength(this: *const BackingStore) -> usize;
@@ -404,7 +402,9 @@ impl ArrayBuffer {
   #[inline(always)]
   pub fn empty<'s>(scope: &mut HandleScope<'s>) -> Local<'s, ArrayBuffer> {
     // SAFETY: This is a v8-provided empty backing store
-    let backing_store = unsafe { UniqueRef::from_raw(v8__BackingStore__EmptyBackingStore(false)) };
+    let backing_store = unsafe {
+      UniqueRef::from_raw(v8__BackingStore__EmptyBackingStore(false))
+    };
     Self::with_backing_store(scope, &backing_store.make_shared())
   }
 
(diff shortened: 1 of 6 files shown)


src/array_buffer.rs
```diff
@@ -526,6 +538,53 @@ impl ArrayBuffer {
     }
   }
 
+  /// Returns a new standalone BackingStore backed by an object that dereferences
+  /// to a mutable slice of bytes. The object is dereferenced once, and the resulting slice's
```
piscisaureus, line 542: I don't understand this. It seems that this does the following:
* Copy the object that holds the bytes to the heap and then:
* Wrap it into a BackingStore.
If you want to do that, why don't you just create a BackingStore from the underlying bytes?

piscisaureus:
I would also like to float this suggestion from years ago: https://github.com/denoland/rusty_v8/pull/926#issuecomment-1088729514

src/array_buffer.rs
piscisaureus: I think it might make more sense to have a generic implementation for any type that can create a `BackingStore` for anything that implements `Into<Box<[u8]>>` which covers (I believe) `Vec<u8>`, `String`, `[u8]`. That's more useful because those `Into` implementations generally try to avoid copying (definitely for `Vec`s and `String`s).

mmastrac: In this case we're boxing something (assuming that it allocates bytes somewhere else on the heap) and then asking it to deref from there. The challenge is with BytesMut, common in the tokio ecosystem, which doesn't have `Into<Box<[u8]>>`. It _does_ have conversion to Vec<u8> so maybe there's another option.

To be honest I'm rethinking this PR and I might tweak it a bit if I can make it work more generically.

aapoalas:
> I would also like to float this suggestion from years ago: [#926 (comment)](https://github.com/denoland/rusty_v8/pull/926#issuecomment-1088729514)

Related: I was working on a V8 bugfix yesterday related to AB pointers and stuff. I happened upon a piece of code that suggests that creating AB backing stores from external pointers is not allowed when pointer sandboxing is on: Only pointers created by the AB allocator seem to be okay.

I expect the reason being that internally the AB object caches the backing store pointer which then means that the pointer is freely readable on the JS heap and thus not sandboxed. Or so it seemed at least.

src/array_buffer.rs
```diff
@@ -246,25 +246,50 @@ pub type BackingStoreDeleterCallback = unsafe extern "C" fn(
...
+
+      unsafe fn drop_raw(ptr: *const (), len: usize) {
+        _ = Self::from_raw(ptr::slice_from_raw_parts_mut(ptr as _, len));
+      }
+    }
+  };
+}
+
+// Implement Rawable for single-ownership container types
+rawable!(Box);
```
mmastrac, line 291: Left the macro as this might make sense to implement for `Rc<[u8]>` and `Arc<[u8]>` if they are single-ownership at the time.

mmastrac:
@piscisaureus I revamped the API a bit to remove all the special casing -- now we just need something that is effectively `AsMut<[u8]>`, and allows us to store it in the ArrayBuffer's backingstore data pointer.

The new method works with `Box<[u8]>`, and `Vec<u8>`, as well as `Box<T>` where `T: AsMut<[u8]>`. There's no easy way to avoid a `Box` when we're not working with a built-in type, but that's not too terrible.

piscisaureus (approved):

Issue #933, opened 2022-03-29
V8 EmbedderHeapTracer


ewoudje:
A handy util for the embedder, so that you can keep rust data alive if V8 heap references it.

In V8 there is support for the embedder to be able to keep track of references to the outside and inside of V8 heap.
With EmbedderHeapTracer together with Object#GetAlignedPointerFromInternalField you can keep a rust Rc alive,
this would be handy for storing rust data inside a JS object and so that after a function call from JS the Rc can be used in Rust code again.

Altough I am unsure if such util is within the scope of the V8 crate.

But at this point in time I cannot use the V8 crate to configure any EmbedderHeapTracer, so if possible that the relevant function calls could be exposed so that I or others can use this feature, such as mentioned above.

lucacasonato:
Instead of exposing `EmbedderHeapTracer`, we should expose the Oilpan APIs instead (https://v8.dev/blog/oilpan-library). `EmbedderHeapTracer` is on it's way out.

PR #1336, opened 2023-09-28
Add `cppgc` bindings

littledivy:
https://v8.dev/blog/high-performance-cpp-gc

Oilpan in Deno design doc: https://www.notion.so/denolandinc/Oilpan-cppgc-in-Deno-e194f4268e9f4135ba97610ff7d3a949?pvs=4

Oilpan can be used to implement GC'able resources in Deno

Closes https://github.com/denoland/rusty_v8/issues/933

### DIFF
--- a/examples/cppgc-object.rs
+++ b/examples/cppgc-object.rs
@@ -19,6 +19,8 @@ impl Drop for Wrappable {
   }
 }
 
+// Set a custom embedder ID for the garbage collector. cppgc will use this ID to
+// identify the object that it manages.
 const DEFAULT_CPP_GC_EMBEDDER_ID: u16 = 0xde90;
 
 fn main() {
--- a/examples/cppgc-object.rs
+++ b/examples/cppgc-object.rs
@@ -1,0 +1,127 @@
+// Copyright 2019-2021 the Deno authors. All rights reserved. MIT license.
+use std::cell::Cell;
+
+struct Wrappable {
+  id: String,
+  trace_count: Cell<u16>,
+}
+
+impl v8::cppgc::GarbageCollected for Wrappable {
+  fn trace(&self, _visitor: &v8::cppgc::Visitor) {
+    println!("Wrappable::trace() {}", self.id);
+    self.trace_count.set(self.trace_count.get() + 1);
+  }
+}
+
+impl Drop for Wrappable {
+  fn drop(&mut self) {
+    println!("Wrappable::drop() {}", self.id);
+  }
+}
+
+const DEFAULT_CPP_GC_EMBEDDER_ID: u16 = 0xde90;
+
+fn main() {
+  let platform = v8::new_default_platform(0, false).make_shared();
+  v8::V8::set_flags_from_string("--no_freeze_flags_after_init --expose-gc");
+  v8::V8::initialize_platform(platform.clone());
+  v8::V8::initialize();
+
+  v8::cppgc::initalize_process(platform.clone());
+
+  {
+    let isolate = &mut v8::Isolate::new(v8::CreateParams::default());
+
+    // Create a managed heap.
+    let heap = v8::cppgc::Heap::create(
+      platform,
+      v8::cppgc::HeapCreateParams::new(v8::cppgc::WrapperDescriptor::new(
+        0,
+        1,
+        DEFAULT_CPP_GC_EMBEDDER_ID,
+      )),
+    );
+
+    isolate.attach_cpp_heap(&heap);
+
+    let handle_scope = &mut v8::HandleScope::new(isolate);
+    let context = v8::Context::new(handle_scope);
+    let scope = &mut v8::ContextScope::new(handle_scope, context);
+    let global = context.global(scope);
+    {
+      let func = v8::Function::new(
+        scope,
+        |scope: &mut v8::HandleScope,
+         args: v8::FunctionCallbackArguments,
+         mut rv: v8::ReturnValue| {
+          let id = args.get(0).to_rust_string_lossy(scope);
+
+          let templ = v8::ObjectTemplate::new(scope);
+          templ.set_internal_field_count(2);
+
+          let obj = templ.new_instance(scope).unwrap();
+
+          let member = v8::cppgc::make_garbage_collected(
+            scope.get_cpp_heap(),
+            Box::new(Wrappable {
+              trace_count: Cell::new(0),
+              id,
+            }),
+          );
+
+          obj.set_aligned_pointer_in_internal_field(
+            0,
+            &DEFAULT_CPP_GC_EMBEDDER_ID as *const u16 as _,
+          );
+          obj.set_aligned_pointer_in_internal_field(1, member.handle as _);
+
+          rv.set(obj.into());
+        },
+      )
+      .unwrap();
+      let name = v8::String::new(scope, "make_wrap").unwrap();
+      global.set(scope, name.into(), func.into()).unwrap();
+    }
+
+    let source = v8::String::new(
+      scope,
+      r#"
+      make_wrap('gc me pls'); // Inaccessible after scope.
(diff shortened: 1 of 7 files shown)


lucacasonato (commented):
Will need a layer ontop in `deno_core` to make more rust idiomatic I think. Looking good

mmastrac:
First of all, this is awesome!

I wonder if we should keep this interface low-level for now. There's a few `Gc` container types in Rust already and we may want to experiment with different designs until we find something that's ergonomic for what we need.

https://manishearth.github.io/blog/2021/04/05/a-tour-of-safe-tracing-gc-designs-in-rust/

I think the ideal level of interface here _might_ be the ability to create a `Box`ed `T` with the correct C++ vtable that offers the low-level tracing API (unopinionated at this level), and automatically drops its inner object when the GC finds it goes out of scope.

What we'd then do is build a library (maybe in the deno_core project?) to offer a `Gc<T>` type based on the low-level interface.

src/cppgc.rs
```diff
@@ -0,0 +1,140 @@
...
+    unsafe {
+      cppgc__heap__enable_detached_garbage_collections_for_testing(
+        self as *const Heap as *mut _,
+      );
+    }
+  }
+}
+
+pub trait GarbageCollected {
+  fn trace(&self, visitor: *mut Visitor) {}
```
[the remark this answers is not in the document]
littledivy: Done

src/binding.cc
```diff
@@ -3571,3 +3574,68 @@ void v8__PropertyDescriptor__set_configurable(v8::PropertyDescriptor* self,
...
+}
+
+void cppgc__shutdown_process() {
+  cppgc::ShutdownProcess();
+}
+
+v8::CppHeap* cppgc__heap__create(v8::Platform* platform) {
+  std::unique_ptr<v8::CppHeap> heap = v8::CppHeap::Create(platform, v8::CppHeapCreateParams {
+    {},
+    v8::WrapperDescriptor(InternalFields::kEmbedderType, InternalFields::kSlot, kDefaultCppGCEmebdderID),
```
littledivy, line 3618: Todo: expose `CppHeapCreateParams`

src/binding.cc
```diff
@@ -3571,3 +3574,73 @@ void v8__PropertyDescriptor__set_configurable(v8::PropertyDescriptor* self,
...
+  public:
+    explicit RustObj(void* obj, RustTraceFn trace, RustDestroyFn destroy): trace_(trace), destroy_(destroy), obj_(obj) {}
+
+    ~RustObj() {
+      destroy_(obj_);
+    }
+
+    void Trace(cppgc::Visitor* visitor) const {
+      trace_(visitor, obj_);
+    }
```
mmastrac, line 3595: I'd recommend swapping the order of these parameters as this may allow us to inline this function into a vtable directly.

src/cppgc.rs
```diff
@@ -0,0 +1,261 @@
...
+
+impl<T: GarbageCollected> std::ops::Deref for Member<T> {
+  type Target = T;
+
+  fn deref(&self) -> &Self::Target {
+    unsafe { &*self.ptr }
+  }
+}
+
+/// Constructs an instance of T, which is a garbage collected type.
```
mmastrac, line 233: Let's describe the lifetime of the Box<T> -- ie, when the memory is deallocated, how it may be kept alive, etc.

src/cppgc.rs
```diff
@@ -0,0 +1,261 @@
...
+pub fn make_garbage_collected<T: GarbageCollected>(
+  heap: &Heap,
+  obj: Box<T>,
```
mmastrac, line 236: I've been thinking about this more and I think we actually want to make this a `*mut T` instead of a Box, and pass a custom destruction function. 

My rationale for this is that I think we can build an `Rc`-like type on top of this, but v8 dropping the last reference should still leave copies alive on the Rust side.

So ideally it's `obj: *mut T` and `drop_fn: fn(*mut T)`, where `drop_fn` can be `Box::from_raw` if it's `Box<U>`.

littledivy: That would make this `unsafe fn`? The reason for this API is to prevent a common pitfall where you can simply pass a pointer to a stack allocation.

littledivy: I think we can expose another function `make_garbage_collected_raw`?

littledivy: Added `make_garbage_collected_raw`

src/cppgc.rs
```diff
@@ -0,0 +1,283 @@
...
+  ptr: *mut T,
+}
+
+impl<T: GarbageCollected> std::ops::Deref for Member<T> {
+  type Target = T;
+
+  fn deref(&self) -> &Self::Target {
+    unsafe { &*self.ptr }
+  }
+}
```
littledivy, line 233: Self-note: accessing object ptr should not be a safe interface.

littledivy: Add `unsafe fn get(&self) -> &T`

examples/cppgc-object.rs
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+
+impl Drop for Wrappable {
+  fn drop(&mut self) {
+    println!("Wrappable::drop() {}", self.id);
+  }
+}
+
+const DEFAULT_CPP_GC_EMBEDDER_ID: u16 = 0xde90;
```
aapoalas, line 22: **suggestion:** Maybe add a comment here about where this comes from or how it's decided.

src/cppgc.rs
```diff
@@ -0,0 +1,294 @@
...
+  fn cppgc__heap__create(
+    platform: *mut Platform,
+    wrappable_type_index: int,
+    wrappable_instance_index: int,
+    embedder_id_for_garbage_collected: u16,
+  ) -> *mut Heap;
+  fn cppgc__heap__DELETE(heap: *mut Heap);
+  fn cppgc__make_garbage_collectable(
+    heap: *mut Heap,
+    obj: *mut (),
```
aapoalas, line 22: **question:**  Would this work and/or make sense with a type parameter `T` and `obj: *mut T` as well as return value (presumably) `*mut T`? Same for `TraceFn<T>` and `DestroyFn<T>`.

src/cppgc.rs
```diff
@@ -0,0 +1,294 @@
...
+  fn cppgc__visitor__trace(visitor: *const Visitor, member: *const ());
+}
+
+/// Process-global initialization of the garbage collector. Must be called before
+/// creating a Heap.
+///
+/// Can be called multiple times when paired with `ShutdownProcess()`.
+pub fn initalize_process(platform: SharedRef<Platform>) {
+  unsafe {
+    cppgc__initialize_process(&*platform as *const Platform as *mut _);
```
aapoalas, line 44: **nitpick:** `&mut *platform as *mut Platform` perhaps? Or does `SharedRef` possibly have an `as_ptr` or `as_mut_ptr` API?

src/cppgc.rs
littledivy: foreign types cannot have type parameters.
https://play.rust-lang.org/?version=stable&mode=release&edition=2021&gist=75dc0a79442c4eaa9947b8f5c5fdbbf6

src/cppgc.rs
littledivy: SharedRef does not deref as mutable. This is a common pattern for passing pointers where Rust owernship does not matter:
https://github.com/denoland/rusty_v8/blob/12dca0cf0379f69a52bd710a4a64bf5efe9bcdc0/src/V8.rs#L188

mmastrac: Unrelated to this PR, but we should probably have `as_ptr` and `as_mut_ptr` convenience methods on SharedRef to avoid the need to deref/ref and cast to pointer.

mmastrac (approved):
LGTM!

PR #1350, opened 2023-10-23
feat: add call_with_context to v8::Function

mmastrac:
Allows calling of a function without creation of a `HandleScope<'s>` (that is, from an `Isolate` -- `HandleScope<'s, ()>`.

This shows a small improvement in some deno_core experiments.


bartlomieju (approved):
LGTM

PR #1351, opened 2023-10-25
Many changes to the build system

33KK:
- Download dependencies directly using the DEPS file
- Don't include any v8 source code in the crate
- Fix builds for android
- Move away from relying on secondary_tree to appending rusty_v8 to BUILD.gn

Still TODO is the CI, prebuilts and some other things

~~Also doesn't use the libicu patch, not sure what it even is for~~ right, crate size concerns, shouldn't be an issue anymore

~~clangd may or may not work~~ `ninja -C target/release/build/v8-****/out/out -t compdb > compile_commands.json`

Not tested yet: macos (can't get a vm working), android with external ndk clang, building with deno
Not working yet: windoze with msvc (fails on libicu, EDIT: this actually also fails with upstream rusty_v8)


bartlomieju:
Thanks for the PR. Could you explain the reason for it? Changing the whole build and CI system is a very invasive change and it would be a lot better to open an issue and discuss it first. The current setup is a result of 4 years of maintenance of this crate and closely follows how V8 is built in Chromium project. I'm wary that we can change it all in one go.

ry:
@33KK This is a big change and we would need to consider it carefully - that said, the build system could definitely be improved. We have been attempting to avoid gclient and instead include all source code in the crate itself, so that no external dependencies needed to be downloaded during the build. I'm not sure how useful that design has been for us. I've often thought that we should refactor rusty_v8 to just depend on gclient and depot_tools and download during the build. It seems you're trying something in between our very manual setup and just depending on gclient. Why not just use gclient? I would very much like to avoid maintaining scripts to parse DEPS.

ry:
Thanks for the detailed explanation @33KK.

While I can appreciate the lightweight nature of your proposed solution, it does bring in additional points of potential failure and maintenance. Even if the setup is trivial now, we would still be, for example, tethered to the format of the DEPS file. Any changes to this would require us to be vigilant and reactive.

Currently, our manual setup has its quirks, but it's consistent, known, and doesn't introduce dependencies on additional scripts. Our decision to include all source code in the crate, though tedious, was deliberate to avoid external build-time dependencies.

I am not against moving to a more automated approach. But, in my opinion, if we decide to step away from our manual system, we might as well lean into the established infrastructure of gclient. A halfway solution, as you've suggested, may not provide the clarity and simplicity we're aiming for in the long run.

I'm genuinely open to reevaluating our approach, but I believe it's crucial to ensure our chosen path is sustainable, clear, and beneficial in the bigger picture.

bartlomieju:
Closing this after far too long open — thanks for the work that went into it, and for the detailed discussion at the time.

Two reasons:

The repo has moved a long way since October 2023 — there are 20 submodules in `.gitmodules` now (fp16, libc++/libc++abi/libunwind, fast_float, llvm-libc and others all postdate this PR), and `build.rs` and the CI workflows have been rewritten several times over. Rebasing this wouldn't be a rebase so much as redoing it from scratch against a different codebase.

More importantly, the central idea here did land, in a lighter form. `tools/v8_deps.py` `exec()`s V8's `DEPS` file and reads `deps`/`vars` straight out of it, and `tools/update_deps.py` drives submodule updates from that — so DEPS is the source of truth for revisions without us also maintaining a gclient/cipd reimplementation and its hook and CIPD semantics. That matches the concern @ry raised in this thread about being tethered to formats and behaviour we'd have to track reactively. The Android breakage that originally motivated you has also been fixed since.

Appreciate the DEPS-parsing idea in particular — it shaped what we ended up with.

PR #1360, opened 2023-10-31
chore: track v8 12.0

lucacasonato:


aapoalas:
This should include a CL I managed to land a few weeks prior and a fix to a bug I filed some time ago.

The first may require some changes in FFI tests but it's tests that test for the irregularities I was fixing with that CL in the first place. The second should just make node Buffer polyfill faster <3

littledivy (approved):

PR #1370, opened 2023-12-01
perf: optimize ScopeData structure

littledivy:
```
18.6 ns per run 53.7 million ops/sec → new_ # main
16.4 ns per run 61.2 million ops/sec → new_ # patch
```

### DIFF
--- a/src/scope.rs
+++ b/src/scope.rs
@@ -1260,13 +1260,13 @@ pub(crate) mod data {
     previous: Option<NonNull<ScopeData>>,
     next: Option<Box<ScopeData>>,
     // The 'status' field is also always valid (but does change).
-    status: Cell<ScopeStatus>,
+    status: ScopeStatus,
     // The following fields are only valid when this ScopeData object is in use
     // (eiter current or shadowed -- not free).
     context: Cell<Option<NonNull<Context>>>,
     escape_slot: Option<NonNull<Option<raw::EscapeSlot>>>,
     try_catch: Option<NonNull<raw::TryCatch>>,
-    scope_type_specific_data: ScopeTypeSpecificData,
+    scope_type_specific_data: Option<ScopeTypeSpecificData>,
   }
 
   impl ScopeData {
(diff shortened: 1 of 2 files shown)


src/scope.rs
```diff
@@ -1797,10 +1800,13 @@ pub(crate) mod data {
   }
 
   #[derive(Debug, Clone, Copy, Eq, PartialEq)]
+  #[repr(u8)]
   enum ScopeStatus {
     Free,
-    Current { zombie: bool },
-    Shadowed { zombie: bool },
+    CurrentZombie,
```
mmastrac, line 1806: If we `#[cfg(debug_assertions)]` here, the compiler may have more opportunities to optimize out the zombie code in release mode.

src/scope.rs
```diff
@@ -1280,8 +1280,8 @@ pub(crate) mod data {
         .map(NonNull::as_ptr)
         .map(|p| unsafe { &mut *p })
         .unwrap();
-      match self_mut.status.get() {
-        ScopeStatus::Current { .. } => self_mut,
+      match self_mut.status {
+        ScopeStatus::CurrentZombie | ScopeStatus::CurrentNoZombie => self_mut,
         _ => unreachable!(),
       }
```
mmastrac, line 1286: While we're in here, can we convert this match/unreachable pattern to a `debug_assert`?

src/scope.rs
```diff
@@ -1811,7 +1817,6 @@ pub(crate) mod data {
 
   #[derive(Debug)]
   enum ScopeTypeSpecificData {
-    None,
```
mmastrac, line 1814: Does Rust use the niche with Option<ScopeTypeSpecificData>? Is it the same size as ScopeTypeSpecificData?

mmastrac:
Testing w/ASAN and opt-level=1. It's possible this is good to go and was just miscompiling.

src/scope.rs
```diff
@@ -1295,13 +1295,13 @@ pub(crate) mod data {
     previous: Option<NonNull<ScopeData>>,
     next: Option<Box<ScopeData>>,
     // The 'status' field is also always valid (but does change).
-    status: Cell<ScopeStatus>,
```
piscisaureus, line 1298: Are you sure removing this Cell is safe?

PR #1379, opened 2023-12-22
Fix misleading BigInt rustdoc

guillaumebort:
The original documentation was unclear about the value of the bool. I interpreted it as being `true` if the value was truncated.

### DIFF
--- a/src/bigint.rs
+++ b/src/bigint.rs
@@ -79,9 +79,9 @@ impl BigInt {
   }
 
   /// Returns the value of this BigInt as an unsigned 64-bit integer, and a
-  /// `bool` indicating whether the return value was truncated was truncated or
-  /// wrapped around. In particular, it will be `false` if this BigInt is
-  /// negative.
+  /// `bool` indicating whether the conversion was lossless or not.
+  /// The boolean value will be `false` if the return value was truncated or wrapped around,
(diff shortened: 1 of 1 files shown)


src/bigint.rs
```diff
@@ -79,9 +79,9 @@ impl BigInt {
   }
 
   /// Returns the value of this BigInt as an unsigned 64-bit integer, and a
-  /// `bool` indicating whether the return value was truncated was truncated or
-  /// wrapped around. In particular, it will be `false` if this BigInt is
-  /// negative.
+  /// `bool` indicating whether the conversion was lossless or not.
```
mmastrac, line 82:
```suggestion
  /// `bool` indicating if the conversion was lossless.
```

src/bigint.rs
```diff
@@ -91,7 +91,8 @@ impl BigInt {
   }
 
   /// Returns the value of this BigInt as a signed 64-bit integer, and a `bool`
-  /// indicating whether this BigInt was truncated or not.
+  /// indicating whether the conversion was lossless or not.
```
mmastrac, line 94:
```suggestion
  /// indicating that the conversion was lossless when `true`.
```

mmastrac (approved):
LGTM, thanks!

PR #1392, opened 2024-02-03
Fix busted doc comment

notriddle:
This doc comment contains raw HTML, when it's supposed to contain generics. Fix this by wrapping them as code.

### DIFF
--- a/src/handle.rs
+++ b/src/handle.rs
@@ -63,7 +63,7 @@ extern "C" {
 ///
 /// It is safe to extract the object stored in the handle by
 /// dereferencing the handle (for instance, to extract the `*Object` from
-/// a Local`<Object>`); the value will still be governed by a handle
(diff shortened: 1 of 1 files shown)


src/handle.rs
```diff
@@ -62,14 +62,14 @@ extern "C" {
 /// longer used.
 ///
 /// It is safe to extract the object stored in the handle by
-/// dereferencing the handle (for instance, to extract the *Object from
-/// a Local<Object>); the value will still be governed by a handle
+/// dereferencing the handle (for instance, to extract the `*Object` from
+/// a Local`<Object>`); the value will still be governed by a handle
```
mmastrac, line 66:
```suggestion
/// a `Local<Object>`); the value will still be governed by a handle
```

mmastrac (approved):
LGTM, thanks!

PR #1407, opened 2024-02-29
musl libc builds (docker)

littledivy:


gongzhengyang:
Any update?

nitsky:
> Any update?

From scanning the log, it appears the only problem is a permissions issue in cargo's target directory.

bartlomieju:
Superseded — musl builds are supported now, via a different approach.

#2016 (x86_64-unknown-linux-musl) and #2018 (aarch64-unknown-linux-musl) are both merged. CI has real matrix entries for both targets with prebuilt artifacts, and assembles a musl sysroot for V8's target clang.

The landed approach cross-compiles on the glibc runner — host tools and the snapshot toolchain stay glibc, and only the target objects are musl. That avoids the problems this PR was working around: no Docker in CI, no Alpine-only toolchain, no building gn from source in a container, no `docker cp` out of a stopped container, and no root-owned target directory (the permissions issue @nitsky spotted in the log).

Thanks for the original push on this — closing as obsolete.

PR #1408, opened 2024-03-01
chore: bump rust-toolchain

mmastrac:


src/function.rs
```diff
@@ -957,7 +957,6 @@ impl Function {
   pub fn create_code_cache(&self) -> Option<UniqueRef<CachedData<'static>>> {
     let code_cache =
       unsafe { UniqueRef::try_from_raw(v8__Function__CreateCodeCache(self)) };
-    #[cfg(debug_assertions)]
```
mmastrac, line 960: Clippy lint -- the if block is debug_assert_eq so we can just always run it.

src/unbound_script.rs
```diff
@@ -34,7 +34,6 @@ impl UnboundScript {
     let code_cache = unsafe {
       UniqueRef::try_from_raw(v8__UnboundScript__CreateCodeCache(self))
     };
-    #[cfg(debug_assertions)]
```
mmastrac, line 37: Clippy lint -- the if block is debug_assert_eq so we can just always run it.

bartlomieju (approved):
LGTM

PR #1409, opened 2024-03-01
feat: Add string access methods to OneByteConst

mmastrac:


aapoalas (approved):
Two unnecessary `as` expressions caught by clippy but looks very nice and good to me!

I'm happy (and lightly horrified) to see `OneByteConst` to take on such a big role in snapshots etc :)

bartlomieju (approved):

PR #1412, opened 2024-03-05
chore: Add a test for context create with object template

bartlomieju:
After latests V8 upgrade, the definer is not called anymore.


bartlomieju:
Included in https://github.com/denoland/rusty_v8/pull/1419

PR #1413, opened 2024-03-07
feat: expose CachedData::rejected

igorzi:


src/script_compiler.rs
```diff
@@ -244,12 +249,23 @@ pub fn compile_module2<'s>(
   mut source: Source,
   options: CompileOptions,
   no_cache_reason: NoCacheReason,
+) -> Option<Local<'s, Module>> {
+  compile_module3(scope, &mut source, options, no_cache_reason)
+}
+
+/// Same as compile_module2, but source is passed as a mutable reference.
+#[inline(always)]
+pub fn compile_module3<'s>(
```
bartlomieju, line 258: I think we should unify all of these and make all options required. I can look into that.

igorzi: done

bartlomieju (approved):

PR #1415, opened 2024-03-08
v0.85.0

littledivy:


littledivy (commented):
Oh do not merge yet, I figured out the cause for the object interceptor in V8.

PR #1418, opened 2024-03-08
V8 builtins PGO

### LINKED ISSUE #1416: Enable V8 builtins PGO optimization
https://chromium-review.googlesource.com/c/v8/v8/+/4335220

littledivy:
Closes https://github.com/denoland/rusty_v8/issues/1416

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -28,3 +28,6 @@
 [submodule "third_party/abseil-cpp"]
 	path = third_party/abseil-cpp
 	url = https://chromium.googlesource.com/chromium/src/third_party/abseil-cpp.git
+[submodule "depot_tools"]
+	path = depot_tools
+	url = https://github.com/denoland/depot_tools
--- a/.gn
+++ b/.gn
@@ -72,6 +72,8 @@ default_args = {
 
   # Enable V8 object print for debugging.
   # v8_enable_object_print = true
+  
+  v8_enable_builtins_optimization = true
 
   # Enable Deno-specific extra bindings
   deno_enable_extras = true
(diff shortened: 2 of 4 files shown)


bartlomieju:
Closing this one — the feature is still worth having, so #1416 stays open, but this implementation isn't the shape we want to land.

The main issue is the `depot_tools` submodule. `download_profiles.py` needs it only to get at `download_from_google_storage.py`/gsutil, and it drags in the vpython3 requirement that the `PATH` munging here is working around — that's exactly the dependency the project has spent years avoiding.

A few other things that would need fixing:

- Windows is silently a no-op: the `Command` call is behind `#[cfg(not(target_os = "windows"))]` but `need_pgo_data_download()` isn't, and `.gn` sets `v8_enable_builtins_optimization` unconditionally — so GN would assert on a missing profile there.
- The `PATH` update is Unix-only (`:` separator) and mutates the env for the rest of the build.
- `need_pgo_data_download()` only checks whether `x64.profile` exists, not whether it matches the current V8 revision, so a profile would go stale across a V8 roll. V8 hashes builtins graphs against the profile, so a mismatch hurts rather than helps.
- V8 only selects a profile when pointer compression + external code space + sandbox are all on, and uses a `-rl` suffix for non-clang/Windows hosts (see `v8/BUILD.gn`). Configs we build — jitless iOS, no-pointer-compression, `v8_lite_mode` — fall outside that, and x64.profile isn't the only file involved.

For whoever picks up #1416: fetching the profiles through the existing DEPS-driven tooling in `tools/` and gating the GN flag on the configs V8 actually supports is probably the way in.

Issue #1421, opened 2024-03-12
Add v8::MicrotaskQueue bindings


bartlomieju:
We need bindings to [`v8::MicrotaskQueue`](https://github.com/denoland/v8/blob/db2b8439ec4d3ae972330cc2e9830ccd517d7235/include/v8-microtask-queue.h#L40). 

These are required to polyfill Node's `vm` module: https://nodejs.org/api/vm.html#scriptruninnewcontextcontextobject-options

PR #1423, opened 2024-03-12
Add `v8::MicrotaskQueue` bindings

littledivy:
Closes https://github.com/denoland/rusty_v8/issues/1421


src/microtask.rs
```diff
@@ -0,0 +1,76 @@
+use crate::support::int;
+use crate::support::Opaque;
```
bartlomieju, line 2: Remember about copyright

bartlomieju (approved):

PR #1424, opened 2024-03-12
Add `Isolate::add_message_listener_with_error_level`

### LINKED ISSUE #1422: `v8::MessageCallback` registered but not called
the handler is registered and default output supressed but it's never called by V8.

```rust
extern "C" fn message_handler(
  msg: v8::Local<v8::Message>,
  _: v8::Local<v8::Value>,
) {
  let scope = &mut unsafe { v8::CallbackScope::new(msg) };
  let msg_str = msg.get(scope).to_rust_string_lossy(scope);
  println!("{}", msg_str);
}

isolate.add_message_listener(message_handler);
```
https://source.chromium.org/chromium/chromium/src/+/main:v8/src/execution/messages.cc;l=168-187;drc=0cfb1b06652d6572895919e6564563a6761ca6fd

littledivy:
Fixes #1422

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -278,7 +278,7 @@ bool v8__Isolate__AddMessageListener(v8::Isolate* isolate,
 
 bool v8__Isolate__AddMessageListenerWithErrorLevel(
     v8::Isolate* isolate, v8::MessageCallback callback, int error_level) {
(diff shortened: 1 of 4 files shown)


bartlomieju:
Can you add a test to `test_api.rs`?

bartlomieju (approved):

PR #1426, opened 2024-03-12
feat: Add IndexedPropertyHandlerConfiguration::*_raw methods

bartlomieju:
Matches implementation of NamedPropertyHandlerConfiguration.

This is necessary to polyfill `node:vm` module in Deno.


littledivy (approved):
LGTM

PR #1433, opened 2024-03-15
feat: use `.cargo/.rusty_v8/<file>` for caching downloads manually

mmastrac:


Cargo.toml
```diff
@@ -86,6 +86,7 @@ once_cell = "1.19"
 [build-dependencies]
 fslock = "0.2"
 which = "5"
+home = "0"
```
bartlomieju, line 89: Wow, that's a huge dependency, good that it's build only

bartlomieju (approved):

Cargo.toml
mmastrac: I ran cargo update to fix the lockfile -- no more extra deps.

PR #1434, opened 2024-03-15
feat: .gz compress release artifacts

mmastrac:
This helps download artifacts on flaky or slow networks. We save about 75% of the download required for most of these archives.

We'll need to do a release after we land this one to avoid a case where main can't be built.


bartlomieju (approved):
That should speed up CI and `cargo build` after bumping `rusty_v8` in deno_core 👍

PR #1442, opened 2024-03-17
Refactored OwnedIsolate construction code in prep for lockers

Earthmark:
Pulled the isolate construction changes from the locker API to a separate PR.

This splits annex and scope construction, and disposal, allowing them to be done separately.

This is the pre-work for https://github.com/denoland/rusty_v8/pull/1411

### DIFF
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -585,6 +585,20 @@ impl Isolate {
     );
   }
 
+  fn new_impl(params: CreateParams) -> *mut Isolate {
+    crate::V8::assert_initialized();
+    let (raw_create_params, create_param_allocations) = params.finalize();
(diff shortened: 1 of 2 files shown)


src/isolate.rs
```diff
@@ -1560,9 +1579,35 @@ impl Drop for OwnedIsolate {
...
-      self.cxx_isolate.as_mut().clear_scope_and_annex();
-      self.cxx_isolate.as_mut().dispose();
+      self.dispose_scope_root();
+      self.dispose_annex();
+      self.dispose();
+    }
+  }
+}
+
+impl OwnedIsolate {
```
Earthmark, line 1589: this moved up so destroying an isolate to create a blob was near where isolates are normally destroyed, I can move it back down if wanted though.

mmastrac (approved):

PR #1443, opened 2024-03-18
Add `v8::Context::set_embedder_data`

littledivy:
Needed for https://github.com/denoland/deno/pull/22978/

TODO: needs test ik


bartlomieju:
Already implemented — closing as obsolete.

`Context::set_embedder_data` is in `src/context.rs` today, along with a matching `get_embedder_data`. The version that landed also offsets the caller's slot by `INTERNAL_SLOT_COUNT` so embedder slots can't collide with rusty_v8's internal ones (and with V8's reserved index 0), and initializes the annex before writing — neither of which this patch did.

Thanks for the original bindings.

PR #1449, opened 2024-04-10
chore: Allow Rust ASAN flags to propagate into v8 build

mmastrac:
Runs ASAN on every build as part of the Mac runs. I wasn't able to get this working for Linux :/ 

 - Requires fixing +nightly warnings.
 - One ASAN issue found: fast functions w/an overload were accessing the stack after a scope was destroyed.

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -8692,7 +8692,9 @@ fn icu_date() {
 
 #[test]
 fn icu_set_common_data_fail() {
-  assert!(v8::icu::set_common_data_73(&[1, 2, 3, 0, 0, 0, 0, 0, 0, 0, 0]).is_err());
+  assert!(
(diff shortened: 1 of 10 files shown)


tests/test_api.rs
```diff
@@ -8690,6 +8690,7 @@ fn icu_date() {
   }
 }
 
+#[ignore]
```
bartlomieju, line 8693: Why ignoring this test?

bartlomieju (approved):
Discussed offline, we're gonna try this approach and evaluate after some time.

tests/test_api.rs
mmastrac: Oh, this test is actually causing UB in the C code. It reads from undefined bytes. I don't have a great way to detect ASAN right now.

PR #1466, opened 2024-04-19
Rolling to V8 12.6.228.3

denobot:


mmastrac (approved):

src/binding.cc
```diff
@@ -31,6 +31,8 @@
 #include "v8/src/objects/objects.h"
 #include "v8/src/objects/smi.h"
 
+#pragma clang diagnostic ignored "-Wdeprecated-declarations"
```
bartlomieju, line 34: Note to self, remove when #1478 is addressed

PR #1470, opened 2024-04-22
api improvement

dev-ardi:
This addresses https://github.com/denoland/rusty_v8/issues/1467

This new API design has the following advantages
1. It makes it sound to initialize isolates
2. It uses builder pattern to create a platform. This is better because it doesn't require users to explicitly think about uncommon options, and better documents existing options in case that users might need them.
3. It makes it impossible at compile time to create an isolate without having initialized the platform
4. It's uncommon to create an isolate with options so make the default way of creating them not take any options.
5. This leads to nicer code and it's more rusty:
```rs
let isolate = v8::V8Options::new()
  .build() // This hints that it's builder pattern and that there are more options that could be used, but doesn't forces them to know them.
  .isolate();
  ```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -22,9 +22,9 @@ dependencies = [
 
 [[package]]
 name = "aho-corasick"
-version = "1.1.2"
+version = "1.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


dev-ardi:
The idea is there. Feel free to do with it what you will.

PR #1471, opened 2024-04-22
fix: don't pass refs to DELETE

mmastrac:
Deleting invalidates the target of `&mut`. While the reference is no longer _used_ after this call, it must continue to be valid, making all calls to DELETE with refs unsound.


mmastrac:
@svix-aaron1011 I think this is the problem.

bartlomieju (approved):

Issue #1478, opened 2024-05-14
Use of deprecated APIs in binding.cc


mmastrac:
When moving to 12.6, we bumped into a large number of deprecated APIs.

To work around it, the entire binding file is now ignoring deprecated APIs, but these will need to be bumped to the replacement APIs.

```
  ../../../../src/binding.cc:1225:24: error: 'SetAccessor' is deprecated: Use SetNativeDataProperty instead [-Werror,-Wdeprecated-declarations]
  ../../../../src/binding.cc:1239:35: error: 'NamedPropertyHandlerConfiguration' is deprecated: Provide interceptor callbacks with new signatures instead (NamedPropertyXxxCallback) [-Werror,-Wdeprecated-declarations]
  ../../../../src/binding.cc:1253:35: error: 'IndexedPropertyHandlerConfiguration' is deprecated: Provide interceptor callbacks with new signatures instead (IndexedPropertyXxxCallbackV2) [-Werror,-Wdeprecated-declarations]
  ../../../../src/binding.cc:3739:11: error: 'WrapperDescriptor' is deprecated: WrapperDescriptor is deprecated, see crbug.com/338411141. [-Werror,-Wdeprecated-declarations]
  ../../../../v8/include/v8-cppgc.h:90:3: note: 'CppHeapCreateParams' has been explicitly marked deprecated here
```

PR #1481, opened 2024-05-15
chore: Update use of deprecated APIs

bartlomieju:
This commit updates APIs related to named and indexed property
handlers to not use deprecated V8 APIs. Main change is the change
of return value in callbacks that now requires to return `v8::Intercepted` enum.

Towards https://github.com/denoland/rusty_v8/issues/1478


mmastrac (approved):
LGTM. I hope this fixes the deleter issue.

PR #1483, opened 2024-05-16
Applying modified patches from FreeBSD ports (www/deno/files)

Munrok:
Patches from FreeBSD port tree been modified to fit current state of repository and were applied.
Source of original patch files https://github.com/freebsd/freebsd-ports/tree/main/www/deno/files
Subset of patches used: patch-cargo-crates_v8_*

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -326,6 +326,8 @@ fn host_platform() -> String {
     "mac"
   } else if cfg!(target_os = "windows") {
     "windows"
+  } else if cfg!(target_os = "freebsd") {
+    "freebsd"
   } else {
     "unknown"
   };


Munrok:
I see that some checks were aborted after dozen of seconds. Does it mean additional work should be done in order to prepare version that can go further in checking process?

bartlomieju:
@Munrok the whole build process requires various directories to be pulled from repos like `https://github.com/denoland/chromium_build.git` or `https://github.com/denoland/v8.git`. If you'd like to upstream your changes you'll need to open PR against these repos. That said, I'm not super happy to float several patches in V8 or abseil. Have you tried opening CL against upstream repos?

bartlomieju:
Closing this after too long open — thanks for putting it together.

As noted at the time, the substance of this isn't in this repo: it lives in the `v8`, `build`, `buildtools` and `abseil-cpp` submodules, and would need to go upstream to `denoland/v8` and `denoland/chromium_build` first. The submodule pins here are also from the V8 12.x era and we're on 15.2 now, so they can't be applied as-is.

The one in-repo change — adding `"freebsd"` to `host_platform()` in `build.rs` — is a single line and would be easy to re-propose on its own if it's useful. There's already some FreeBSD handling in `build.rs` for C++ stdlib linking. The bigger obstacle is that we have no FreeBSD CI target, so nothing would keep support working once it landed.

PR #1489, opened 2024-05-29
fix: fix size checking on 32-bit platforms

liushuyu:
This pull request fixes the type size checking on 32-bit platforms in `src/binding.cpp`.

On 32-bit platforms like armv7 or ppc32, the size of `v8::ScriptCompiler::CompilationDetails` and `v8::ScriptCompiler::Source` will be different than currently expected. This is due to how those 32-bit platforms lay out the 64-bit elements in a structure.

You can see a reduced example on Compiler Explorer: https://godbolt.org/z/TG1bPGG8n.


ry (approved):
LGTM

PR #1497, opened 2024-06-10
Fix uninitialized context annex slot

littledivy:
Attempts to `clear_all_slots` would crash when annex slot is not intialized but is also non-null garbage value. This patch makes sure we initialize a valid annex slot.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -322,6 +322,9 @@ impl Context {
     slot: i32,
     data: *mut c_void,
   ) {
+    // Initialize the annex when slot count > INTERNAL_SLOT_COUNT.
+    self.get_annex_mut(&mut *v8__Context__GetIsolate(self), true);
+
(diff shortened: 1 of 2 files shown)


src/context.rs
```diff
@@ -322,6 +322,9 @@ impl Context {
     slot: i32,
     data: *mut c_void,
   ) {
+    // Initialize the annex when slot count > INTERNAL_SLOT_COUNT.
+    self.get_annex_mut(&mut *v8__Context__GetIsolate(self), true);
```
littledivy, line 326: Should we change the signature to accept a `&mut Isolate` or leave it to the caller as part of the unsafe contract? (this function is marked as unsafe anyways)

bartlomieju: I think this signature is fine

bartlomieju (approved):

PR #1500, opened 2024-06-12
chore: don't ignore deprecated declarations

ry:


bartlomieju:
This is being addressed in https://github.com/denoland/rusty_v8/pull/1488 but is blocked by V8 bugs.

PR #1521, opened 2024-07-10
chore: output more information in build

dsherret:
Probably won't fix things, but at least we'll output more information.


bartlomieju (approved):
Want to bump the version as well in this PR so we can test it in Deno?

PR #1532, opened 2024-07-18
fix: ExternalOneByteStringResource is not guaranteed to be valid UTF-8

### LINKED ISSUE #1531: `ExternalOneByteStringResource::as_str` is unsound
The method https://github.com/denoland/rusty_v8/blob/main/src/string.rs#L153-L168 converts to a Rust `&str` without checking the validity of the data because "The data is guaranteed to be ASCII."

But both https://github.com/denoland/rusty_v8/blob/main/src/string.rs#L582-L596 and https://github.com/denoland/rusty_v8/blob/main/src/string.rs#L627-L643 are documented as accepting Latin-1 characters.

A UTF-8 string borrowed from a `ExternalOneByteStringResource` may thus not actually be valid UTF-8.

aapoalas:
A subtle unsoundness / undefined behaviour made its way into the fairly recently added `ExternalOneByteStringResource` object: The `as_str` API is not sound as the data inside may be be Latin-1, not ASCII.

As the API was not used anywhere in deno or deno_core, I opted to simply remove it and replace it with an `as_bytes` API. I also modified the test to showcase the Latin-1 string case and added copious notes and explanations around the code to make sure this doesn't accidentally happen again. The likely reason why the API originally slipped in is because the `OneByteConst` has this API where it _is_ safe because the `OneByteConst` creation checks the data for ASCII-ness.

I also tried to add an API to extract an `Option<&'static OneByteConst>` from an `&ExternalOneByteStringResource` but run into https://github.com/rust-lang/rust/issues/119618 ie. OneByteConst is actually duplicating the vtables... which is not great.

Closes #1531

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -135,12 +135,16 @@ pub struct ExternalStringResourceBase(Opaque);
 #[repr(C)]
 /// An external, one-byte string resource.
 /// This corresponds with `v8::String::ExternalOneByteStringResource`.
+///
+/// Note: The data contained in a one-byte string resource is guaranteed to be
+/// Latin-1 data. It is not safe to assume that it is valid UTF-8, as Latin-1
+/// only has commonality with UTF-8 in the ASCII range and differs beyond that.
 pub struct ExternalOneByteStringResource(Opaque);
 
 impl ExternalOneByteStringResource {
   /// Returns a pointer to the data owned by this resource.
   /// This pointer is valid as long as the resource is alive.
-  /// The data is guaranteed to be ASCII.
+  /// The data is guaranteed to be Latin-1.
   pub fn data(&self) -> *const char {
     unsafe { v8__ExternalOneByteStringResource__data(self) }
   }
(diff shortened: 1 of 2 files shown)


bartlomieju:
Closes https://github.com/denoland/rusty_v8/issues/1531?

aapoalas:
> Closes #1531?

Oh, yes

src/string.rs
```diff
@@ -151,23 +155,15 @@ impl ExternalOneByteStringResource {
...
-      // SAFETY: We know this is ASCII and length > 0
-      unsafe {
-        std::str::from_utf8_unchecked(std::slice::from_raw_parts(
-          self.data().cast(),
-          len,
-        ))
-      }
-    }
+    // SAFETY: We know this is Latin-1
+    unsafe { std::slice::from_raw_parts(self.data().cast(), len) }
```
0f-0b, line 162: Is `self.data()` always non-null?

bartlomieju (approved):
LGTM provided that the answer to https://github.com/denoland/rusty_v8/pull/1532#discussion_r1683850045 is "yes"

src/string.rs
aapoalas: That's actually a good question, and I'm not sure. At least if the length is non-zero then it should always be.

I returned the length check that I had removed to be absolutely sure.

aapoalas:
> LGTM provided that the answer to [#1532 (comment)](https://github.com/denoland/rusty_v8/pull/1532#discussion_r1683850045) is "yes"

I couldn't give a full "yes" so I fixed the code to check for length being zero. If length is zero, then the pointer could also theoretically be null.

PR #1644, opened 2024-10-17
chore(build): do not install sysroot if doing native builds

liushuyu:
This pull request changes the build script logic. If we are building v8 from the source without using cross-compilation, we skip downloading the sysroot files as they will not be used anyway.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -237,43 +237,45 @@ fn build_v8(is_asan: bool) {
       gn_args.push(arg.to_string());
     }
   }
-  // cross-compilation setup
-  if target_arch == "aarch64" {
-    gn_args.push(r#"target_cpu="arm64""#.to_string());
-    gn_args.push("use_sysroot=true".to_string());
-    maybe_install_sysroot("arm64");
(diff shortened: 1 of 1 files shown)


bartlomieju:
Closing this — the problem you identified is real, but the patch has drifted too far from the current `build.rs` to apply.

We still install both the arm64 and amd64 sysroots whenever `target_arch == "aarch64"`, including on a native aarch64 Linux host where neither is needed, so the underlying complaint stands.

Two things stop this specific diff from landing now. `build_v8` has since grown a riscv64 block, a musl cross-compilation block, and the Android logic has moved, so it conflicts throughout. More importantly, the aarch64 path now guards sysroot installation on `target_os == "linux"`, and restructuring around `target_triple != host_triple` drops that guard — cross-compiling to `aarch64-apple-darwin` or iOS would then set `use_sysroot=true` and try to install Debian sysroots.

#1663 is open and covers overlapping ground using a native aarch64 runner. Happy to look at a fresh patch for the native-host sysroot skip if you're still interested.

PR #1649, opened 2024-10-24
Musl cross

RonnyChan-okta:
Rebase of #1406 on main


bartlomieju:
Superseded — musl builds landed via #2016 (x86_64-unknown-linux-musl) and #2018 (aarch64-unknown-linux-musl), and both targets now have CI matrix entries and prebuilt artifacts.

The approach that landed teaches the vendored build config a target-scoped `use_musl` arg and keeps the host tools (torque, mksnapshot, code generators) on glibc, so only the target objects are musl. That avoids bootstrapping a GCC cross-toolchain from source at build time and the extra `musl-cross-make` submodule this PR needed. As @littledivy noted, the patch had also stopped working.

Thanks for the rebase work here even so.

PR #1650, opened 2024-10-24
Musl docker

RonnyChan-okta:


bartlomieju:
Superseded — musl builds landed via #2016 (x86_64-unknown-linux-musl) and #2018 (aarch64-unknown-linux-musl).

This uses the same Docker/Alpine approach as #1407 (identical `musl-build.sh`), which we've also closed. The landed approach cross-compiles on the glibc runner instead, so there's no Docker in CI, no Alpine-only toolchain, no building gn from source in a container, and no root-owned target directory to chown afterwards.

Thanks regardless.

PR #1669, opened 2024-12-30
Add `v8::Eternal<T>`

littledivy:
Eternal handles are set-once handles that live for the lifetime of the isolate.

https://v8.github.io/api/head/classv8_1_1Eternal.html

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -11977,3 +11977,24 @@ fn use_counter_callback() {
 
   assert_eq!(COUNT.load(Ordering::Relaxed), 1);
 }
+
+#[test]
+fn test_eternals() {
+  let _setup_guard = setup::parallel_test();
+  let eternal1 = v8::Eternal::empty();
+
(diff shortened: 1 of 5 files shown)


ry (approved):

tests/test_api.rs
```diff
@@ -11977,3 +11977,24 @@ fn use_counter_callback() {
...
+  assert!(eternal1.is_empty());
+  eternal1.set(&mut scope, str1);
+  assert!(!eternal1.is_empty());
+
+  let str1_get = eternal1.get(&mut scope);
+  assert_eq!(str1, str1_get);
+
+  eternal1.clear();
+  assert!(eternal1.is_empty());
+}
```
bartlomieju, line 12000: Is there a way to test behavior after the isolate is dropped? Ie. it won't segfault when accessed

littledivy: Yep, added.

bartlomieju (approved):

PR #1674, opened 2025-01-02
feat: add bindings for RegExp

crowlKats:


src/regexp.rs
```diff
@@ -0,0 +1,73 @@
+use crate::support::int;
```
littledivy, line 1: Missing copyright header

src/lib.rs
```diff
@@ -138,6 +139,7 @@ pub use property_attribute::*;
 pub use property_descriptor::*;
 pub use property_filter::*;
 pub use property_handler_flags::*;
+pub use regexp::RegExpCreationFlags;
```
littledivy, line 142:
```suggestion
pub use regexp::RegExpCreationFlags;
pub use regexp::Regexp;
```

` regexp::Regexp` needs to be public too?

crowlKats: `RegExp` is public already as it is defined in `data.rs`

src/regexp.rs
crowlKats: none of the other files i looked at have it 🤔

littledivy (approved):
LGTM

src/binding.cc
```diff
@@ -2367,6 +2367,22 @@ bool v8__PropertyCallbackInfo__ShouldThrowOnError(
   return self.ShouldThrowOnError();
 }
 
+const v8::RegExp* v8__RegExp__new(const v8::Context& context,
+                                  const v8::String& pattern, int options) {
```
littledivy, line 2371:
```suggestion
const v8::RegExp* v8__RegExp__new(const v8::Context& context,
                                  const v8::String& pattern, v8::Flags options) {
```

PR #1682, opened 2025-01-08
fix: add back incorrectly spelled initalize_process

dsherret:
We need to add this back until we do a major release.


ry (approved):
RSLGTM

Issue #1716, opened 2025-02-24
Feature request: Expose `Isolate:: SetModifyCodeGenerationFromStringsCallback`


rebu-dt:
See https://v8.github.io/api/head/classv8_1_1Isolate.html#a0e71905d16905c2bf060854a23e2ba75.

This callback is useful for tracking whenever `eval`/the `Function` constructor is called and/or disabling code generation from user-provided strings entirely. Note that there was already [an attempt](https://github.com/denoland/rusty_v8/pull/435) to expose this callback, but the PR was closed due to inactivity.

PR #1726, opened 2025-03-05
Expose `ModifyCodeGenerationFromStrings` callback

rebu-dt:
Implements https://github.com/denoland/rusty_v8/issues/1716. Most of the code was taken from https://github.com/denoland/rusty_v8/pull/435 and adapted to work with the current V8 version (credits for the original PR go to @NeoLegends).

The approach implemented in this PR is somewhat limited (but does match the feature set of the previous PR):
- There's no `HandleScope` available in the callback, so it doesn't seem to be possible to read or modify the source code
- The attribute `#[allow(improper_ctypes_definitions)]` must be added to the callback function for it to not be (as far as I know, incorrectly) flagged with a warning

### DIFF
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -13722,6 +13722,56 @@ fn code_generation_from_strings_callback() {
   let result: Option<v8::Local<'_, v8::Value>> = eval(scope, "eval('1 + 1')");
   assert_eq!(result.unwrap().number_value(scope).unwrap(), 42.0);
   assert_eq!(COUNT.load(Ordering::Relaxed), 3);
+
+  // `modified_source` is only considered when `codegen_allowed` is true, so
+  // blocking still wins even though the callback hands back a replacement.
+  CODEGEN_ALLOWED.store(false, Ordering::Relaxed);
+  {
+    let tc = pin!(v8::TryCatch::new(scope));
+    let tc = &mut tc.init();
+    eval(tc, "eval('1 + 1')");
+    assert_eq!(
+      tc.message().unwrap().get(tc).to_rust_string_lossy(tc),
+      "Uncaught EvalError: Code generation from strings disallowed for this context"
+    );
+    assert_eq!(COUNT.load(Ordering::Relaxed), 4);
+  }
+}
+
+#[test]
+fn code_generation_from_strings_callback_non_string_source() {
+  static COUNT: AtomicUsize = AtomicUsize::new(0);
+
+  fn callback<'s, 'i>(
+    scope: &mut v8::PinScope<'s, 'i>,
+    source: v8::Local<'s, v8::Value>,
+    is_code_like: bool,
+  ) -> v8::ModifyCodeGenerationFromStringsResult<'s> {
+    COUNT.fetch_add(1, Ordering::Relaxed);
+    // `is_code_like` is only true for objects marked via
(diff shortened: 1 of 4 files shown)


tests/test_api.rs
```diff
@@ -12102,6 +12102,50 @@ fn use_counter_callback() {
...
+  #[allow(improper_ctypes_definitions)]
+  extern "C" fn callback<'s>(
+    _context: v8::Local<'s, v8::Context>,
+    _source: v8::Local<'s, v8::Value>,
+    _is_code_like: bool,
+  ) -> v8::ModifyCodeGenerationFromStringsResult<'s> {
+    CALLED.store(true, Ordering::Relaxed);
+    v8::ModifyCodeGenerationFromStringsResult {
+      codegen_allowed: CODEGEN_ALLOWED.load(Ordering::Relaxed),
+      modified_source: None,
```
ry, line 12119: Use `modified_source` to demonstrate the feature

tests/test_api.rs
```diff
@@ -12102,6 +12102,50 @@ fn use_counter_callback() {
   assert_eq!(COUNT.load(Ordering::Relaxed), 1);
 }
 
+#[test]
+fn code_generation_from_strings_callback() {
+  static CODEGEN_ALLOWED: AtomicBool = AtomicBool::new(false);
+  static CALLED: AtomicBool = AtomicBool::new(false);
```
ry, line 12108: make called an integer counter

tests/test_api.rs
```diff
@@ -12102,6 +12102,50 @@ fn use_counter_callback() {
...
+  let scope = &mut v8::ContextScope::new(&mut scope, context);
+
+  // Code generation should be disallowed
+  {
+    let tc = &mut v8::TryCatch::new(scope);
+    eval(tc, "eval('1 + 1')");
+    assert_eq!(
+      tc.message().unwrap().get(tc).to_rust_string_lossy(tc),
+      "Uncaught EvalError: Code generation from strings disallowed for this context"
+    );
```
ry, line 12140: check CALLED

tests/test_api.rs
```diff
@@ -12102,6 +12102,50 @@ fn use_counter_callback() {
...
+  }
+
+  // Enable code generation
+  CODEGEN_ALLOWED.store(true, Ordering::Relaxed);
+  let result: Option<v8::Local<'_, v8::Value>> = eval(scope, "eval('1 + 1')");
+  assert_eq!(result.unwrap().number_value(scope).unwrap(), 2.0);
```
ry, line 12146: check CALLED

tests/test_api.rs
```diff
@@ -12102,6 +12102,50 @@ fn use_counter_callback() {
   assert_eq!(COUNT.load(Ordering::Relaxed), 1);
 }
 
+#[test]
+fn code_generation_from_strings_callback() {
+  static CODEGEN_ALLOWED: AtomicBool = AtomicBool::new(false);
+  static CALLED: AtomicBool = AtomicBool::new(false);
+
+  #[allow(improper_ctypes_definitions)]
```
ry, line 12110: I wonder if we should change how `Local` is defined to avoid this..?

ry (dismissed):
thanks for the patch - test could use some improvement

bartlomieju (approved):
Thanks

PR #1788, opened 2025-05-21
Remove v8_enable_shared_ro_heap=false and v8_enable_verify_heap=false

goffrie:
v8_enable_shared_ro_heap is no longer configurable since https://github.com/v8/v8/commit/b3054f77d65ea96ddf6e0a526abadebd91d7114e.

This also removes the justification for setting `v8_enable_verify_heap=false` so remove that too.

### DIFF
--- a/.gn
+++ b/.gn
@@ -64,17 +64,6 @@ default_args = {
   v8_array_buffer_internal_field_count = 2
   v8_array_buffer_view_internal_field_count = 2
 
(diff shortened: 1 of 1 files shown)


goffrie:
(as expected, this doesn't affect the release build at all, and sccache is even able to reuse all artifacts, so CI ran very quickly for release configurations)

PR #1833, opened 2025-08-13
feat: add enable-i18n-support build for mobile device

zeronight:
Enabling compilation with v8_enable_i18n_support=false can significantly reduce the library size，this is very useful for some mobile devices.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -98,8 +98,9 @@ exclude = [
 opt-level = 1
 
 [features]
-default = ["use_custom_libcxx"]
+default = ["use_custom_libcxx", "v8_enable_i18n_support"]
 use_custom_libcxx = []
+v8_enable_i18n_support = []
 v8_enable_pointer_compression = []
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


bartlomieju:
Thanks for this — closing in favour of #1996, which takes the same idea further.

The main difference is how the feature is modelled. This PR makes i18n a default-on feature, which means any consumer building with `default-features = false` silently loses ICU — and the Deno CLI does exactly that. #1996 inverts it into an opt-out `no_icu` feature, so existing consumers and prebuilt archive names are unaffected and only the slim build opts in.

#1996 also gates the bindings behind an explicit `rusty_v8_enable_i18n` GN arg rather than V8's internal `V8_INTL_SUPPORT` define, and updates `src/icu.rs`, the tests and CI — without which `icu::get_default_locale`/`set_default_locale` would be left as undefined symbols on the Rust side when i18n is off.

Appreciate you raising it; the size win for mobile builds is exactly the motivation there too.

PR #1852, opened 2025-09-16
fix: cargo package build from source

ry:
Ref https://github.com/Homebrew/homebrew-core/pull/236828

Test with: `V8_FROM_SOURCE=1 cargo publish --dry-run -vv`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ exclude = [
   "*.m4",
   "*.md",
   "*.settings",
-  "*.txt",
   "*OWNERS*",
   ".*",
   "AUTHORS",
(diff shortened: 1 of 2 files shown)


bartlomieju:
Closing as stale — nothing here landed, and the tree has moved on. `third_party/rust` is a submodule now and `chromium_crates_io` is present in the checkout, so the fetch-at-build-time path this added no longer matches how the directory gets there.

Worth noting for anyone revisiting the underlying Homebrew packaging issue: cloning Chromium's `third_party/rust` from `build.rs` means network access during the package build, which is usually exactly what distro packagers disallow — so that approach may have been at odds with the motivating use case anyway.

Reopen or file fresh if source packaging is still broken.

PR #1853, opened 2025-09-17
refactor: rework inspector to take shared references and fix unsoundness

nathanwhit:
Previously we were storing pointers to the raw values, which V8 was holding onto. We didn't actually enforce they were pinned. So this was trivially unsound:

```
let channel = v8::inspector::ChannelBase::new::<ChannelThing>();
let mut channel = ChannelThing { channel };
let client_trust_level = v8::inspector::V8InspectorClientTrustLevel::FullyTrusted;

let session = inspector.connect(
    1,
    &mut channel,
    v8::inspector::StringView::empty(),
    client_trust_level,
);

drop(channel);

// any usage of the session that accesses the channel after this is a UAF
```

We also forced you to embed the base value into your implementation, and all the methods were taking &mut self.

Now, we take shared references (which is ok because we either use an UnsafeCell to get a pointer to hand to V8, or we only use raw pointers)

### DIFF
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -424,7 +424,7 @@ pub struct RawV8InspectorClient {
 }
 
 impl V8InspectorClient {
-  pub fn run_message_loop_on_pause(&mut self, context_group_id: i32) {
+  pub fn run_message_loop_on_pause(&self, context_group_id: i32) {
     unsafe {
       v8_inspector__V8InspectorClient__runMessageLoopOnPause(
         self.raw(),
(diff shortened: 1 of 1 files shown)


src/support.rs
```diff
@@ -173,10 +176,6 @@ fn assert_unique_ptr_layout_compatible<U, T>() {
   // as a raw C pointer.
   assert_eq!(size_of::<U>(), size_of::<*mut T>());
   assert_eq!(align_of::<U>(), align_of::<*mut T>());
-
-  // Assert that `T` (probably) implements `Drop`. If it doesn't, a regular
-  // reference should be used instead of UniquePtr/UniqueRef.
-  assert!(needs_drop::<T>());
```
nathanwhit, line 179: Removed because I don't think it makes sense to enforce this, the comment is wrong in that a regular reference is not a safe replacement since it would carry aliasing invariants.

Also in general I prefer only using raw pointers to refer to C++ values, and implementing drop forces you to take a mutable ref

src/support.rs
```diff
@@ -428,15 +427,6 @@ impl<F> FieldOffset<F> {
         .unwrap()
     }
   }
-
-  #[allow(clippy::wrong_self_convention)]
-  pub unsafe fn to_embedder_mut<E>(self, field: &mut F) -> &mut E {
```
nathanwhit, line 433: now unused

PR #1863, opened 2025-10-13
Upgrade V8 to 14.2

ry:


nathanwhit (approved):
Library changes LGTM, not as familiar with the bindgen/build.rs stuff. Maybe @devsnek  should take a look at those

PR #1891, opened 2025-12-17
Store IsolateHandle data in its own allocation

coolreader18:
Currently, `IsolateHandle` is unsound - it mutates `IsolateAnnex.isolate` through an immutable reference with no `UnsafeCell`. A much cleaner solution is to store the data for `IsolateHandle` in its own allocation, since that also means an extant `IsolateHandle` won't prevent the other data in `IsolateAnnex` from being deallocated. I originally was going to store the pointer to `IsolateHandleInner` in its own isolate data slot, but realized it was also possible to just still store it in `IsolateAnnex` for simplicity.

### DIFF
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -43,6 +43,7 @@ use crate::support::int;
 use crate::support::size_t;
 use crate::wasm::WasmStreaming;
 use crate::wasm::trampoline;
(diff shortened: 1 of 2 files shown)


bartlomieju (approved):
Thanks!

PR #1893, opened 2025-12-23
feat: Callable objects

Aseminaunz:
Implements IsCallable, IsConstructor, CallAsFunction, CallAsConstructor for Objects. Also SetCallAsFunctionHandler for ObjectTemplates.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1440,6 +1440,13 @@ void v8__ObjectTemplate__SetImmutableProto(const v8::ObjectTemplate& self) {
   return ptr_to_local(&self)->SetImmutableProto();
 }
 
+void v8__ObjectTemplate__SetCallAsFunctionHandler(
+    const v8::ObjectTemplate& self, v8::FunctionCallback callback,
+    const v8::Value* data_or_null) {
+  ptr_to_local(&self)->SetCallAsFunctionHandler(callback,
+                                                ptr_to_local(data_or_null));
+}
+
 const v8::Object* v8__Object__New(v8::Isolate* isolate) {
   return local_to_ptr(v8::Object::New(isolate));
 }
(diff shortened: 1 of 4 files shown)


bartlomieju:
Thanks for this, and sorry it sat for so long — the feature is a real gap. There's no way to invoke a `Proxy` with a `[[Call]]` trap or anything else callable-but-not-a-`Function` today, and no way to build a callable object from a template at all.

I pushed a commit rather than send it back to you. Two of the extern declarations didn't match `binding.cc`:

```
v8___Object__IsConstructor                    (three underscores)
v8__ObjecTemplate__SetCallAsFunctionHandler   (missing 't' in Object)
```

Both would fail to link for anyone calling `is_constructor()` or `set_call_as_function_handler()` — so two of the five new APIs, including the template one. The reason it looked fine is that nothing in the PR called the new methods, so the symbols were never resolved and nothing surfaced it.

So I also added tests in `tests/test_api.rs` covering all five entry points, including both `_with_context` variants. That's what turns a mismatch like this into an immediate failure. I verified by compiling `binding.cc` to an object file and checking all five exported symbols against the Rust externs, and `cargo check --tests` plus `clippy` are clean both on your original base and rebased onto current `main`.

One cosmetic thing: dropped the top-level `const` on the two `bool`-returning binding functions, which is meaningless on a by-value return and doesn't match the neighbouring predicates.

The Rust side otherwise looked good and matched house style — the `_with_context` variants mirror `Function::call_with_context` exactly, redundant null check included.

Note I left the branch on its original base rather than rebasing, since it still merges into `main` cleanly. CI builds from source, so it will now actually exercise the new symbols.

bartlomieju (approved):
Thanks

PR #1896, opened 2026-01-10
Add v8::Locker, v8::Unlocker, and UnenteredIsolate bindings

max-lt:
Adds bindings for multi-threaded isolate pooling, enabling architectures like Cloudflare Workers.

## Changes

- **`UnenteredIsolate`**: New isolate type without auto-enter, removes LIFO drop constraint
- **`v8::Locker`**: Acquires exclusive thread access to an isolate
- **`v8::Unlocker`**: Temporarily releases a lock
- **`unsafe impl Send`** for `UnenteredIsolate` (safe with Locker)

## Key Features

- Drop isolates in any order (no LIFO panic)
- Share isolates across threads with `Arc<Mutex<UnenteredIsolate>>`
- <10µs warm start vs ~3-5ms creating new isolates
- 10,000+ req/s throughput per cached isolate

## Tests

- 298 lines of tests including multi-threaded usage, no LIFO constraint, concurrent access prevention
- Working POC with cooperative multitasking: https://gist.github.com/max-lt/2201664fbb7d6fef432da4f91b2eb004

**Note:** Each thread must call `unsafe { isolate.enter() }` before use to setup V8's thread-local state.

No breaking changes - all new APIs


max-lt:
Commit [de789a7](https://github.com/denoland/rusty_v8/pull/1896/commits/de789a7c3c84b0580ffb0a68f2b7f382a18ef581) makes Locker a pure mutex, add IsolateScope for Enter/Exit

Previous changes introduced Locker + unentered isolates to avoid OwnedIsolate's LIFO drop constraint. But the Locker itself still called Enter()/Exit() internally, which reintroduced the same LIFO constraint at the Locker level, when multiple Lockers for different isolates coexist on the same thread (async pool), dropping them out of order still corrupts the entry stack.

Last fix completes the decoupling:
  - Locker → pure mutex. No more Enter/Exit.
  - IsolateScope (new) → RAII Enter()/Exit(), short-lived around each synchronous V8 block, always properly nested.

```
  let locker = v8::Locker::new(&mut isolate);   // mutex only
  let _scope = locker.enter();                  // Enter(isolate)
  // ... V8 work ...
  // _scope dropped → Exit(isolate), before any await
  // locker dropped → unlock mutex, safe in any order
```

Tests: 10 safety tests (non-LIFO drop, concurrent lockers, stress) + 6 IsolateScope tests (interleaved isolates, round-robin).

bartlomieju:
Closing this in favour of #2044, which landed on `main` today and covers the same ground: `v8::Locker`, `Locker::unlock()` in place of a raw `Unlocker`, and `SharedIsolate` for moving an isolate across threads without the LIFO drop constraint. `SharedIsolate` is internally `Arc`'d and `lock()` takes `&self`, so the `Arc<Mutex<...>>` wrapper isn't needed any more.
 
Thanks for the persistence here, four rounds of rework across seven months, and your last comment landed on exactly the ordering `main` uses (lock before `Enter()`, `Exit()` before `unlock`, because `entry_stack_` isn't thread-safe). Sorry it didn't merge as-is.

If you port `openworkers` over, two constraints in the new API that weren't in your version are worth checking against your use case: `try_into_shared()` rejects isolates with a cppgc heap or live `Weak` handles/pending finalizers, and `Global::open()`/`Borrow` are disabled once shared (use a `Local` under a handle scope instead). If either blocks you, please open an issue, that's useful signal on the new API.

PR #1902, opened 2026-01-21
Fix memory leak in Function::get_script_origin

goffrie:
Previously v8__Function__GetScriptOrigin returned a newly allocated object from the C++ heap, then the caller never freed it.
Return it by-value instead.
I believe a handle scope is also required, as Function::GetScriptOrigin internally creates Local handles.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -4,6 +4,8 @@
 #include <cstddef>
 #include <cstdint>
 #include <cstdio>
+#include <map>
+#include <mutex>
 #include <thread>
 
 #include "cppgc/allocation.h"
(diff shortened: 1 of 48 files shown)


src/binding.cc
```diff
@@ -2713,16 +2711,16 @@ void v8__ScriptOrigin__CONSTRUCT(
       ptr_to_local(host_defined_options));
 }
 
-int v8__ScriptOrigin__ScriptId(const v8::ScriptOrigin& self) {
-  return ptr_to_local(&self)->ScriptId();
```
goffrie: also removed these ptr_to_local calls as v8::ScriptOrigin is not meant to be inside of a v8::Local<>, and at best this does nothing since `Local::operator->` just returns the original pointer again.

src/binding.cc
```diff
@@ -2713,16 +2712,16 @@ void v8__ScriptOrigin__CONSTRUCT(
       ptr_to_local(host_defined_options));
 }
 
-int v8__ScriptOrigin__ScriptId(const v8::ScriptOrigin& self) {
```
devsnek: these can all remain as references instead of pointers

PR #1915, opened 2026-02-25
Use CARGO_CFG_TARGET_FEATURE to detect CRT linkage on MSVC

jbg:
Replace the CARGO_FEATURE_CRT_STATIC feature flag check (which is testing whether the non-existent Cargo feature `crt-static` is enabled) with CARGO_CFG_TARGET_FEATURE to correctly detect crt-static.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -804,7 +804,10 @@ fn print_link_flags() {
   if target_env == "msvc" {
     // On Windows, including libcpmt[d]/msvcprt[d] explicitly links the C++
     // standard library, which libc++ needs for exception_ptr internals.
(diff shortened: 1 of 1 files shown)


bartlomieju (approved):
Fix is correct. `CARGO_FEATURE_CRT_STATIC` checks for a Cargo *feature* named `crt-static`, but no such feature exists in `Cargo.toml` — it's a target feature set via rustflags. `CARGO_CFG_TARGET_FEATURE` is the right way to detect it in a build script. The current code effectively always links `msvcprt` dynamically, which is wrong for static CRT builds.

Merged latest main to bring it up to date.

PR #1917, opened 2026-02-27
build: try to use Deno to download binary archive

bartlomieju:
This will allow us to skip installing Python on Deno CI


dsherret (approved):
Maybe extracting out to a separate function would be good, but LGTM

PR #1924, opened 2026-03-05
feat: add PlatformImpl trait for custom platform callbacks

fraidev:
Add a trait-based API (following the V8InspectorClientImpl pattern) that lets embedders hook into platform virtual methods. A CustomPlatform C++ class subclasses DefaultPlatform and delegates on_foreground_task_posted and on_isolate_shutdown to a Rust PlatformImpl trait object.

This allows embedders like Deno to receive notifications when V8 background threads post foreground tasks, enabling event-driven wakeups instead of polling.

### DIFF
--- a/tests/test_custom_platform.rs
+++ b/tests/test_custom_platform.rs
@@ -1,6 +1,6 @@
+use std::sync::Arc;
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
-use std::sync::Arc;
 
 struct TestPlatformImpl {
   post_task_count: Arc<AtomicUsize>,


kajukitli (commented):
few things:

1. **Thread leak in NotifyAfterDelay** - spawning and detaching a thread for every delayed task is bad. if you post 1000 delayed tasks you're spawning 1000 threads. this will OOM or hit thread limits on busy apps. consider using a single dedicated timer thread with a priority queue, or just call the callback immediately and let the embedder handle timing.

2. **Race in NotifyingTaskRunner** - the callback is invoked *after* posting the task to the wrapped runner. if another thread pumps the message loop immediately, it could execute the task before your callback fires. depending on what the callback does (wake event loop?), this might be fine, but worth documenting the ordering.

3. **Missing `UnregisterIsolate` equivalent** - NotifyIsolateShutdown cleans up runners, but DefaultPlatform also has `UnregisterIsolate()`. should you override that too? or is shutdown sufficient?

4. **`int` in rust ffi** - the extern decl uses `int` but rust ffi uses `c_int`. should be `std::os::raw::c_int` or `i32` explicitly.

5. **gen file committed** - is `gen/src_binding_release_aarch64-apple-darwin.rs` supposed to be committed? 946 lines of generated bindgen output seems like it should be in gitignore or generated at build time.

the overall design makes sense for waking event loops when background work completes. just needs the thread spawning fixed before this is viable for production.

src/binding.cc
```diff
@@ -3010,6 +3012,137 @@ v8::StartupData v8__SnapshotCreator__CreateBlob(
...
+ private:
+  void NotifyAfterDelay(double delay_in_seconds) {
+    if (delay_in_seconds <= 0) {
+      callback_(static_cast<void*>(isolate_));
+    } else {
+      // Schedule the callback to fire when the delay expires, so the
+      // event loop wakes and pumps the V8 message loop at the right time.
+      auto cb = callback_;
+      auto iso = isolate_;
+      std::thread([cb, iso, delay_in_seconds]() {
```
devsnek: if we go with a concrete "NotifyingTaskRunner" impl, instead of doing this, you should instead implement this by calling the callback with the given time, so that we can schedule it properly within tokio.

PR #1928, opened 2026-03-11
feat: expose simdutf bindings behind cargo feature

bartlomieju:
## Summary

- Exposes the bundled simdutf (v7.7.0) API through rusty_v8 behind a `simdutf` cargo feature flag
- Solves C++ symbol clashes when consumers (e.g. deno_core) try to use a separate simdutf Rust crate alongside rusty_v8
- SIMD-accelerated Unicode validation, transcoding, and base64 operations

## What's included

- **Validation**: UTF-8, ASCII, UTF-16 (LE/BE), UTF-32
- **Conversion**: All combinations of UTF-8, UTF-16 (LE/BE), UTF-32, Latin-1 (with and without length output)
- **Length calculation**: Output length estimation for all conversion pairs
- **Counting**: UTF-8/UTF-16 code point counting
- **Encoding detection**: Autodetect BOM-based encoding
- **Base64**: Encode/decode with configurable alphabets and chunk handling

## Design

- `simdutf` cargo feature → `rusty_v8_enable_simdutf` GN arg → `RUSTY_V8_ENABLE_SIMDUTF` preprocessor guard
- Thin `extern "C"` wrappers in `binding.cc` (follows existing `v8__`/`cppgc__` naming with `simdutf__` prefix)
- Safe Rust API layer in `src/simdutf.rs` — validation/length functions are safe, conversion functions are `unsafe` (caller provides output buffer)
- **Prebuilt binaries**: CI matrix updated to produce `_simdutf` prebuilt artifacts (mirroring the `_ptrcomp` pattern), so consumers can use `--features simdutf` without `V8_FROM_SOURCE`

## Test plan

- Tests in `tests/test_simdutf.rs` cover:
  - [x] UTF-8, ASCII, UTF-16LE, UTF-32 validation (valid + invalid inputs)
  - [x] Validation with error position reporting
  - [x] UTF-8 ↔ UTF-16LE round-trip conversion
  - [x] Latin-1 ↔ UTF-8 round-trip conversion
  - [x] UTF-8 ↔ UTF-32 round-trip conversion
  - [x] Codepoint counting (UTF-8, UTF-16LE)
  - [x] Length calculation cross-checks
  - [x] Encoding detection
  - [x] Base64 encode/decode round-trip (standard + URL-safe)
- [x] CI matrix produces prebuilt `_simdutf` artifacts for all major targets
- [ ] Verify builds without the feature still work (no simdutf dependency)

### DIFF
--- a/tests/test_simdutf.rs
+++ b/tests/test_simdutf.rs
@@ -24,10 +24,10 @@ fn validate_utf8() {
 #[test]
 fn validate_utf8_with_errors() {
   let r = simdutf::validate_utf8_with_errors(b"hello");
-  assert!(r.error.is_success());
+  assert!(r.is_ok());
 
   let r = simdutf::validate_utf8_with_errors(&[b'a', 0x80, b'b']);
-  assert!(!r.error.is_success());
+  assert!(!r.is_ok());
   assert_eq!(r.count, 1); // error at byte index 1
 }
 
@@ -217,7 +217,7 @@ fn base64_roundtrip() {
       simdutf::LastChunkHandling::Loose,
     )
   };
-  assert!(result.error.is_success());
+  assert!(result.is_ok());
   assert_eq!(&decoded[..result.count], input);
 }
 
@@ -248,6 +248,6 @@ fn base64_url_safe() {
       simdutf::LastChunkHandling::Loose,
     )
   };
-  assert!(result.error.is_success());
+  assert!(result.is_ok());
   assert_eq!(&decoded[..result.count], input);
 }


kajukitli (commented):
not comfortable approving this as-is

biggest issue: the new `simdutf` feature changes the prebuilt artifact suffix (`_simdutf`) but the PR doesn't show any corresponding prebuilt artifacts being produced/published. that means consumers enabling the cargo feature will miss the prebuilt binary and fall back to source builds, or just break depending on how they're consuming rusty_v8.

for rusty_v8, feature-gated native code isn't just a Rust API change — it changes the binary compatibility matrix. if we add `_simdutf` to `prebuilt_features_suffix()`, we need the release/build pipeline to actually produce those artifacts too.

also, the test plan is still basically unchecked for the important parts (`cargo build --features simdutf`, no-clash verification, builds without feature, etc). for a 1k+ line FFI surface, that's too hand-wavy.

PR #1931, opened 2026-03-14
fix: correct ExternalOneByteString destructor memory accounting

bartlomieju:
## Summary

- Fix double-negative bug in `ExternalOneByteString` destructor (`src/binding.cc:1237-1238`) where `-static_cast<int64_t>(-length_)` was used instead of `-static_cast<int64_t>(length_)`. Since `length_` is `size_t` (unsigned), the unary minus wraps it before the cast, causing V8 to think memory is being *allocated* instead of *freed* on every external string destruction.
- Over many string allocation/deallocation cycles (e.g. large file uploads in Deno), the external memory counter accumulates until it exceeds V8's 32GB sanity check (`kMaxReasonableBytes`), triggering a fatal `CHECK` failure.
- Add test `external_onebyte_string_frees_external_memory` that verifies the external memory counter correctly decreases after GC collects an external one-byte string.

Fixes denoland/deno#32693

## Test plan

- [x] New test `external_onebyte_string_frees_external_memory` allocates a 1MB external string, verifies the external memory counter increases, forces GC, then verifies the counter drops back down
- [ ] CI passes

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1235,7 +1235,7 @@ class ExternalOneByteString : public v8::String::ExternalOneByteStringResource {
   ~ExternalOneByteString() override {
     (*rustDestroy_)(data_, length_);
     isolate_->AdjustAmountOfExternalAllocatedMemory(
-        -static_cast<int64_t>(-length_));
+        -static_cast<int64_t>(length_));
   }
 
   const char* data() const override { return data_; }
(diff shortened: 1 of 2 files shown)


kajukitli (approved):
lgtm - fix is correct and test properly validates the memory accounting. the test allocates 1MB, checks external memory increases, forces GC, then verifies it drops back down. before this fix, the destructor would have increased the counter instead of decreasing.

PR #1945, opened 2026-03-25
build: Notify when no pre-built archive is available for the target

TheJanzap:
This PR adds a panic into `build.rs` when a v8 build is started with a target that v8 does not provide a pre-built archive for and `RUSTY_V8_MIRROR` is unset (i.e. the script downloads the archive from this GitHub repo). In this case, the panic message advises to rebuild v8 with `V8_FROM_SOURCE=1`.

I currently use a hardcoded list of targets for this, but if you have a better solution, please suggest so :)

I am also a bit unsure if the placement of the code is appropriate, as the build script is quite expansive, so feel free to also suggest improvements there.

Implements the feature discussed in #1877

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -644,7 +644,7 @@ fn build_dir() -> PathBuf {
     .to_path_buf()
 }
 
-/// Checks if the target is available as a pre-built archive from the rusty_v8 GitLab.
+/// Checks if the target is available as a pre-built archive from the rusty_v8 GitHub. 
 /// Should only be called when `RUSTY_V8_MIRROR` is unset.
 fn is_target_prebuilt(target: &str) -> bool {
   const PREBUILT_TARGETS: [&str; 6] = [
--- a/build.rs
+++ b/build.rs
@@ -596,7 +596,6 @@ fn static_lib_url() -> String {
   let target = env::var("TARGET").unwrap();
   let profile = prebuilt_profile();
   let features = prebuilt_features_suffix();
-
   format!(
     "{base}/v{version}/{}.gz",
     static_lib_name(&format!("{features}_{profile}_{target}")),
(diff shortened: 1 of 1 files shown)


bartlomieju (commented):
## Review

The idea is good — failing early with a clear message is much better than a confusing 404. But the implementation has several issues:

### Bug: The target list is wrong

The PR's `PREBUILT_TARGETS` list has `"aarch64-unknown-linux"`, but the actual Rust target triple (and the release asset name) is `"aarch64-unknown-linux-gnu"`. This means `is_target_prebuilt` would return `false` for aarch64 Linux, **incorrectly panicking** on one of the most common targets.

### Bug: Windows targets lack debug prebuilts

The Windows targets (`aarch64-pc-windows-msvc`, `x86_64-pc-windows-msvc`) only have release prebuilts in the actual GitHub releases. A debug build targeting Windows would pass the `is_target_prebuilt` check but still 404. The check doesn't account for `profile`.

### Design concern: Hardcoded list is fragile

The list will drift as targets are added/removed. An alternative approach: catch the 404 from the download and provide the helpful error message with the `V8_FROM_SOURCE=1` suggestion at that point. This would be self-maintaining and always accurate.

### Minor: `RUSTY_V8_MIRROR` detection

The check `base == default_base` is an indirect proxy for "RUSTY_V8_MIRROR is unset." A cleaner approach would be `env::var("RUSTY_V8_MIRROR").is_err()`.

### Summary

The `aarch64-unknown-linux` typo is a blocker — it would break aarch64 Linux builds. The missing debug/release distinction for Windows is a secondary concern. I'd suggest either:
1. Fix the list and add a `profile` check, or
2. (Better) Replace the hardcoded list with a 404-triggered error message in the download logic

bartlomieju:
Thanks for working on this — the problem is real and worth solving. However, I think we can get a better result with a simpler approach that doesn't require maintaining a hardcoded target list.

Instead of checking against a list upfront, we can improve the error message at the point where the download actually fails. Since many users will have Deno installed, we can get a good error message from the Deno downloader by checking the response status:

```js
const resp = await fetch(url);
if (resp.status === 404) {
  console.error(`No prebuilt V8 archive found at ${url}`);
  console.error(`If no prebuilt is available for your target, compile V8 from source with V8_FROM_SOURCE=1`);
  Deno.exit(1);
}
if (!resp.ok) Deno.exit(1);
```

That covers the majority of users. For the Python and curl fallback paths, we can replace the bare `assert!(status.success())` with a more helpful panic:

```rust
if !status.success() {
  panic!(
    "Failed to download V8 prebuilt archive from {url}\n\
     If no prebuilt archive is available for your target, \
     compile V8 from source by setting V8_FROM_SOURCE=1"
  );
}
```

This approach is self-maintaining (no list to keep in sync with releases), handles all edge cases automatically (features, profiles, new targets), and is a much smaller change.

TheJanzap:
@bartlomieju Implemented your changes and rebased on main :)

bartlomieju (approved):
Thanks

PR #1948, opened 2026-03-30
ci: build riscv64

Xeonacid:
Unlike aarch64, the riscv64 sccache binary is not static, so use the host architecture one. [1]

[1]: https://github.com/mozilla/sccache/pull/2658

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -117,6 +117,18 @@ jobs:
             v8_enable_pointer_compression: false
             cargo: cargo
 
+          - os: ${{ github.repository == 'denoland/rusty_v8' && 'ubuntu-22.04-xl' || 'ubuntu-22.04' }}
+            target: riscv64gc-unknown-linux-gnu
+            variant: debug
+            v8_enable_pointer_compression: false
+            cargo: cargo
+
+          - os: ${{ github.repository == 'denoland/rusty_v8' && 'ubuntu-22.04-xl' || 'ubuntu-22.04' }}
+            target: riscv64gc-unknown-linux-gnu
+            variant: release
+            v8_enable_pointer_compression: false
+            cargo: cargo
+
           # simdutf builds
           - os: macos-15-large
             target: x86_64-apple-darwin
@@ -174,6 +186,20 @@ jobs:
             simdutf: true
             cargo: cargo
 
+          - os: ${{ github.repository == 'denoland/rusty_v8' && 'ubuntu-22.04-xl' || 'ubuntu-22.04' }}
+            target: riscv64gc-unknown-linux-gnu
+            variant: debug
+            v8_enable_pointer_compression: false
+            simdutf: true
+            cargo: cargo
+
+          - os: ${{ github.repository == 'denoland/rusty_v8' && 'ubuntu-22.04-xl' || 'ubuntu-22.04' }}
+            target: riscv64gc-unknown-linux-gnu
+            variant: release
+            v8_enable_pointer_compression: false
+            simdutf: true
+            cargo: cargo
+
           - os: ${{ github.repository == 'denoland/rusty_v8' && 'windows-2022-xxl' || 'windows-2022' }}
             target: x86_64-pc-windows-msvc
             variant: release # Note: we do not support windows debug builds.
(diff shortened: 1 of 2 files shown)


bartlomieju (commented):
Thanks for adding riscv64 support! The build.rs changes look solid — the underlying V8/chromium infrastructure already supports riscv64 sysroots, so the plumbing is straightforward.

A few suggestions:

**sccache version bump (v0.8.2 → v0.14.0)**: Could you split this into a separate PR? It's an unrelated change that bumps the version only in the `build` job while the other three jobs (`build-asan`, `build-windows-arm64`, `build-windows-arm64-simdutf`) still use v0.8.2. Separating it will make both PRs easier to review and merge independently.

.github/workflows/ci.yml
```diff
@@ -234,6 +260,23 @@ jobs:
...
+          rustup target add riscv64gc-unknown-linux-gnu
+
+          sudo apt update
+          sudo apt install -yq --no-install-suggests --no-install-recommends \
+            binfmt-support g++-10-riscv64-linux-gnu \
+            gcc-10-riscv64-linux-gnu libc6-riscv64-cross debootstrap qemu qemu-user \
+            qemu-user-binfmt
+
+          sudo ln -s /usr/riscv64-linux-gnu/lib/ld-linux-riscv64-lp64d.so.1 \
+                     /lib/ld-linux-riscv64-lp64d.so.1
```
bartlomieju: This will fail if the symlink already exists (e.g. on re-runs or different runner images). Use `ln -sf` for idempotency:
```suggestion
          sudo ln -sf /usr/riscv64-linux-gnu/lib/ld-linux-riscv64-lp64d.so.1 \
                     /lib/ld-linux-riscv64-lp64d.so.1
```

.github/workflows/ci.yml
```diff
@@ -234,6 +260,23 @@ jobs:
           echo "CARGO_TARGET_AARCH64_UNKNOWN_LINUX_GNU_LINKER=/usr/bin/aarch64-linux-gnu-gcc-10" >> ${GITHUB_ENV}
           echo "QEMU_LD_PREFIX=/usr/aarch64-linux-gnu" >> ${GITHUB_ENV}
 
+      - name: Install cross compilation toolchain (riscv64)
+        if: matrix.config.target == 'riscv64gc-unknown-linux-gnu'
+        run: |
+          rustup target add riscv64gc-unknown-linux-gnu
+
+          sudo apt update
+          sudo apt install -yq --no-install-suggests --no-install-recommends \
```
bartlomieju, line 269: The aarch64 step doesn't install `debootstrap`. Is it actually needed for the riscv64 cross-compilation? If not, removing it would keep the install leaner.

build.rs
```diff
@@ -372,6 +372,15 @@ fn build_v8(is_asan: bool) {
     maybe_install_sysroot("i386");
     maybe_install_sysroot("arm");
   }
```
bartlomieju, line 374: The aarch64 block only sets `target_cpu` but not `v8_target_cpu`. Is the extra `v8_target_cpu` needed for riscv64? If so, a brief comment explaining why would help.

Xeonacid: Updated. Cross-compiling needs to set `v8_target_cpu`. The `arm` block also sets this.

.github/workflows/ci.yml
Xeonacid: Oops. It was added when I did some experiments to run riscv64 sccache. Dropped.

.github/workflows/ci.yml
Xeonacid: Thanks for the review! Updated. (and also the aarch64 block)

Xeonacid:
> sccache version bump (v0.8.2 → v0.14.0): Could you split this into a separate PR? It's an unrelated change that bumps the version only in the build job while the other three jobs (build-asan, build-windows-arm64, build-windows-arm64-simdutf) still use v0.8.2. Separating it will make both PRs easier to review and merge independently.

Thanks for the review! Reverted the changed and will send a PR later.

Xeonacid:
Any chance to get this PR reviewed/merged? Thank you very much!

littledivy (approved):
LGTM

PR #1957, opened 2026-04-13
build: fix static_lib_name for non-MSVC Windows toolchains

ognevny:
only MSVC uses such library names, other toolchains (like MinGW) apply lib prefix and use .a extension for static libraries


bartlomieju:
Thanks for the PR, but I'm wary to fix something that is working correctly without associated test or rationale. Can you provide more context on this change?

ognevny:
when we use GCC or Clang even on Windows, they use Linux-like way to name libraries. that includes lib prefix and .a extension for static libraries. also Rust's definition of windows-gnu and windows-gnullvm follow this

PR #1959, opened 2026-04-15
feat: add new_external_twobyte and new_external_twobyte_raw

bartlomieju:
## Summary

Adds two-byte (UTF-16) equivalents of the existing external one-byte string APIs, enabling zero-copy creation of V8 strings from owned UTF-16 data.

## New APIs

- **`String::new_external_twobyte(scope, buffer: Box<[u16]>)`** -- creates a `v8::String` from owned `Box<[u16]>`. V8 takes ownership and frees via `free_rust_external_twobyte` on GC.
- **`String::new_external_twobyte_raw(scope, buffer, len, destructor)`** -- creates a `v8::String` from a raw `*mut u16` with a custom destructor. Called when the string is GC'd. Unsafe.
- **`free_rust_external_twobyte`** -- public destructor for `Box<[u16]>` buffers.

These mirror `new_external_onebyte` / `new_external_onebyte_raw` / `free_rust_external_onebyte`.

## C++ binding

Adds `ExternalTwoByteString` class (mirrors `ExternalOneByteString`) which:
- Implements `v8::String::ExternalStringResource`
- Tracks external memory via `AdjustAmountOfExternalAllocatedMemory` (using `length * 2` for byte count)
- Calls the Rust destructor on destruction

## Motivation

Needed by denoland/deno#33283 for `node_api_create_external_string_utf16` to support zero-copy external UTF-16 strings (matching Node.js behavior).

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1291,6 +1291,35 @@ size_t v8__ExternalOneByteStringResource__length(
   return self->length();
 }
 
+class ExternalTwoByteString : public v8::String::ExternalStringResource {
+ public:
+  using RustDestroy = void (*)(uint16_t*, size_t);
+  ExternalTwoByteString(uint16_t* data, size_t length, RustDestroy rustDestroy,
+                        v8::Isolate* isolate)
+      : data_(data),
+        length_(length),
+        rustDestroy_(rustDestroy),
+        isolate_(isolate) {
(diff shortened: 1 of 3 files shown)


nathanwhit (approved):
LGTM but could use a test

PR #1960, opened 2026-04-15
feat: implement NewEscapableHandleScope for Isolate

bartlomieju:
## Summary

Implement `NewEscapableHandleScope` for `Isolate`, allowing `EscapableHandleScope::new(isolate)` without requiring a parent `PinnedRef<HandleScope>`.

This matches V8's C++ API where `v8::EscapableHandleScope` can be constructed from just an `Isolate*`.

## Motivation

Deno's Node-API implementation (denoland/deno#33285) needs to create escapable handle scopes in NAPI callbacks where only an `Isolate` reference is available (not a `PinnedRef<HandleScope>`). Without this, `napi_open_escapable_handle_scope` can't use V8's proper escape slot mechanism and has to fall back to copying handle values directly.

## Changes

- Add `impl NewEscapableHandleScope for Isolate` in `scope.rs` (19 lines), mirroring the existing `impl NewHandleScope for Isolate`
- Creates an `EscapeSlot` + `HandleScope` pair, same as the `PinnedRef<HandleScope>` impl but without inheriting a context
- Add test `escapable_handle_scope_from_isolate` verifying escape works correctly

### DIFF
--- a/src/scope.rs
+++ b/src/scope.rs
@@ -1471,6 +1471,25 @@ pub trait NewEscapableHandleScope<'s> {
   fn make_new_scope(me: &'s mut Self) -> Self::NewScope;
 }
 
+impl<'s> NewEscapableHandleScope<'s> for Isolate {
+  type NewScope = EscapableHandleScope<'s, 's, ()>;
+
+  fn make_new_scope(me: &'s mut Self) -> Self::NewScope {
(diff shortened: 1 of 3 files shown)


bartlomieju:
Spoke with @nathanwhit, this seems desirable

PR #1962, opened 2026-04-17
fix: correct v8__String__NewExternalOneByte length parameter type

bartlomieju:
## Summary

- Fix FFI type mismatch in `v8__String__NewExternalOneByte`: the C++ function declared `int length` (4 bytes) while the Rust FFI binding declared `size_t` (8 bytes)
- Aligns with `v8__String__NewExternalTwoByte` which already correctly uses `size_t length`

Related: denoland/deno#32693, denoland/deno#32904

## Test plan

- [x] `external_onebyte_string` test passes
- [x] `external_onebyte_string_frees_external_memory` test passes


nathanwhit (approved):
good catch

PR #1968, opened 2026-04-22
perf: cache IsolateAnnex pointer in scope structs

bartlomieju:
## Summary
- Caches a `NonNull<IsolateAnnex>` pointer in `HandleScope`, `EscapableHandleScope`, and `CallbackScope`, avoiding an FFI call to `v8__Isolate__GetData` on every `get_slot`/`set_slot`/`get_slot_mut`/`remove_slot` call through a scope
- Adds slot accessor methods on `PinnedRef<HandleScope>` that shadow the `Isolate` methods in the Deref chain — existing user code benefits without any API changes
- The FFI cost is paid once at scope creation; nested scopes just copy the pointer from the parent at zero cost
- Refactors slot logic into `IsolateAnnex` methods to avoid duplication between the `Isolate` and scope paths

## Test plan
- [x] `cargo check` passes
- [x] `cargo test --test test_api` — 238/239 pass (only pre-existing `get_version` mismatch fails)
- [x] Verified slot tests (`isolate_data_slots`, `clear_slots_annex_uninitialized`, `host_create_shadow_realm_context_callback`, `wasm_streaming_callback`, `request_interrupt_small_scripts`) all pass

### DIFF
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -1036,6 +1036,17 @@ impl Isolate {
     unsafe { &mut *annex_ptr }
   }
 
+  /// Returns a non-null pointer to the isolate's annex data.
+  /// This is used by scopes to cache the annex pointer and avoid
+  /// repeated FFI calls to `v8__Isolate__GetData`.
+  #[inline(always)]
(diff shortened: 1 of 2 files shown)


nathanwhit (approved):
LGTM, makes sense

PR #1973, opened 2026-04-30
Add bindings for String::Concat

bartlomieju:
## Summary
- Add C++ binding, Rust FFI, and public `v8::String::concat(scope, left, right)` method wrapping `v8::String::Concat`
- Add test covering normal concatenation and concatenation with an empty string

## Test plan
- [ ] CI passes (`string_concat` test)

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1184,6 +1184,13 @@ const v8::String* v8__String__NewFromTwoByte(v8::Isolate* isolate,
       v8::String::NewFromTwoByte(isolate, data, new_type, length));
 }
 
+const v8::String* v8__String__Concat(v8::Isolate* isolate,
+                                     const v8::String& left,
+                                     const v8::String& right) {
+  return local_to_ptr(
+      v8::String::Concat(isolate, ptr_to_local(&left), ptr_to_local(&right)));
+}
+
 int v8__String__Length(const v8::String& self) { return self.Length(); }
 
 int v8__String__Utf8Length(const v8::String& self, v8::Isolate* isolate) {
--- a/src/string.rs
+++ b/src/string.rs
@@ -106,6 +106,12 @@ unsafe extern "C" {
     length: int,
   ) -> *const String;
 
+  fn v8__String__Concat(
+    isolate: *mut RealIsolate,
+    left: *const String,
+    right: *const String,
+  ) -> *const String;
+
   fn v8__String__Length(this: *const String) -> int;
 
   fn v8__String__Utf8Length(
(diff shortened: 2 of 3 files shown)


fibibot (commented):
Substance is right for the happy path — `v8__String__Concat` matches the V8 signature (`static Local<String> Concat(Isolate*, Local<String>, Local<String>)`), and the `cast_local` wrapper follows the standard rusty_v8 pattern. Test covers normal concatenation and the empty-string identity. ✓

## One concern — `.unwrap()` is a panic vector for valid input

```rust
pub fn concat<'s>(
  scope: &PinScope<'s, '_, ()>,
  left: Local<String>,
  right: Local<String>,
) -> Local<'s, String> {
  unsafe {
    scope
      .cast_local(|sd| {
        v8__String__Concat(sd.get_isolate_ptr(), &*left, &*right)
      })
      .unwrap()
  }
}
```

`v8::String::Concat` returns an empty `Local<String>` when the combined length would exceed `v8::String::kMaxLength` (≈2^29 − 24 chars on 64-bit, ≈2^28 − 16 on 32-bit). At that point V8 also raises a `RangeError` on the isolate. Two ways this panics on legitimate input:

1. Two ~256 MB strings concatenated → empty Local → `.unwrap()` panics.
2. A program that already produced a large string and concatenates it with anything pushing past the limit.

These aren't unreachable cases — they're the exact category of input where users would *want* to handle the failure rather than get a process abort.

Other string constructors in `string.rs` already return `Option`:
- `String::new_from_utf8(...) -> Option<Local<'s, String>>`
- `String::new_external_one_byte_static(...) -> Option<Local<'s, String>>`

Suggested signature change:

```rust
pub fn concat<'s>(
  scope: &PinScope<'s, '_, ()>,
  left: Local<String>,
  right: Local<String>,
) -> Option<Local<'s, String>> {
  unsafe {
    scope.cast_local(|sd| {
      v8__String__Concat(sd.get_isolate_ptr(), &*left, &*right)
    })
  }
}
```

Caller-side this is a one-character change (`.unwrap()` if they really want today's behavior, `?` if they're propagating, `else { ... }` if they want to surface the RangeError).

A regression test under the new signature would just need a `String::repeat`-driven payload that brushes against `kMaxLength` and asserts `concat(big, big).is_none()` — the empty-string test stays as is.

## Minor

- The `<'s>` lifetime on the return type is correct (output lives as long as the scope), but worth noting it's distinct from the lifetime on `left`/`right`. The new string is *not* aliased to either input — V8 may return one of the inputs unchanged for an empty-other-side, but that's still a fresh `Local` bound to `scope`. Not a bug, just future-reader food.
- C++ binding ordering: it's placed between `NewFromTwoByte` and `Length` in `binding.cc`, which fits the "string constructors" cluster. ✓

## CI

Mostly pending. Holding **COMMENT** until CI lands and the panic-on-overflow concern is addressed.

PR #1944, opened 2026-03-25
Move V8 patch management into rusty_v8

bartlomieju:
## Summary
- Move V8 patch files from `denoland/v8` repo into `patches/` directory in this repo
- Rewrite `tools/auto_update_v8.ts` to be a single unified script that fetches upstream V8 directly, applies patches, pushes to the denoland/v8 fork, and updates all deps
- Remove dependency on `denoland/v8`'s separate autoroll cron job

Previously managing V8 patches required coordinating two repos (`denoland/v8` and `denoland/rusty_v8`) with two separate cron jobs. Cherry-picking a commit or adding a patch meant working in `denoland/v8` first, waiting for its autoroll to run, then waiting for `rusty_v8`'s autoroll to pick it up.

Now:
- **Add a patch**: drop a `.patch` file in `patches/`
- **Bump V8 version**: change `V8_VERSION` at the top of `auto_update_v8.ts`
- **Cherry-pick a commit**: generate a patch file and add it to `patches/`

The `denoland/v8` repo remains as a submodule host but its autoroll workflow can be disabled.

## Test plan
- [ ] Verify `deno run -A ./tools/auto_update_v8.ts` works end-to-end
- [ ] Disable autoroll workflow in `denoland/v8` after merging

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1977, opened 2026-05-06
perf: speed up V8 snapshot deserialization

bartlomieju:
## Summary

- **Skip write barriers during snapshot deserialization.** During startup/shared-heap/context deserialization all objects go to old space and incremental marking is inactive, making every write barrier a guaranteed no-op. Each still pays ~8 instructions (MemoryChunk lookup + branches) per slot write — for ~875K slots, that's millions of wasted instructions.
- **Compile out `trace_deserialization` flag checks.** 31 `if (v8_flags.trace_deserialization)` checks in the bytecode dispatch loop bloat `ReadSingleBytecodeData` and hurt icache. Replaced with `TRACE_DESER` macros that compile to nothing unless `V8_ENABLE_TRACE_DESERIALIZATION` is defined.

Estimated combined improvement: **~15-25%** of total snapshot deserialization time (~2-3ms out of ~11ms on Apple Silicon).

## Test plan

- [ ] Verify patch applies: `cd v8 && git am -3 ../patches/0004-*.patch`
- [ ] Build rusty_v8 and run existing tests
- [ ] Measure with `deno run --v8-flags=--profile-deserialization empty.js` before/after

### DIFF
--- a/patches/0004-Speed-up-snapshot-deserialization-for-embedders.patch
+++ b/patches/0004-Speed-up-snapshot-deserialization-for-embedders.patch
@@ -1,0 +1,505 @@
+From 8a266cd14e00e02d7e8f2c5f7f9fe2d407b3dec8 Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Bartek=20Iwa=C5=84czuk?= <biwanczuk@gmail.com>
+Date: Thu, 7 May 2026 09:26:04 +0200
+Subject: [PATCH] Speed up snapshot deserialization for embedders
+
+Two optimizations targeting the hot deserialization loop:
+
+1. Skip write barriers during non-user-code deserialization.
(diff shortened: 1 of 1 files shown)


bartlomieju:
Closing this for now — it's blocked on #1944 landing (it drops a file into `patches/`, which doesn't exist yet), and it was never validated: the test plan is unchecked and CI only ran the CLA check, so the ~15-25% figure is still an estimate.

If picked back up, two things worth doing: add a DCHECK asserting the invariant the write-barrier skip relies on (all objects in old space, incremental marking inactive, across startup/shared-heap/context deserialization) rather than documenting it in a comment, since the failure mode after a future V8 roll would be silent heap corruption; and consider upstreaming to V8 rather than carrying it as a local patch, since it's a general embedder win and a carried patch has to survive every roll.

PR #1978, opened 2026-05-07
fix: keep isolate annex alive during teardown

disjukr:
## Summary

**This PR was investigated and implemented with AI assistance. The diagnosis, patch, and verification were reviewed locally before opening.**

This fixes an isolate teardown panic where rusty_v8 callbacks/destructors can re-enter APIs that need `IsolateAnnex` after `ANNEX_SLOT` has already been cleared.

The panic was observed from Deno 2.7.14 while building `disjukr/bunja` with tsdown/rolldown:

https://github.com/disjukr/bunja/actions/runs/25501830756/job/74836577957

The Deno-side regression was introduced when Deno upgraded rusty_v8 from `147.2.1` to `147.4.0`:

https://github.com/denoland/deno/commit/d752d2e26ad1317355f01c80353f34c2d9e5cc6f

The rusty_v8-side behavioral regression appears to come from:

https://github.com/denoland/rusty_v8/commit/7a75586a6ad65f6a2234b12281d32fbd9c46d02c

That commit changed `IsolateHandle` ownership and made `dispose_annex()` take the annex, clearing `ANNEX_SLOT`, before V8 isolate disposal had finished. V8 teardown can still invoke Rust-side callbacks, such as weak second-pass callbacks, that access the annex.

## Changes

- Add a focused regression test for annex access during isolate teardown.
- Split annex teardown into:
  - preparation while keeping `ANNEX_SLOT` available
  - V8 isolate disposal
  - final annex ownership recovery/drop
- Keep `ANNEX_SLOT` alive until after `Isolate::dispose()`.
- Drain guaranteed finalizers before isolate disposal, preserving the documented ordering.
- Drain finalizers again when finishing annex disposal as a defensive cleanup.
- Move slot storage out before dropping it to avoid dropping user slot values while mutating `annex.slots`.
- Store `create_param_allocations` as `Option<Box<dyn Any>>` so it can be taken during annex preparation without replacing the annex allocation.

## Testing

- Verified the new regression test fails on unpatched `v147.4.0` with:

  ```text
  assertion failed: !annex_ptr.is_null()
  ```

- Verified the regression test passes with this fix on `v147.4.0`.

- Built Deno with this rusty_v8 branch patched in.

- Verified the original bunja reproduction no longer panics:

  ```powershell
  deno task build
  ```

  Result: CJS and ESM builds complete successfully.

## Notes

The regression test is a reduced repro for the same teardown invariant rather than the exact tsdown/rolldown/N-API path. The production failure goes through a weak second-pass callback during V8 isolate disposal; the test uses an isolate slot destructor that re-enters `thread_safe_handle()` after teardown has started. Both hit the same previous failure mode: `ANNEX_SLOT` was cleared too early.

### DIFF
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -1010,6 +1010,12 @@ impl Isolate {
     self.set_data_internal(Self::ANNEX_SLOT, annex_ptr as *mut _);
   }
 
+  /// Prepare annex teardown while keeping `ANNEX_SLOT` pointing at the annex.
+  ///
+  /// The returned pointer must be passed exactly once to
+  /// `finish_annex_dispose`. Until then, callers may only derive temporary
+  /// mutable references from that pointer and must not let them overlap,
+  /// because `ANNEX_SLOT` still points at the same allocation.
   unsafe fn prepare_annex_for_dispose(
     &mut self,
   ) -> (*mut IsolateAnnex, Box<dyn Any>) {
@@ -1052,10 +1058,11 @@ impl Isolate {
       unsafe { self.prepare_annex_for_dispose() };
     let annex = unsafe { &mut *annex_ptr };
     Self::run_remaining_guaranteed_finalizers(annex);
-    // SAFETY: `dispose_annex()` is only called once.
+    // SAFETY: `create_blob` is the only caller. It consumes the isolate and
+    // takes `ANNEX_SLOT` here, so the annex is dropped exactly once.
     let taken_annex =
       self.take_data_internal(Self::ANNEX_SLOT) as *mut IsolateAnnex;
-    assert_eq!(taken_annex, annex_ptr);
+    debug_assert_eq!(taken_annex, annex_ptr);
     unsafe { drop(Box::from_raw(annex_ptr)) };
     create_param_allocations
   }
@@ -2107,8 +2114,8 @@ impl IsolateHandle {
   // TODO: have this return an `Option<NonNull<RealIsolate>>`
   pub(crate) unsafe fn get_isolate_ptr(&self) -> *mut RealIsolate {
     // SAFETY: this function must only be called from the main thread of the
-    // isolate, which means that `Isolate::dispose_annex` can't concurrently
-    // be setting this to null.
+    // isolate. On that thread, the caller cannot race with teardown code that
+    // sets this pointer to null.
     unsafe { *self.0.isolate.get() }
   }
 
@@ -2237,6 +2244,8 @@ impl Drop for OwnedIsolate {
       Isolate::run_remaining_guaranteed_finalizers(annex);
       Platform::notify_isolate_shutdown(&get_current_platform(), self);
       self.dispose();
+      // ANNEX_SLOT still points at annex_ptr, but V8 has disposed the isolate
+      // so the embedder slot storage is no longer reachable.
       Isolate::finish_annex_dispose(annex_ptr);
     }
   }
--- a/src/isolate.rs
+++ b/src/isolate.rs
@@ -1012,54 +1012,113 @@ impl Isolate {
 
   /// Prepare annex teardown while keeping `ANNEX_SLOT` pointing at the annex.
   ///
+  /// Nulls the `IsolateHandle`'s inner pointer, reclaims
+  /// `create_param_allocations`, and drops the slot storage. The annex
+  /// allocation itself stays alive and `ANNEX_SLOT` keeps pointing at it,
+  /// so code that runs during the subsequent V8 teardown GC (weak
+  /// callbacks, guaranteed finalizers, embedder slot drops) can still
+  /// resolve the annex through `get_annex()` / `get_annex_mut()`.
+  ///
   /// The returned pointer must be passed exactly once to
-  /// `finish_annex_dispose`. Until then, callers may only derive temporary
-  /// mutable references from that pointer and must not let them overlap,
-  /// because `ANNEX_SLOT` still points at the same allocation.
+  /// [`Self::finish_annex_dispose`] (or, on the snapshot path, dropped by
+  /// [`Self::dispose_annex`]).
+  ///
+  /// # Safety
+  ///
+  /// Called once per isolate, from teardown paths only.
   unsafe fn prepare_annex_for_dispose(
     &mut self,
   ) -> (*mut IsolateAnnex, Box<dyn Any>) {
     let annex_ptr =
       self.get_data_internal(Self::ANNEX_SLOT) as *mut IsolateAnnex;
     assert!(!annex_ptr.is_null());
-    let annex = unsafe { &mut *annex_ptr };
 
-    // Set the `isolate` pointer inside the annex struct to null, so any
-    // IsolateHandle that outlives the isolate will know that it can't call
-    // methods on the isolate.
-    annex.global_liveness().dispose();
-    annex.isolate_handle.dispose();
+    // Each step below operates through the raw pointer rather than a
+    // long-lived `&mut IsolateAnnex`. The borrows we form here are
+    // narrowly-scoped expressions that end before any user-controlled
+    // Drop runs. This matters because slot Drops, weak callbacks, and
+    // guaranteed finalizers may re-enter the isolate (e.g. via
+    // `Isolate::thread_safe_handle()`) and resolve the annex through
+    // `get_annex()`. An outer `&mut IsolateAnnex` held across that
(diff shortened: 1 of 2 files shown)


src/isolate.rs
```diff
@@ -994,26 +994,52 @@ impl Isolate {
     self.set_data_internal(Self::ANNEX_SLOT, annex_ptr as *mut _);
   }
 
-  unsafe fn dispose_annex(&mut self) -> Box<dyn Any> {
-    // SAFETY: `dispose_annex()` is only called once, when the `Isolate` is dropped.
-    let mut annex = unsafe { self.take_annex() };
+  unsafe fn prepare_annex_for_dispose(
```
bartlomieju, line 1029: Worth a short doc comment here documenting the contract: pair with `finish_annex_dispose` exactly once, and don't let any other `&mut IsolateAnnex` overlap with the returned `*mut IsolateAnnex` (since `ANNEX_SLOT` still points at the same allocation in between). The function is internal but the prepare/finish split is non-obvious without it.

src/isolate.rs
```diff
@@ -994,26 +994,52 @@ impl Isolate {
...
+    let mut annex = unsafe { Box::from_raw(annex_ptr) };
+    Self::run_remaining_guaranteed_finalizers(&mut annex);
+  }
 
+  unsafe fn dispose_annex(&mut self) -> Box<dyn Any> {
+    let (annex_ptr, create_param_allocations) =
+      unsafe { self.prepare_annex_for_dispose() };
+    let annex = unsafe { &mut *annex_ptr };
+    Self::run_remaining_guaranteed_finalizers(annex);
+    // SAFETY: `dispose_annex()` is only called once.
```
bartlomieju: This SAFETY comment is now stale — `dispose_annex()` is no longer reachable from `OwnedIsolate::drop` (that path was inlined). It's only called from `create_blob` now. Either update the comment to reflect that, or consider inlining `dispose_annex` into `create_blob` since it has a single caller.

src/isolate.rs
```diff
@@ -994,26 +994,52 @@ impl Isolate {
...
+    let taken_annex =
+      self.take_data_internal(Self::ANNEX_SLOT) as *mut IsolateAnnex;
+    assert_eq!(taken_annex, annex_ptr);
```
bartlomieju: `debug_assert_eq!` would be more consistent with the other annex accessors (`get_annex_ptr` uses `debug_assert!`) and avoid a runtime check on the teardown hot path. The invariant is purely internal.

src/isolate.rs
```diff
@@ -2161,9 +2178,13 @@ impl Drop for OwnedIsolate {
...
       // self.dispose_scope_root();
       self.exit();
-      self.dispose_annex();
+      let (annex_ptr, _create_param_allocations) =
+        self.prepare_annex_for_dispose();
+      let annex = &mut *annex_ptr;
+      Isolate::run_remaining_guaranteed_finalizers(annex);
       Platform::notify_isolate_shutdown(&get_current_platform(), self);
       self.dispose();
+      Isolate::finish_annex_dispose(annex_ptr);
```
bartlomieju, line 2321: Note that `ANNEX_SLOT` is never explicitly cleared on this path — it stays pointing at the (now-dropped) box, but the V8 isolate has been disposed so the slot storage is unreachable. A one-line comment here would save the next reader from flagging it as a UAF in audit.

tests/test_api.rs
```diff
@@ -10918,6 +10918,31 @@ fn finalizer_on_kept_global() {
...
+    }
+  }
+
+  {
+    let isolate = &mut v8::Isolate::new(Default::default());
+    let isolate_ptr = unsafe { isolate.as_raw_isolate_ptr() };
+    isolate.set_slot(DropCheck {
+      isolate: isolate_ptr,
+    });
+  }
```
bartlomieju, line 11083: Test repros the failure mode well, but the assertion is implicit (passes on no-panic). Consider asserting something positive too — e.g. inside `DropCheck::drop`, check that the cloned handle reports the isolate as already disposed (`with_isolate_ptr` returns `None`), since `annex.isolate_handle.dispose()` was called in `prepare_annex_for_dispose`. That documents the invariant explicitly rather than just "shouldn't panic."

bartlomieju (commented):
Nice fix — the root-cause analysis matches what I see in the diff: the pre-patch `take_annex()` nulled `ANNEX_SLOT` before V8's `Dispose()` ran weak callbacks, so any callback (or user slot `Drop`) reaching for `thread_safe_handle()` hit the `assert!(!annex_ptr.is_null())` in `get_annex()`. Splitting into prepare/V8-dispose/finish, and switching slot teardown from `clear()` to `mem::take` + `drop`, both look correct.

A few smaller things, mostly comments/cleanup. Nothing blocking.

**Other notes (not inline since they're broader):**
- The `Option<Box<dyn Any>>` on `create_param_allocations` is needed because we now hold `&mut IsolateAnnex` instead of an owned `Box<IsolateAnnex>` and can't move fields out. The phrasing "avoid a dummy teardown allocation" in the PR description is a stretch — there was no dummy in the original — but the new shape is fine.
- Double-draining guaranteed finalizers (before and after `self.dispose()`) is the right call: V8 weak callbacks during disposal can register new ones, and `drain()` consumes them so there's no double-run risk.
- The stale safety comment at `IsolateHandle::get_isolate_ptr` (mentions `dispose_annex` setting the pointer to null) should be reworded — on the `Drop` path it's now V8's own `Dispose()` that effectively invalidates the slot. Couldn't leave it inline since the line isn't in a diff hunk.
- Worth running the snapshot-creation tests too, since the `create_blob` path (which still calls `dispose_annex`) got the same reordering.

disjukr:
Thanks for the detailed review.

The latest commit addresses the feedback:
- added a contract comment for the prepare/finish annex teardown split
- updated the stale safety comments
- changed the internal annex pointer check to `debug_assert_eq!`
- documented why `ANNEX_SLOT` is not explicitly cleared after V8 disposal
- added a positive assertion to the regression test for the disposed-handle invariant

The PR description was also updated to avoid the misleading "dummy teardown allocation" wording.

For local verification, the new regression test and the integration test targets pass.

bartlomieju:
Pushed a follow-up commit (9a217b9) directly to the branch since maintainer edits are on.

Two things it tightens:

1. **Aliasing under user Drop.** The previous `prepare_annex_for_dispose` held a `&mut IsolateAnnex` across `drop(slots)`. The added regression test `isolate_slot_drop_can_access_annex_during_teardown` exercises a slot `Drop` that calls `Isolate::thread_safe_handle()` — which derives a `&IsolateAnnex` from `ANNEX_SLOT` while the enclosing `&mut` is still in scope. Overlapping shared/mutable borrows to the same allocation; Miri / Stacked Borrows / Tree Borrows would flag it. Rewritten to operate through `*mut IsolateAnnex` field projections so each borrow ends before any user-controlled code runs. Same pattern applied to `run_remaining_guaranteed_finalizers`: `mem::take` the `finalizer_map` into a local before draining, so re-entrant `get_annex_mut()` doesn't alias an in-flight `&mut`.

2. **Doc + invariant comments.** Spelled out the safety contract on `prepare_annex_for_dispose`, `run_remaining_guaranteed_finalizers`, `finish_annex_dispose`, and `dispose_annex`; documented why `create_param_allocations: Option<…>` is sound to `.unwrap()`; noted the two-phase guaranteed-finalizer drain (pre-V8-dispose + inside `finish_annex_dispose`) in `OwnedIsolate::drop`.

No behavior change vs. your fix — same teardown ordering, same `ANNEX_SLOT`-stays-installed-across-`dispose()` invariant, same regression test passes pre-fix on the slot-drop path and clean post-fix.

I also took a swing at adding a direct weak-second-pass-callback test (`Weak::with_finalizer` leaked via `into_raw` + explicit GC + drop isolate), but couldn't get V8 to defer `second_pass_callback` past the explicit GC into the dispose-time GC — it consistently runs it in the same cycle. The slot-drop test covers the annex-keep-alive invariant via the same `get_annex_mut` entry point, so I dropped the redundant test.

Ran `cargo test --test test_api` (245 passed) and `cargo test --test test_cppgc` (3 passed) on aarch64-apple-darwin. The lone library-test flake (`scope::tests::deref_types`, "Invalid global state" in V8.rs:191) reproduces on the original branch without my changes, so it's pre-existing.

Once CI is green I'd like to merge and cut 149.2.0 so we can ship denoland/deno#34311.

bartlomieju (approved):

PR #1979, opened 2026-05-11
feat: add v8_lite_mode cargo feature for jitless builds

bartlomieju:
Adds an opt-in cargo feature, `v8_lite_mode`, that produces a smaller
prebuilt static-library archive by building V8 with the optimizing JIT
tiers compiled out: Turbofan, Maglev and Sparkplug are disabled and the
runtime is switched to jitless mode. WebAssembly support is retained
via the DrumBrake interpreter, so embedders that don't need the JIT
tiers can still run Wasm.

DrumBrake requires pointer compression and only supports x64/arm64 on
macOS, Linux and Windows, so the feature transitively enables
`v8_enable_pointer_compression` and the CI matrix publishes the
resulting `_ptrcomp_lite` release archive for exactly those targets.
build.rs maps the feature to the corresponding GN args
(`v8_jitless`, `v8_enable_turbofan`, `v8_enable_maglev`,
`v8_enable_sparkplug`, `v8_enable_webassembly`, `v8_enable_drumbrake`)
and threads `_lite` through the prebuilt binary suffix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ permissions: write-all
 
 jobs:
   build:
-    name: ${{ matrix.config.variant }} ${{ matrix.config.target }} ${{ matrix.config.v8_enable_pointer_compression && 'ptrcomp' || '' }} ${{ matrix.config.simdutf && 'simdutf' || '' }}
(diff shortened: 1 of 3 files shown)


Hajime-san:
Oh, I was just about to try out JIT-less build in a side project.😃
Would it be speed up the time of bootstrapping on the environment like AWS Lambda if JIT is disabled?

bartlomieju:
> Would it be speed up the time of bootstrapping on the environment like AWS Lambda if JIT is disabled?

Not directly, this is only for experiment to see how small V8 build can be.

PR #1981, opened 2026-05-14
ci: declare empty permissions on release workflow

arpitjain099:
Pins the release job to `permissions: {}`. The job checks out the repo with `secrets.DENOBOT_PAT` and runs `deno run -A .../publish_release.ts` with `GITHUB_TOKEN: ${{ secrets.DENOBOT_PAT }}`. The workflow's own `GITHUB_TOKEN` is never used.

Defense-in-depth angle: a compromised third-party action (cf. [CVE-2025-30066](https://github.com/advisories/GHSA-mrrh-fwg8-r2c3) on `tj-actions/changed-files`) runs inside the existing job context and exfiltrates whatever scope the workflow token holds via build logs. With `permissions: {}` the token has zero scope to leak; release authority lives only in `DENOBOT_PAT`.

Matches the per-job permission style already used in this repo's hardened workflows. YAML validated locally with `yaml.safe_load`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,6 +12,8 @@ on:
           - patch
         required: true
 
+permissions: {}
+
 jobs:
   rust:
     name: release


bartlomieju:
Thanks for the patch, and apologies for the slow response.

The change itself is harmless — nothing in `release.yml` reads the workflow's own `GITHUB_TOKEN`. Checkout gets an explicit `token: ${{ secrets.DENOBOT_PAT }}`, and the release step shadows `GITHUB_TOKEN` with the same PAT, so stripping the default token's scope wouldn't break anything.

I'm going to close it anyway, because it doesn't buy us what the description says it does. The CVE-2025-30066 scenario is a compromised action running inside the job — but in that scenario `permissions: {}` doesn't help here, since anything running in this job can read `DENOBOT_PAT` directly out of the step environment. That PAT is strictly more powerful than any scope the workflow token would have carried. So the one workflow in this repo that already isolates its authority in a secret is also the one where this hardening has the least effect.

The description also says this matches a per-job permission style in our hardened workflows. There isn't one — `ci.yml`, `nightly.yml`, `update-v8.yml` and `autoroll-notify.yml` all set `permissions: write-all` at the top level. That's the actual exposure, and `ci.yml` most of all since it's what runs against outside contributions.

If you're interested in a follow-up that would genuinely help, scoping those four down from `write-all` to what each one needs, and pinning third-party actions to commit SHAs instead of floating tags, is the mitigation that CVE actually argues for. Happy to review that one.

arpitjain099:
That's a fair close and the reasoning is right. `DENOBOT_PAT` sitting in the step environment does dominate anything the workflow token would have carried, so `permissions: {}` on `release.yml` was hardening the wrong thing. I should have checked what the job's authority actually came from before writing the description.

I'd like to take you up on the follow-up. I went through the three workflows on main, and a couple of details differ from your description, so let me lay out what I'd actually change:

- `ci.yml` (`permissions: write-all`, line 11) and `update-v8.yml` (line 8) are the two that set it. `release.yml` has no `permissions` block at all, and I could not find `nightly.yml` or `autoroll-notify.yml` on main, so I assume those are gone or live elsewhere.
- In `ci.yml` the only step that needs write is `softprops/action-gh-release`, in `build` and `build-asan`, gated on a tag ref. So top-level `contents: read` plus `contents: write` on those two jobs covers it, and the remaining jobs need nothing.
- `update-v8.yml` does all its writing through `DENOBOT_PAT`, exactly like `release.yml`. The only thing using the workflow token is `actions/checkout`, so `contents: read` is sufficient there and `write-all` buys nothing.

One correction to the outside-contributions point, because it changes which runs matter. For pull requests from forks GitHub issues a read-only `GITHUB_TOKEN` regardless of what the `permissions` key says, unless the org has explicitly enabled write tokens for fork PRs. So `write-all` in `ci.yml` is not what an outside contributor's PR run gets. Where it does apply is push and tag runs on the repo itself, which is also where `cargo-bins/cargo-binstall@main`, `denoland/setup-deno@main`, `taiki-e/install-action@nextest` and `dtolnay/rust-toolchain@nightly` execute from mutable refs. That is the CVE-2025-30066 shape precisely: the compromise landed by moving tags on an action, and every one of those four resolves at run time.

I'll open one PR with the permission scoping and a second with the SHA pins, since the pins are noisy to review and better not tangled with a behavioural change. Anything you'd rather see done differently, or any action you'd prefer left on a floating ref, say so and I'll work around it.

PR #1984, opened 2026-05-17
Add bindings for V8 inspector async-task hooks and debug break-on-next-call

bartlomieju:
Adds two related batches of debugger-side bindings that Deno (and any
other embedder porting Node features) needs to drive async stack
traces and `--inspect-brk` / `vm.Script({ breakOnFirstLine: true })`
correctly.

**V8Inspector async-task and idle hooks.** V8 already tracks async
stack traces internally for sources it owns (promises), but anything
scheduled outside V8 — Node's setTimeout / setInterval, host I/O
completion callbacks, etc. — has to call these hooks itself or the
resulting Debugger.paused notifications come back with an empty
asyncStackTrace. Exposes V8Inspector::idleStarted / idleFinished,
asyncTaskScheduled, asyncTaskCanceled, asyncTaskStarted,
asyncTaskFinished, and allAsyncTasksCanceled. The async-task methods
are marked unsafe because the void* task pointer is used by V8 as an
opaque identity key — callers must keep the identity stable across
the schedule/start/finish/cancel sequence.

**Isolate::set_break_on_next_function_call /
clear_break_on_next_function_call.** Wraps v8::debug::
SetBreakOnNextFunctionCall / ClearBreakOnNextFunctionCall, the entry
points Node uses to make --inspect-brk and `vm.Script({
breakOnFirstLine: true })` pause at the first user-visible statement
of the script rather than at any embedder wrapper frame above it. The
symbols live behind V8_EXPORT_PRIVATE in v8/src/debug/debug-interface.
h — outside the public include path but still exported from the V8
static archive — so the shim forward-declares them in v8::debug
rather than pulling the private header chain into binding.cc.

Includes smoke tests for both that exercise the C++ shim paths.
End-to-end behavior (Debugger.paused carrying asyncStackTrace, pause-
on-first-line landing on the user code line) is best verified against
the embedder; I have not run V8_FROM_SOURCE locally, so CI is the
first place the new C++ shims actually compile.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -3417,6 +3417,26 @@ void v8_inspector__V8Inspector__allAsyncTasksCanceled(
   self->allAsyncTasksCanceled();
 }
 
+// Forward declarations from v8/src/debug/debug-interface.h. These functions
+// are marked V8_EXPORT_PRIVATE — they're not in the public include/ path but
+// they are exported from the V8 static archive that embedders link against,
+// and they're the official way for embedders (Node, Chromium) to implement
+// "break on next function call" / `vm.Script({ breakOnFirstLine: true })`.
+namespace v8 {
+namespace debug {
+V8_EXPORT_PRIVATE void SetBreakOnNextFunctionCall(v8::Isolate* isolate);
+V8_EXPORT_PRIVATE void ClearBreakOnNextFunctionCall(v8::Isolate* isolate);
+}  // namespace debug
+}  // namespace v8
+
+void v8__debug__SetBreakOnNextFunctionCall(v8::Isolate* isolate) {
+  v8::debug::SetBreakOnNextFunctionCall(isolate);
+}
+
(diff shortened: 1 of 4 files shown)


src/binding.cc
```diff
@@ -3383,6 +3383,65 @@ void v8_inspector__V8Inspector__contextDestroyed(
...
+}
+
+// Forward declarations from v8/src/debug/debug-interface.h. These functions
+// are marked V8_EXPORT_PRIVATE — they're not in the public include/ path but
+// they are exported from the V8 static archive that embedders link against,
+// and they're the official way for embedders (Node, Chromium) to implement
+// "break on next function call" / `vm.Script({ breakOnFirstLine: true })`.
+// Wrapped in `extern "C++"` because this whole region is inside an
+// `extern "C"` block — without this, the namespaced declarations would
+// inherit C linkage and the linker would look for unmangled symbols.
```
devsnek: no, they use the actual public api: `schedulePauseOnNextStatement`

PR #1994, opened 2026-05-29
fix: pass `flags` to `v8__ObjectTemplate__SetIndexedPropertyHandler` FFI

popematt:
The FFI declaration for `SetIndexedPropertyHandler` omitted the `flags` parameter that the C++ binding accepts, causing any flags set on `IndexedPropertyHandlerConfiguration` (e.g., `NON_MASKING`) to be silently discarded. I think it also results in UB because we're calling a C/C++ function with fewer args than expected... and the Rust compiler cannot detect because we correctly called the binding function. The problem is that the `extern "C"` binding function does not match the actual signature of the C++ function.

The named handler's declaration already passes flags correctly—this aligns the indexed handler to match.

Added a test that sets `NON_MASKING` on an indexed handler and verifies real indexed elements take priority over the interceptor.

For reference, here is the C++ function that we're trying to match:
https://github.com/denoland/rusty_v8/blob/5d0e31ea6bf67f4559faa759b91e22bc3f1cd696/src/binding.cc#L1505-L1517

### DIFF
--- a/src/template.rs
+++ b/src/template.rs
@@ -149,6 +149,7 @@ unsafe extern "C" {
     definer: Option<IndexedPropertyDefinerCallback>,
     descriptor: Option<IndexedPropertyDescriptorCallback>,
     data_or_null: *const Value,
+    flags: PropertyHandlerFlags,
   );
 
   fn v8__ObjectTemplate__SetImmutableProto(this: *const ObjectTemplate);
@@ -1006,6 +1007,7 @@ impl ObjectTemplate {
         configuration.definer,
         configuration.descriptor,
         configuration.data.map_or_else(null, |p| &*p),
+        configuration.flags,
       );
     }
   }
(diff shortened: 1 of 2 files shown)


bartlomieju:
Can you explain how this is an undefined behavior? The declaration in C++ has flags as optional arg:
https://github.com/denoland/v8/blob/73d19698991616a34a00ca691a6e697dbb69e2ef/include/v8-template.h#L929

aapoalas:
> Can you explain how this is an undefined behavior? The declaration in C++ has flags as optional arg

Long time no see! Hopping in from the sidelines here, but I'm quite sure the C++ optional/default values are just syntax sugar for an automatic fill-in of the default on the caller side. The callee does not change its behaviour or code generation one bit, ie. the parameter is still required in FFI, only C++ callers get to omit it.

bartlomieju (approved):
That makes sense, thanks!

PR #1996, opened 2026-05-31
feat: add `no_icu` feature to build V8 without i18n/ICU support

bartlomieju:
Adds an opt-out `no_icu` cargo feature that builds V8 with
`v8_enable_i18n_support=false`. This drops the embedded ICU data (around
10MB) and the `Intl` API from the resulting static library, producing a
slimmer build for consumers that do not need internationalization. The
motivation is a slim `denort` variant for the Deno CLI, where the ICU and
`Intl` surface is rarely used.

The feature is an opt-out rather than a default-on `icu` feature on
purpose. A default-on feature would silently drop ICU for every consumer
that sets `default-features = false` (which the Deno CLI does), and it
would change existing prebuilt archive names. Modelling it as `no_icu`
keeps all current consumers and their archive names unchanged, and only
the slim build opts in.

A new `rusty_v8_enable_i18n` GN arg gates the ICU bindings in
`src/binding.cc`. When i18n is disabled, `icu_get_default_locale` and
`icu_set_default_locale` fall back to stub implementations that report a
fixed `en-US` tag and ignore locale changes, so the static library still
links without ICU. `set_common_data_77` and its underlying ICU symbol are
compiled out, since there is no ICU to hand data to. The ICU-specific
tests are gated behind `not(feature = "no_icu")`.

The prebuilt download suffix gains a `_noicu` component. CI builds a
`simdutf + no_icu` release variant across the targets the Deno CLI ships
(including the separate `aarch64-pc-windows-msvc` cross-compile job), so a
`librusty_v8_simdutf_noicu_release_<target>.a` archive is published. This
is the exact feature combo the CLI consumes (it always builds with
simdutf), so it does not expand the matrix into a full cross product.

Note: this is a prototype/experiment. The slim deno consumer plumbing (a
`no_icu` cargo feature threaded through `deno_core`/`deno_runtime` and a
`denort-slim` target) lives in the Deno repo and consumes the archives
produced here.

A possible future cleanup is to make simdutf a separately linkable
archive so it stops being a build-matrix axis: denoland/rusty_v8#1997.


bartlomieju:
Closing for now. Note this leaves no i18n/ICU opt-out in the tree — #1833 was closed in favour of this PR, so if the slim-build motivation comes back, this is the branch to revive rather than starting over.

For whoever picks it up: the diff has three files that shouldn't be committed — `tools/__pycache__/v8_deps.cpython-311.pyc`, `gen/src_binding_release_aarch64-apple-darwin.tmp`, and a stray `empty.js` at the repo root. Everything else was green across the CI matrix.

PR #1998, opened 2026-06-01
build: require runtime ICU data files

divybot:
Closes denoland/divybot#459

PR #1999, opened 2026-06-01
build: keep ICU data out of the V8 static library

bartlomieju:
The 14.9 bump (12d7ef1) added icu_use_data_file = false to .gn, which
embeds the ICU data blob (icudt*_dat, ~10MB) directly into the V8 static
library. That made ICU data an inseparable part of libv8 and added a
roughly 11MB regression to downstream binary sizes. It also makes it
impossible to experiment with alternative data sets (small-icu, no data)
without rebuilding V8.

This restores icu_use_data_file to ICU's default of true, so the data is
no longer compiled in and the embedder supplies it at runtime via
v8::icu::set_common_data (or by loading icudtl.dat). Keeping ICU data as
a separate, swappable artifact is what lets downstreams choose full-icu,
small-icu, or no data at all.

mksnapshot initializes ICU while generating the snapshot, so it needs
icudtl.dat present at build time even though the data is not embedded
into the final library. To satisfy that, v8_depend_on_icu_data_file and
icu_copy_icudata_to_root_build_dir are aligned with icu_use_data_file:
the copy_icudata target stages the file in root_out_dir and mksnapshot
depends on it. The staged file is a build-time artifact only and is not
linked into libv8. Without this the build fails with "mksnapshot: Failed
to initialize ICU".

Verified with a from-source build (V8_FROM_SOURCE=1) on macOS arm64: the
library builds, icudtl.dat is not embedded, and the icu_date, icu_format,
icu_collator and get_default_locale tests pass once set_common_data is
called in the test harness.

Downstream: denoland/deno#34681 re-enables deno_core's include_icu_data
to consume this (verified end-to-end against a from-source build of this
branch).

### DIFF
--- a/.gn
+++ b/.gn
@@ -78,9 +78,6 @@ default_args = {
 
   use_relative_vtables_abi = false
 
-  v8_depend_on_icu_data_file = false
-  icu_copy_icudata_to_root_build_dir = false
-
   # Keep ICU data OUT of the V8 static library. With `icu_use_data_file = true`
   # (ICU's default) the data is not embedded as the `icudt*_dat` symbol; the
   # embedder must supply it at runtime via `v8::icu::set_common_data` (or by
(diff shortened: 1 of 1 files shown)


bartlomieju:
Closing for now. Recording the state so this isn't lost:

`.gn` still has `icu_use_data_file = false` (plus `v8_depend_on_icu_data_file` and `icu_copy_icudata_to_root_build_dir` false), so the ~10MB `icudt*_dat` blob introduced by the 14.9 bump (12d7ef1) remains embedded in the static library and in every downstream binary. This branch is the fix if that's revisited — it was 28/29 green, and the inline comments explain the mksnapshot coupling and why the flags must stay aligned on a V8 bump.

Worth noting it would be a breaking change for embedders: with the data no longer embedded, they'd need `icu::set_common_data_78` or a shipped `icudtl.dat`.

#1998 was closed in favour of this one.

PR #2013, opened 2026-07-05
Fixes typo and emphasizes the role of the V8_FROM_SOURCE env var

gabrielfalcao:
Hi folks,

I was building rusty_v8 from scratch, ran into issues, sook help in the readme and got a bit confused about the role of the V8_FROM_SOURCE because of what appears to be a typo. I made a small change to address the confusing statement. I hope this helps the next one doing the same thing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,8 +49,8 @@ a prebuilt version of V8. We publish static libs for every version of rusty v8
 on [Github](https://github.com/denoland/rusty_v8/releases).
 
 Binaries builds are turned on by default: `cargo build` will initiate a download
-from github to get the static lib. To disable this build using the
-`V8_FROM_SOURCE` environmental variable.
+from github to get the static lib. To disable this, set the environmental variable
+`V8_FROM_SOURCE=0`. To explicitly enable building from source, set `V8_FROM_SOURCE=1`.
 
 When making changes to rusty_v8 itself, it should be tested by build from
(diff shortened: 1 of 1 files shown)


bartlomieju:
Thanks can you sign the CLA and run formatting so we can merge it?

gabrielfalcao:
> Thanks can you sign the CLA and run formatting so we can merge it?

I just signed.

gabrielfalcao:
Hey folks, just as FYI, I already signed the CLA

bartlomieju:
Thanks for the patience here, and sorry for the delay — the CLA is confirmed signed.

I pushed a commit to your branch rather than send you round again. Two things:

The original sentence was indeed broken, but `V8_FROM_SOURCE=0` doesn't do what the replacement claimed. `build.rs` reads it through `env_bool`, which only accepts `"1"`, `"true"` and `"yes"` — so `0` is indistinguishable from leaving the variable unset, and both keep the prebuilt download. There's no disable knob, just one opt-in switch, so I reworded it to describe that directly.

I also wrapped the paragraph at 80 columns, which was the formatting nit from earlier. Note that running `deno fmt` over the whole README reflows a couple of unrelated sections that landed unformatted, so I kept the change to just this paragraph to avoid the noise.

Should be good to merge once CI is green.

bartlomieju (approved):

PR #2017, opened 2026-07-15
build: Clean up downloading process

TheJanzap:
This PR cleans up the pre-built archive download functionality in `build.rs`. I originally worked on this because on a fresh install of Windows, none of the downloaders are available and the build fails.

I implemented the following changes:
- `python()` now checks `PATH` for `python` and `python3` if the `PYTHON` env var doesn't exist or is empty. Additionally, it checks whether the recieved path from those methods is a valid executable. For most calls, I used `expect()`, as the build from source requires Python to be present. However, just downloading the prebuilt archive does not.
- Refactored `download_file()` to more cleanly show in what order the downloaders are executed. The execution of the downloaders has been moved to seperate functions
- Downloads with curl now check if `curl` is present. ~This also uses `curl.exe` instead of the [alias for `Invoke-WebRequest`](https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/invoke-webrequest?view=powershell-5.1#notes) on the preinstalled PowerShell 5.1 on Windows.~

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -180,14 +180,12 @@ fn build_binding() {
     );
   }
 
-  let output = Command::new(
-    python().expect("Python must be available to build rusty_v8!"),
-  )
-  .arg("./tools/get_bindgen_args.py")
-  .arg("--gn-out")
-  .arg(build_dir().join("gn_out"))
-  .output()
-  .unwrap();
+  let output = Command::new(python_or_die())
+    .arg("./tools/get_bindgen_args.py")
+    .arg("--gn-out")
+    .arg(build_dir().join("gn_out"))
+    .output()
+    .unwrap();
   let args = String::from_utf8(output.stdout).unwrap();
   let args = args.split('\0').collect::<Vec<_>>();
 
@@ -632,12 +630,7 @@ fn build_v8(is_asan: bool) {
 fn print_gn_args(gn_out_dir: &Path) {
   assert!(
     Command::new(gn())
-      .arg(format!(
-        "--script-executable={}",
-        python()
-          .expect("Python must be available to build rusty_v8!")
-          .display()
-      ))
+      .arg(format!("--script-executable={}", python_or_die().display()))
       .arg("args")
       .arg(gn_out_dir)
       .arg("--list")
@@ -666,14 +659,12 @@ fn maybe_install_sysroot(arch: &str) {
   let sysroot_path = format!("build/linux/debian_sid_{arch}-sysroot");
   if !PathBuf::from(sysroot_path).is_dir() {
     assert!(
-      Command::new(
-        python().expect("Python must be available to build rusty_v8!")
-      )
-      .arg("./build/linux/sysroot_scripts/install-sysroot.py")
-      .arg(format!("--arch={arch}"))
-      .status()
-      .unwrap()
-      .success()
+      Command::new(python_or_die())
+        .arg("./build/linux/sysroot_scripts/install-sysroot.py")
+        .arg(format!("--arch={arch}"))
+        .status()
+        .unwrap()
+        .success()
     );
   }
 }
@@ -690,15 +681,13 @@ fn download_ninja_gn_binaries() {
 
   if !gn.exists() || !ninja.exists() {
     assert!(
-      Command::new(
-        python().expect("Python must be available to build rusty_v8!")
-      )
-      .arg("./tools/ninja_gn_binaries.py")
-      .arg("--dir")
-      .arg(&target_dir)
-      .status()
-      .unwrap()
-      .success()
+      Command::new(python_or_die())
+        .arg("./tools/ninja_gn_binaries.py")
+        .arg("--dir")
+        .arg(&target_dir)
+        .status()
+        .unwrap()
+        .success()
     );
   }
   assert!(gn.exists());
@@ -715,13 +704,11 @@ fn download_ninja_gn_binaries() {
 
 fn download_rust_toolchain() {
   assert!(
-    Command::new(
-      python().expect("Python must be available to build rusty_v8!")
-    )
-    .arg("./tools/rust_toolchain.py")
-    .status()
-    .unwrap()
-    .success()
+    Command::new(python_or_die())
+      .arg("./tools/rust_toolchain.py")
+      .status()
+      .unwrap()
+      .success()
   );
 }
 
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -789,63 +798,14 @@ fn download_file(url: &str, filename: &Path) {
...
-            .status()
-            .unwrap()
-        }
-      }
-    }
-  };
+  let status = download_with_deno(url, &tmpfile)
+    .or_else(|_| download_with_python(url, &tmpfile))
+    .or_else(|_| download_with_curl(url, &tmpfile))
+    .expect("Neither deno, python nor curl were available to download the V8 prebuilt archive.");
```
bartlomieju, line 842: This is the blocker: it removes the fallback chain rather than refactoring it.

`or_else` only fires on `Err`, but `Command::status()` returns `Ok(status)` for a *failed* exit code. So if deno is installed and runs but the download fails — non-2xx response hitting `Deno.exit(1)`, a TLS error, a truncated body — this returns `Ok(<non-success status>)` and we never try Python or curl. We go straight to the `if !status.success()` panic below.

The code on `main` handles exactly this, deliberately: `.filter(|s| s.success())` on the deno arm and `Ok(status) if status.success()` on the Python arm. The Python case is the one that bites in practice — a machine where `download_file.py` exists but fails on certificates currently falls back to curl and succeeds; after this change it hard-fails.

So the chain needs to treat "ran but exited non-zero" as a failure too, not just "couldn't spawn". Something like giving each helper an `Option<ExitStatus>` return (`None` for both spawn failure and unsuccessful exit) and chaining with `.or_else(|| ...)`, then `.expect(...)` at the end. That keeps the readability win you were after without changing behaviour.

build.rs
```diff
@@ -1207,12 +1238,28 @@ fn gn() -> String {
...
+  if let Ok(python_path) = env::var("PYTHON") {
+    return which(python_path).map_err(|_| {
+      io::Error::new(
+        io::ErrorKind::InvalidInput,
+        "Path in PYTHON environment variable does not point to an executable!",
+      )
+    });
+  }
+
+  which("python").or_else(|_| which("python3")).map_err(|_| {
```
bartlomieju: Please try `python3` before `python`. On plenty of Linux distros and inside some conda/virtualenv setups `python` is still Python 2, and this feeds `--script-executable` for GN plus every Chromium build script, all of which require Python 3. Preferring `python` first turns a working build into a confusing py2 syntax error.

With the order flipped, the README change below also needs revisiting — the macOS `PYTHON=python3` note was there because macOS doesn't ship `python` as Python 3, and `python3`-first is what makes it safe to drop.

build.rs
```diff
@@ -868,6 +828,75 @@ fn download_file(url: &str, filename: &Path) {
...
+}
+
+/// Downloads a file from `url` with curl and stores it at `path`.
+fn download_with_curl<P: AsRef<std::ffi::OsStr>>(
+  url: &str,
+  path: P,
+) -> io::Result<std::process::ExitStatus> {
+  println!("Downloading with curl...");
+
+  // Using `which` here also allows us to use `curl.exe` instead of the
```
bartlomieju: The rationale in this comment isn't right. `Command::new("curl")` doesn't go through PowerShell at all — it does a direct `CreateProcess` against `PATH`, so the `curl` → `Invoke-WebRequest` alias never applies here. That alias only affects commands typed into a PowerShell session.

Going through `which` is harmless and it does let us report "curl not found" instead of panicking, which is the real win — but could you reword the comment to say that rather than citing the alias?

build.rs
```diff
@@ -613,12 +616,14 @@ fn maybe_install_sysroot(arch: &str) {
...
   if !PathBuf::from(sysroot_path).is_dir() {
     assert!(
-      Command::new(python())
-        .arg("./build/linux/sysroot_scripts/install-sysroot.py")
-        .arg(format!("--arch={arch}"))
-        .status()
-        .unwrap()
-        .success()
+      Command::new(
+        python().expect("Python must be available to build rusty_v8!")
```
bartlomieju: This `.expect("Python must be available to build rusty_v8!")` is repeated at six call sites and forces an extra indent level at each one (see also lines 153, 644, 669, 1114, and the `format!` in `print_gn_args`/`run_gn_gen`).

A small `fn python_or_die() -> PathBuf { python().expect(...) }` would keep all of these as one-liners and leave `Command::new(python_or_die())` reading the way it does on `main`.

bartlomieju (changes requested):
Thanks for digging into this — the underlying problem is real. On a fresh Windows install `python()` unconditionally returns `"python3"`, which doesn't exist there, and then the curl fallback does `Command::new("curl").status().unwrap()`, so a missing curl panics with a bare "program not found" instead of our helpful "set `V8_FROM_SOURCE=1` / check your network" message. Probing `PATH` for the Python binary and giving a clear error when no downloader exists are both worth having.

As written though I can't merge it: the `or_else` chain silently drops the download fallbacks (details inline), which would regress builds that work today. There are also a couple of smaller things about the Python lookup order and the curl rationale.

What I'd like to see is a tighter diff scoped to the actual bug:

1. `python() -> io::Result<PathBuf>` probing `PYTHON` → `python3` → `python`.
2. A `python_or_die() -> PathBuf` helper so the `V8_FROM_SOURCE` call sites stay as short as they are today instead of growing a six-times-repeated `.expect(...)` and an extra rustfmt indent level.
3. In `download_file`, keep the existing success-filtered fallback structure and just swap the final curl `unwrap()` for the clear panic message.

That fixes fresh-Windows without changing download semantics. Worth noting that CI being green here doesn't tell us much — the runners have deno, python and curl all present, so none of the changed paths are actually exercised.

build.rs
TheJanzap: Just for clarity: Do current MacOS versions provide Python 3 as `python`? I have now changed the lookup order to `PYTHON` -> `python3` -> `python`. The macOS note in the README shouldn't be necessary anymore with this change, correct?

bartlomieju: Good catch, you're right on both counts. Apple removed `/usr/bin/python` entirely back in macOS 12.3, so a stock macOS box has python3 (the Xcode CLT shim) and no bare python at all; Homebrew's formula only links `python3` too. With `python3` tried first, macOS resolves cleanly with no env var, so yes — the note can go. Thanks for flipping the order.

Two tiny follow-ups while you're in here:

- The doc comment still reads 2. Look for `python` or `python3` in `PATH` — the order is backwards relative to the code now.
- `README.md:136` says the build scripts "require Python 3 to be available as python3 in your PATH". Since we fall back to python now, maybe reword to
something like "require Python 3, located via the PYTHON environment variable, or as python3 or python in your PATH".

build.rs
TheJanzap: I was a bot unsure on how to handle this without writing too much code. In its new form, `download_with_python` and `download_with_deno` check `.filter(|s| s.success())`, while `download_with_curl` does not - due to it being the last one in the chain (and it is also being called by another part of the build script). 

When curl exists but the download fails, it is passed to the "build from source" message. When curl is not installed, the "no downloader" message appears. 

But a new edge case is present: If deno or python return with an unsucessful error code and curl is not installed, the "no downloader" message also appears, even if it were better to show the "build from source" message.

TheJanzap:
Unsure if this is still needed with the changes in #2036 and #2066

PR #2020, opened 2026-07-17
perf(string): ASCII fast path for String::new_from_utf8

bartlomieju:
## Summary

`String::new_from_utf8` (and thus `String::new`) always went through V8's
`NewFromUtf8`, which validates the UTF-8 and determines the one-byte/two-byte
representation. Pure-ASCII UTF-8 is byte-identical to its Latin-1 (one-byte)
form, so we can skip all of that and call `NewFromOneByte` directly after a
cheap `is_ascii()` SIMD check.

Since `String::new`/`new_from_utf8` is how essentially every Rust string is
handed to V8, this speeds up ASCII string creation — identifiers, property
keys, paths, error messages, the overwhelmingly common case.

## Benchmark

`benches/string_encode.rs` (added), `new_from_utf8` on ASCII input, ns/op:

| length | before | after | delta |
| ------ | ------ | ----- | ----- |
| 8      | 23.4   | 7.8   | **-67%** |
| 16     | 15.1   | 7.0   | **-54%** |
| 32     | 12.7   | 7.3   | **-43%** |
| 64     | 13.0   | 11.1  | -15% |

## Correctness / no regression

- Non-ASCII input keeps the original `NewFromUtf8` path, and `is_ascii`
  short-circuits on the first non-ASCII byte, so there is no measurable
  regression for non-ASCII strings.
- Round-trip correctness verified for ASCII, Latin-1, two-byte (emoji), and
  mixed ASCII/CJK inputs (see the bench's correctness gate).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -150,5 +150,10 @@ name = "function"
 path = "benches/function.rs"
 harness = false
 
+[[bench]]
+name = "string_encode"
+path = "benches/string_encode.rs"
+harness = false
+
 [workspace]
 members = ["examples/android"]
--- a/benches/string_encode.rs
+++ b/benches/string_encode.rs
@@ -1,0 +1,70 @@
+// Measures Rust->V8 string creation: new_from_utf8 (validates UTF-8 + picks
+// one-byte/two-byte) vs new_from_one_byte (skips validation) for ASCII input.
+use std::time::Instant;
+
+fn main() {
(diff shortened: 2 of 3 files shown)


bartlomieju:
Superseded by the current-main version which does the full simdutf transcode (ASCII + Latin-1 + UTF-16), not just the ASCII half. This PR was also based on a stale main (v146).

PR #2021, opened 2026-07-17
perf(string): fast path for short one-byte to_rust_string_lossy

bartlomieju:
## Summary

`to_rust_string_lossy` always calls `utf8_length` (a full FFI scan of the
string) before allocating and writing. For short one-byte (Latin-1) strings we
can instead write the raw one-byte content once and determine in Rust whether
it's already valid UTF-8 (pure ASCII) or needs Latin-1 transcoding — skipping
the separate `utf8_length` FFI round-trip.

An ASCII string now costs one FFI write + a cheap Rust `is_ascii` check, instead
of two full FFI scans (`utf8_length` + write).

## Benchmark

`benches/string_conversion.rs` (added), `to_rust_string_lossy`, ns/op:

| case | before | after | delta |
| ---- | ------ | ----- | ----- |
| short ASCII (11) | 18.1 | 13.4 | **-27%** |
| ASCII 32 | 17.6 | 15.2 | -14% |
| ASCII 64 | 17.7 | 16.3 | -8% |
| ASCII 128 | 19.9 | 19.4 | -2% |

## Gating / trade-offs

- The win is a fixed-cost FFI-call saving, so it's largest for short strings and
  fades as per-byte work grows. Measured crossover is past ~128 UTF-16 code
  units, so the fast path is **gated to `len <= 128`**; longer strings keep the
  exact-sized general path (no regression there).
- Short **non-ASCII Latin-1** strings do an extra transcode allocation and are
  somewhat slower. Such strings (accented Western text with no two-byte chars)
  are uncommon relative to ASCII on this boundary, and in absolute terms the
  per-string ASCII saving (~5 ns) is comparable to the Latin-1 cost — net
  positive whenever ASCII outnumbers accented-Latin, which it heavily does.

## Correctness

Round-trip verified for ASCII, Latin-1, two-byte (emoji/CJK), and empty inputs
(see the bench's correctness gate).

Independent of #2020 (encode fast path); either can land first.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -150,5 +150,10 @@ name = "function"
 path = "benches/function.rs"
 harness = false
 
+[[bench]]
+name = "string_conversion"
+path = "benches/string_conversion.rs"
+harness = false
+
 [workspace]
 members = ["examples/android"]
--- a/benches/string_conversion.rs
+++ b/benches/string_conversion.rs
@@ -1,0 +1,49 @@
+// Benchmark + correctness check for String::to_rust_string_lossy across the
+// representations it special-cases: short/long ASCII (one-byte), Latin-1
+// (one-byte non-ASCII), and two-byte (UTF-16).
+use std::time::Instant;
+
+fn main() {
+  let platform = v8::new_default_platform(0, false).make_shared();
+  v8::V8::initialize_platform(platform);
+  v8::V8::initialize();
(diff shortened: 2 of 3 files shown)


bartlomieju:
Closing as obsolete: this optimized the old `utf8_length`-based `to_rust_string_lossy`, but on current main (v150.2.0) that function is already ValueView-based, so the change no longer applies. The ValueView path is being optimized separately (see the string-boundary proposals).

PR #2023, opened 2026-07-17
perf: drop redundant String::Length() FFI call on ValueView conversions

bartlomieju:
## Summary

`to_rust_string_lossy`, `write_utf8_into`, and `to_rust_cow_lossy` each called
`self.length()` — an out-of-line `v8::String::Length()` FFI call — purely to
early-return on empty strings, even though the `ValueView` they build a line
later already reports the length. This removes those redundant calls and lets
the empty case fall out of the `ValueView` path:

- `data()` now returns an empty slice for a zero-length string. Empty strings
  may carry a null `data8_`/`data16_` pointer, so this also avoids passing null
  to `from_raw_parts`.
- The ASCII arm then turns the empty slice into an empty `String` / borrowed
  empty `&str`.

## Impact

Incremental over #2022 (interleaved, median of 6), `to_rust_string_lossy`:

| kind | n=4 | n=16 | n=64 | n=256 | n=4096 |
| --- | --- | --- | --- | --- | --- |
| ascii | −2.6% | −6.5% | −4.5% | −5.2% | +0.6% |
| latin1 | −5.6% | −4.4% | −6.1% | −2.6% | −0.4% |
| twobyte | −7.3% | −2.5% | +0.2% | −0.7% | +0.5% |

A steady ~3–6% across short-to-medium strings — a redundant length lookup on
every conversion, gone. Fades to noise once copy/transcode cost dominates.

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -964,10 +964,9 @@ impl String {
   /// data. When the `simdutf` feature is enabled, uses SIMD-accelerated
   /// transcoding for Latin-1 and two-byte strings.
   pub fn to_rust_string_lossy(&self, scope: &Isolate) -> std::string::String {
-    if self.length() == 0 {
-      return std::string::String::new();
-    }
-
+    // No preliminary `self.length()` FFI call: the `ValueView` reports the
+    // length, and `data()` yields an empty slice for empty strings, which the
+    // ASCII arm below turns into an empty `String`.
     // SAFETY: `self` is a valid V8 string reachable from a handle scope.
     let view = unsafe { ValueView::new_from_ref(scope, self) };
 
@@ -999,11 +998,8 @@ impl String {
     buf: &mut std::string::String,
   ) {
     buf.clear();
-    let len = self.length();
-    if len == 0 {
-      return;
-    }
-
+    // No preliminary `self.length()` FFI call; an empty string yields an empty
+    // `data()` slice and leaves `buf` cleared.
     // SAFETY: `self` is a valid V8 string reachable from a handle scope.
     // The ValueView is dropped before we return.
     let view = unsafe { ValueView::new_from_ref(scope, self) };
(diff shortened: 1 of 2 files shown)


bartlomieju:
One minor semantic note on the `data()` change: the `length == 0` branch unconditionally returns `ValueViewData::OneByte(&[])` without checking `is_one_byte_`, so an empty *two-byte* string now reports as `OneByte` rather than `TwoByte`.

```rust
let s = v8::String::new_from_two_byte(scope, &[], v8::NewStringType::Normal).unwrap();
let view = v8::ValueView::new(scope, s);
matches!(view.data(), v8::ValueViewData::TwoByte(_)); // was true, now false
```

Contents are empty either way, so none of the in-crate conversions (`to_rust_string_lossy` / `write_utf8_into` / `to_rust_cow_lossy`) are affected — this is purely about `data()`'s public contract of reporting the actual encoding. If it's worth preserving, reading `is_one_byte_` in the empty branch too (one in-buffer byte, empty strings only) and returning the matching empty variant would keep it accurate; otherwise a one-line doc note that empty strings always report `OneByte` would set expectations. Non-blocking either way.

(Also worth confirming: this same early-return fixes a latent null-`from_raw_parts` for direct `ValueView::data()` callers on empty strings, which is a nice side effect.)

bartlomieju:
Good catch — fixed by reading `is_one_byte_` in the empty-length branch and returning the matching empty variant, rather than a hardcoded `OneByte(&[])`.

Interesting empirical wrinkle while adding a test for it: V8 canonicalizes *every* empty string to the one-byte empty string, so `new_from_two_byte(scope, &[], …)` actually comes back with `is_one_byte_ == true` — i.e. `data()` reports `OneByte` for it both before and after this change (proposal 1's version would too). So there's no observable regression in practice, but reading `is_one_byte_` is the accurate/robust thing regardless, and I added a `string_valueview` case documenting the canonicalization. Also confirmed the empty early-return fixes the latent null-`from_raw_parts` for direct `data()` callers, as you noted.

PR #2024, opened 2026-07-17
perf: SIMD one-byte ASCII detection for long strings

bartlomieju:
## Summary

The one-byte read paths chose their ASCII vs Latin-1 branch with std's inline
`is_ascii` SWAR scan. For long strings simdutf's wide-SIMD scan (AVX2/NEON) is
substantially faster, so gate on length (measured crossover ~128 bytes): use
simdutf above it, keep inline `is_ascii` below (where the simdutf FFI-call
overhead would lose).

Applied to **all three** one-byte read paths:
- `to_rust_string_lossy` — one `utf8_length_from_latin1` pass both detects ASCII
  and sizes the Latin-1 transcode.
- `write_utf8_into` / `to_rust_cow_lossy` — share an `onebyte_is_ascii` helper
  (`validate_ascii` above the threshold, `is_ascii` below) for the same
  decision.

## Impact

`to_rust_string_lossy` on long ASCII: **~−6% at 256 chars, ~−30% at 4096**
(large text / JSON / source decode). The same detection speedup applies to
`write_utf8_into` and `to_rust_cow_lossy`. Short strings and Latin-1/two-byte
content are unchanged.

## Correctness

Long-string test coverage added for all three paths (ASCII + Latin-1, exercising
the ≥128 simdutf branch). All `test_api` string tests pass with and without the
`simdutf` feature.

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -964,10 +964,9 @@ impl String {
   /// data. When the `simdutf` feature is enabled, uses SIMD-accelerated
   /// transcoding for Latin-1 and two-byte strings.
   pub fn to_rust_string_lossy(&self, scope: &Isolate) -> std::string::String {
-    if self.length() == 0 {
-      return std::string::String::new();
-    }
-
+    // No preliminary `self.length()` FFI call: the `ValueView` reports the
+    // length, and `data()` yields an empty slice for empty strings, which the
+    // ASCII arm below turns into an empty `String`.
     // SAFETY: `self` is a valid V8 string reachable from a handle scope.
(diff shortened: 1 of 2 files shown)


bartlomieju:
Good call — extended the same gated ASCII-detection to `write_utf8_into` and `to_rust_cow_lossy` in this PR (via a shared `onebyte_is_ascii` helper: `validate_ascii` above the ~128-byte threshold, inline `is_ascii` below). It wasn't in any other PR — I'd originally noted it as a follow-up, but there's no reason to split it. Added long-string test coverage (ASCII + Latin-1) for all three paths.

PR #2029, opened 2026-07-23
perf(string): SIMD ASCII detection in new_from_utf8, fused one-byte read paths

bartlomieju:
## Summary

Follow-up to the string-perf work in #2023–#2026. Those optimizations gave the **read** paths (`to_rust_string_lossy` etc.) SIMD ASCII detection and gave `TextDecoder`-style decoding a big win — but the **create** path (`new_from_utf8`) still used a scalar `is_ascii` scan, and the one-byte read paths still did a separate ASCII scan before copying. This PR closes that gap.

All changes are in `src/string.rs`, use existing simdutf bindings, touch no C++/binding code, and preserve behavior (verified by round-trip correctness checks over ASCII/Latin-1/two-byte/emoji/mixed inputs across sizes).

### 1. SIMD ASCII detection in `new_from_utf8`
The pure-ASCII fast path used `buffer.is_ascii()` (std SWAR). It now uses `onebyte_is_ascii`, the same helper the read paths use, which switches to `simdutf::validate_ascii` for long inputs.

Because `validate_ascii` scans the *whole* buffer even when the first byte is non-ASCII (unlike std `is_ascii`, which short-circuits), a naive switch regressed Latin-1 input −6…−9%. Added a **cheap 32-byte inline early-reject** before the SIMD scan so non-ASCII text bails immediately; pure ASCII still gets the fast wide scan.

### 2. Fused convert in the one-byte read paths (`to_rust_cow_lossy`, `to_rust_string_lossy`)
The one-byte arms did `onebyte_is_ascii(bytes)` (a full scan) *then* a memcpy — two passes. `convert_latin1_to_utf8` does the transcode and, for pure ASCII, is just a copy while implicitly detecting ASCII (`written == len`) in a single pass.
- `to_rust_cow_lossy`: fused when the worst-case 2× expansion fits the borrow buffer (so the convert can't overflow).
- `to_rust_string_lossy`: fused above a 4 KB threshold; the 2× worst-case allocation only pays off once the input is large enough, so mid-size strings keep the exact-length pre-scan.

## Benchmarks (Apple M-series, aarch64, best-of-5, isolated micro-bench per function)

Δ = improvement of this PR over stock `main` (v150.3.0).

**`new_from_utf8` (Rust bytes → V8 string):**
| content | 256 B | 4 KB | 64 KB |
|---|--:|--:|--:|
| ascii | **+22.8%** | **+49.2%** | **+45.1%** |
| latin1 | +0.7% | −3.4% | −0.6% (neutral) |

**`to_rust_cow_lossy` (V8 string → `Cow<str>`):**
| content | 256 B | 4 KB | 64 KB |
|---|--:|--:|--:|
| ascii | **+7.6%** | ~0 | ~0 |
| latin1 | **+17.4%** | **+4.7%** | **+4.4%** |

**`to_rust_string_lossy` (V8 string → owned `String`):**
| content | 256 B | 4 KB | 64 KB |
|---|--:|--:|--:|
| ascii | ~0 | **+18.4%** | **+17.8%** |
| latin1 | +2.0% | **+13.3%** | **+12.0%** |

No two-byte/emoji/mixed regressions (27 cases spanning the three functions: range −2.0%…+25.8%, worst −2.0% is noise).

## Notes for reviewers
- The `4096` fuse threshold and the `32`-byte early-reject window were tuned empirically on aarch64; they're conservative but a maintainer may want to confirm the crossover on x86_64. Change 1 (the headline win) has no new magic constant beyond the existing `ONEBYTE_SIMD_THRESHOLD` and is the least controversial if you'd prefer to land it alone.
- Verified locally: compiles against the v150.3.0 prebuilt and passes round-trip equality (`to_utf8`/`to_rust_string_lossy`/`to_rust_cow_lossy` == source) for every content/size combination benchmarked.

### DIFF
--- a/src/string.rs
+++ b/src/string.rs
@@ -1392,7 +1392,9 @@ unsafe fn transcode_latin1_to_utf8(
 ) -> usize {
   // SAFETY: the caller guarantees `out_ptr` is valid for `out_len` writes.
   let out = unsafe { std::slice::from_raw_parts_mut(out_ptr, out_len) };
-  crate::simdutf::convert_latin1_to_utf8(bytes, out)
+  // SAFETY: `out` covers the full UTF-8 expansion, so simdutf's write stays in
+  // bounds; it always produces valid UTF-8 from Latin-1 input.
+  unsafe { crate::simdutf::convert_latin1_to_utf8(bytes, out) }
 }
 
 /// Whether one-byte string data is pure ASCII. Uses simdutf's wide SIMD scan
--- a/src/string.rs
+++ b/src/string.rs
@@ -1149,10 +1149,8 @@ impl String {
           && bytes.len().saturating_mul(2) <= N
         {
           let written = unsafe {
-            let out = std::slice::from_raw_parts_mut(
-              buffer.as_mut_ptr() as *mut u8,
-              N,
-            );
+            let out =
+              std::slice::from_raw_parts_mut(buffer.as_mut_ptr() as *mut u8, N);
             crate::simdutf::convert_latin1_to_utf8(bytes, out)
           };
           // SAFETY: simdutf wrote `written` valid UTF-8 bytes into `buffer`.
--- a/src/string.rs
+++ b/src/string.rs
@@ -1148,10 +1148,10 @@ impl String {
         if bytes.len() >= ONEBYTE_SIMD_THRESHOLD
           && bytes.len().saturating_mul(2) <= N
         {
+          // SAFETY: `buffer` is valid for `N` writes and `N >= bytes.len() * 2`
+          // (guarded above), so it fits the full UTF-8 expansion.
           let written = unsafe {
-            let out =
-              std::slice::from_raw_parts_mut(buffer.as_mut_ptr() as *mut u8, N);
-            crate::simdutf::convert_latin1_to_utf8(bytes, out)
+            transcode_latin1_to_utf8(bytes, buffer.as_mut_ptr() as *mut u8, N)
           };
           // SAFETY: simdutf wrote `written` valid UTF-8 bytes into `buffer`.
           return unsafe {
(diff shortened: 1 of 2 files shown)


bartlomieju:
Reviewed the diff plus the surrounding functions and simdutf bindings. Memory safety holds throughout: the guards guarantee `out_len >= 2*len` (or exact `utf8_len`) into `transcode_latin1_to_utf8`, and simdutf officially supports exact-length allocation for `convert_latin1_to_utf8`, so the exact-length retained paths are sound. The `onebyte_is_ascii` head/tail split is per-byte-equivalent to a full scan. No correctness or memory-safety issues found.

A few things worth addressing:

**1. `onebyte_to_string` fused path (>=4096) retains 2x capacity for the string's lifetime.** It does `Vec::with_capacity(len*2)`, `set_len(written)`, and returns without shrinking. For pure ASCII `written == len`, so the returned `String` has `capacity == 2*len` with `len == len` — half the heap wasted and retained. `to_rust_string_lossy` on a retained 1 MB pure-ASCII string now holds a 2 MB buffer vs exactly 1 MB before (the replaced ASCII branch used `from_utf8_unchecked(bytes).to_owned()`, an exact-length alloc). This hits the most common input (JSON/source/identifiers) on a hot public API. The comments frame the 2x only as a one-time *allocation* cost to amortize and never mention the retained capacity — worth either calling it out explicitly as an accepted speed-for-memory tradeoff, or mitigating (e.g. `shrink_to_fit` when `written` is well under `cap`, accepting the realloc).

**2. Test gap: the 32-byte early-reject head/tail split is never exercised with a high byte past offset 32.** All Latin-1 units (`"café"`) put the non-ASCII byte in the first 4 bytes, so only the `!bytes[..32].is_ascii()` early-reject fires; the `validate_ascii(&bytes[head..])` remainder path is never validated against content that's ASCII for 32+ bytes then turns non-ASCII. This is the one branch where a bug could hide silently — an off-by-one in the `&bytes[head..]` offset would round-trip fine on the current corpus. Cheap to close: add a case like `"a".repeat(40) + "é"`.

**3. Test gap: boundary sizes land *at* thresholds, not below.** `unit.repeat(target / cpc + 1)` quantizes one-byte lengths to multiples of the unit char count (4 for `abcd`/`café`), so the "just below 32/128/4096" intent (targets 31/127/4095) actually produces 32/128/4096 bytes — the strictly-below cases don't exist, and the smallest one-byte string built is 4 bytes (sizes 1..3 untested). A low-side off-by-one in the `>=` threshold comparisons wouldn't be caught. Minor related note: the two-byte units `世界`/`🦕` exercise the wtf16 paths, not the one-byte thresholds the comment claims to straddle.

**4. Test gap: the cow owned-overflow paths are never reached.** With `N = 1<<16` dwarfing every input (max ~8 KB), the `bytes.len() > N` ASCII-owned branch and the `utf8_len > N` -> `latin1_to_cow_str` owned branch are dead. A sub-case with a small `N` (e.g. 256) would exercise them.

The optimization itself is correct and safe — the headline win (SIMD ASCII detection in `new_from_utf8`) is clean. I'd document the 2x retained-capacity tradeoff (1) and add the head/tail test case (2), which is the only gap covering an actual correctness claim.

PR #2037, opened 2026-07-29
Rolling to V8 15.2.124.1

nathanwhit:
Rolls the V8 submodule and README to V8 15.2.124.1.

This also fixes fork handling in `tools/auto_update_v8.ts`. The updater previously used `origin` for checkout and push, but hardcoded `denoland:autoroll` as the pull-request head. In a fork checkout, that combined the locally calculated title with an unrelated upstream branch.

The updater now:

- resolves the `denoland/rusty_v8` remote independently for the `main` base;
- derives the pull-request head owner from the actual push remote;
- uses explicit repository, base, and head arguments for `gh`;
- preserves and stages manual updater changes across its checkout; and
- prints command stderr when a subprocess fails.

Validation:

- `deno fmt --check tools/auto_update_v8.ts`
- `deno lint tools/auto_update_v8.ts`
- `deno check tools/auto_update_v8.ts`
- verified fork and GitHub Actions remote selection
- verified `c4ca1eccb90c5464d826b7713dc27178a53b0cfe` is the current `15.2-lkgr-denoland` head and reports V8 15.2.124.1
- verified V8 15.2 changes none of the dependency gitlinks pinned by upstream `main`

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -5,6 +5,7 @@
 #include <cstdint>
 #include <cstdio>
 #include <map>
+#include <memory>
 #include <mutex>
 #include <thread>
 
@@ -2246,13 +2247,39 @@ const v8::Value* v8__Context__GetContinuationPreservedEmbedderData(
   return local_to_ptr(value);
 }
 
-v8::MicrotaskQueue* v8__MicrotaskQueue__New(v8::Isolate* isolate,
-                                            v8::MicrotasksPolicy policy) {
-  return v8::MicrotaskQueue::New(isolate, policy);
+class RustyMicrotaskQueueHandle {
+ public:
+  RustyMicrotaskQueueHandle(v8::Isolate* isolate, v8::MicrotasksPolicy policy)
+      : queue_(v8::MicrotaskQueue::New(isolate, policy)) {}
+
+  v8::MicrotaskQueue* Get() const {
+#ifdef V8_CPPGC_MICROTASK_QUEUE
+    return queue_.Get();
+#else
+    return queue_.get();
+#endif
+  }
+
+ private:
+#ifdef V8_CPPGC_MICROTASK_QUEUE
+  cppgc::Persistent<v8::MicrotaskQueue> queue_;
+#else
+  std::unique_ptr<v8::MicrotaskQueue> queue_;
+#endif
+};
+
+RustyMicrotaskQueueHandle* v8__MicrotaskQueueHandle__New(
+    v8::Isolate* isolate, v8::MicrotasksPolicy policy) {
+  return new RustyMicrotaskQueueHandle(isolate, policy);
 }
 
-void v8__MicrotaskQueue__DESTRUCT(v8::MicrotaskQueue* self) {
-  self->~MicrotaskQueue();
+void v8__MicrotaskQueueHandle__DELETE(RustyMicrotaskQueueHandle* self) {
+  delete self;
+}
+
+v8::MicrotaskQueue* v8__MicrotaskQueueHandle__Get(
+    const RustyMicrotaskQueueHandle* self) {
+  return self->Get();
 }
 
 void v8__MicrotaskQueue__PerformCheckpoint(v8::Isolate* isolate,
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -1482,11 +1482,11 @@ void v8__ObjectTemplate__SetNativeDataProperty(
 
 void v8__ObjectTemplate__SetNamedPropertyHandler(
     const v8::ObjectTemplate& self, v8::NamedPropertyGetterCallback getter,
-    v8::NamedPropertySetterCallback setter,
+    v8::NamedPropertySetterCallbackV2 setter,
     v8::NamedPropertyQueryCallback query,
     v8::NamedPropertyDeleterCallback deleter,
     v8::NamedPropertyEnumeratorCallback enumerator,
-    v8::NamedPropertyDefinerCallback definer,
+    v8::NamedPropertyDefinerCallbackV2 definer,
     v8::NamedPropertyDescriptorCallback descriptor,
     const v8::Value* data_or_null, v8::PropertyHandlerFlags flags) {
   ptr_to_local(&self)->SetHandler(v8::NamedPropertyHandlerConfiguration(
@@ -1496,11 +1496,11 @@ void v8__ObjectTemplate__SetNamedPropertyHandler(
 
 void v8__ObjectTemplate__SetIndexedPropertyHandler(
     const v8::ObjectTemplate& self, v8::IndexedPropertyGetterCallbackV2 getter,
-    v8::IndexedPropertySetterCallbackV2 setter,
+    v8::IndexedPropertySetterCallback setter,
     v8::IndexedPropertyQueryCallbackV2 query,
     v8::IndexedPropertyDeleterCallbackV2 deleter,
     v8::IndexedPropertyEnumeratorCallback enumerator,
-    v8::IndexedPropertyDefinerCallbackV2 definer,
+    v8::IndexedPropertyDefinerCallback definer,
     v8::IndexedPropertyDescriptorCallbackV2 descriptor,
     const v8::Value* data_or_null, v8::PropertyHandlerFlags flags) {
   ptr_to_local(&self)->SetHandler(v8::IndexedPropertyHandlerConfiguration(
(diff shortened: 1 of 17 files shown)


ry (approved):
LGTM

bartlomieju (commented):
Converting my three earlier top-level comments into inline ones anchored to the relevant code, plus two smaller notes that were buried at the end of the crdtp comment. Deleting the originals.

src/binding.cc
```diff
@@ -2248,7 +2248,7 @@ const v8::Value* v8__Context__GetContinuationPreservedEmbedderData(
 
 v8::MicrotaskQueue* v8__MicrotaskQueue__New(v8::Isolate* isolate,
                                             v8::MicrotasksPolicy policy) {
-  return v8::MicrotaskQueue::New(isolate, policy).release();
+  return v8::MicrotaskQueue::New(isolate, policy);
```
bartlomieju: `MicrotaskQueue` is now cppgc-allocated, so this is handing back a GC-managed pointer rather than transferring ownership. Dropping `.release()` only compiles under `V8_CPPGC_MICROTASK_QUEUE`:

```c++
// include/v8-microtask-queue.h @ 15.2
#ifdef V8_CPPGC_MICROTASK_QUEUE
  static MicrotaskQueue* New(Isolate*, MicrotasksPolicy = kAuto);
  virtual void Trace(cppgc::Visitor*) const {}
#else
  V8_DEPRECATE_SOON("Use MicrotaskQueue allocated in cppgc, ...")
  static std::unique_ptr<MicrotaskQueue> New(Isolate*, MicrotasksPolicy = kAuto);
#endif
```

`v8_cppgc_microtask_queue` defaults to `true` in `gni/v8.gni` and is in `enabled_external_v8_defines`, so the define reaches our headers, and `build.rs` doesn't override it.

The Rust side still models this as owned memory, and the chain ends in an explicit destructor call:

- `MicrotaskQueue::new` returns `UniqueRef<Self>` (`src/microtask.rs:56`)
- `UniqueRef::drop` -> `drop_in_place` (`src/support.rs:131`)
- -> `impl Drop for MicrotaskQueue` -> `v8__MicrotaskQueue__DESTRUCT` (`src/microtask.rs:106`)
- -> `self->~MicrotaskQueue()` (two functions below this one, line 2254)

`MicrotaskQueue` has a `virtual ~MicrotaskQueue()`, so it is non-trivially-destructible and cppgc registers a finalizer -- the collector destructs it again on sweep. That is a double-destruction, and any use between the Rust drop and the sweep is a use-after-free.

The lifetime contract also inverted. The doc comment we carry over from V8 ("The embedder must keep the MicrotaskQueue instance alive until all associated Contexts are gone or detached") described `unique_ptr` ownership; under cppgc, liveness comes from the GC tracing the object, not from us holding a `UniqueRef`.

Suggested direction:

1. Return a non-owning handle from `MicrotaskQueue::new` instead of `UniqueRef<Self>`.
2. Delete `impl Drop for MicrotaskQueue` and `v8__MicrotaskQueue__DESTRUCT`.
3. Update the doc comment to describe GC-managed lifetime.
4. Add a regression test that creates a queue, drops the Rust handle, forces a GC, then runs a microtask checkpoint on a context using it -- under ASan that should catch the current behaviour.

Worth double-checking step 1 against how the queue stays alive across `Context::New` -- if nothing else traces it, we may need a cppgc `Persistent` rather than just dropping ownership on the floor.

tools/auto_update_v8.ts
```diff
@@ -71,32 +129,46 @@ readme = readme.replace(
...
+      "pr",
+      "list",
+      "--repo",
+      UPSTREAM_REPOSITORY,
+      "--state",
+      "open",
+      "--head",
+      prHead,
+      "--json",
+      "number",
```
bartlomieju: `gh pr list --head` does not accept an `owner:branch` prefix, so `prHead` never matches here. `pr create --head` (below) does accept it -- the two flags take different syntax, and `pr list` filters on `headRefName`.

Checked against this repo:

```console
$ gh pr list --repo denoland/rusty_v8 --state open --head "denoland:autoroll" --json number,headRefName
[]

$ gh pr list --repo denoland/rusty_v8 --state open --head "autoroll" --json number,headRefName
[{"headRefName":"autoroll","number":2037}]
```

So `openPrs` is always empty, the `pr edit` branch below is never taken, and we fall through to `pr create` -- which GitHub rejects because a PR for that head already exists. `run()` then prints the failure and `Deno.exit(1)`. Net effect: the second and every subsequent autoroll fails instead of retitling the open PR, which is exactly what the old `isPrOpen` check existed to handle.

```suggestion
      "--head",
      AUTOROLL_BRANCH,
      "--json",
      "number,headRepositoryOwner",
```

Since the bare branch name is ambiguous across forks, filter on `headRepositoryOwner.login === pushOwner` before taking `openPrs[0]`. `prHead` stays as-is for the `pr create` call.

Worth a manual end-to-end run against an already-open autoroll PR before landing -- this path isn't covered by the `deno check` / `deno lint` validation in the description.

src/crdtp_binding.cc
```diff
@@ -64,7 +55,8 @@ void crdtp__Serializable__AppendSerialized(const Serializable* self,
 }
 
 Dispatchable* crdtp__Dispatchable__new(const uint8_t* data, size_t len) {
-  return new Dispatchable(span<uint8_t>(data, len));
+  return new Dispatchable(span<uint8_t>(data, len), std::string_view(),
+                          FallthroughCallback());
```
bartlomieju: Hardcoding both new parameters away removes fall-through permanently rather than porting it. V8 15.2 didn't delete fall-through -- it moved it off `FrontendChannel` and onto a per-`Dispatchable` callback:

```c++
// third_party/inspector_protocol/crdtp/dispatch.h @ 15.2
using FallthroughCallback =
    std::function<void(int call_id, span<uint8_t> method,
                       span<uint8_t> serialized_message,
                       std::string_view associated_data)>;
```

With `FrontendChannelImpl::fall_through` and `crdtp__FrontendChannel__BASE__fallThrough` also deleted in this PR, fall-through becomes unreachable through this crate, and `associated_data` (which the new API threads through to handlers that opt into receiving it) is never settable.

To be clear on severity: this is **not** a crash. `UberDispatcher::Dispatch` guards the empty callback and degrades to a method-not-found response, which is why the new `assert_eq!(state.borrow().responses.len(), 3)` in `crdtp_domain_dispatcher_wire` passes:

```c++
  if (auto fallthrough = dispatchable.TakeFallthroughCallback()) {
    fallthrough(dispatchable.CallId(), dispatchable.Method(),
                dispatchable.Serialized(), std::string_view());
  } else {
    SendMethodNotFound(dispatchable.CallId(), method);
  }
```

It's a capability regression: an embedder that layered its own handling behind fall-through -- dispatching unknown methods to a second protocol layer instead of erroring -- has no equivalent after this roll, and the behaviour change is silent (an error response where a delegation used to happen).

The port would be to accept an optional Rust closure in `Dispatchable::new` and box it into the `FallthroughCallback`, mirroring how `FrontendChannel` already trampolines into `FrontendChannelImpl`. `associated_data` could stay empty if there's no caller for it.

If that's more than you want in a roll PR, fair -- but better landed before the next crates.io publish than after, and worth calling out in the release notes either way, since removing `fall_through` is a breaking change to a public trait.

src/crdtp.rs
```diff
@@ -736,12 +676,7 @@ unsafe extern "C" fn crdtp__DomainDispatcher__BASE__Dispatch(
     let dd = &mut *(rust_dispatcher as *mut DomainDispatcherData);
     let command = std::slice::from_raw_parts(command_data, command_len);
     let handle = DomainDispatcherHandle { ptr: dd.ptr };
-    let dispatchable_ref = if dispatchable.is_null() {
-      None
-    } else {
-      Some(&*dispatchable)
-    };
-    dd.imp.dispatch(command, dispatchable_ref, &handle)
+    dd.imp.dispatch(command, &*dispatchable, &handle)
```
bartlomieju: Minor: now that the probe phase is gone, `dispatchable` is always non-null, but the parameter is still `*const Dispatchable` and this dereferences it unconditionally. Making it a reference in the signature (or adding a debug assertion) would make the new invariant explicit.

tests/test_api.rs
```diff
@@ -13417,12 +13415,10 @@ fn crdtp_uber_dispatcher_basic() {
...
   let json = r#"{"id":1,"method":"Custom.unknownMethod","params":{}}"#;
   let cbor = v8::crdtp::json_to_cbor(json.as_bytes()).unwrap();
-  let dispatchable = v8::crdtp::Dispatchable::new(&cbor);
+  let mut dispatchable = v8::crdtp::Dispatchable::new(&cbor);
   assert!(dispatchable.ok());
 
-  let result = dispatcher.dispatch(&dispatchable);
-  assert!(!result.method_found());
-  result.run();
+  dispatcher.dispatch(&mut dispatchable);
```
bartlomieju, line 13448: This test no longer asserts anything -- the `method_found()` check went away with the API change and nothing replaced it, so it now only verifies that dispatch doesn't crash. Since the empty fallthrough callback makes V8 emit a method-not-found response, you could assert that directly: one response on the channel carrying error code `-32601`.

Same applies to `crdtp_dispatch_unregistered_domain` below (line 13840). `crdtp_domain_dispatcher_wire` got a response-count assertion -- these two could use the same treatment.

bartlomieju (approved):

PR #2038, opened 2026-08-01
feat: expose inspector value subtype callbacks

### LINKED ISSUE #2032: Expose V8InspectorClient value subtype callbacks
## Summary

V8 exposes two public embedder callbacks through
[`V8InspectorClient`](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/include/v8-inspector.h):

- `valueSubtype`
- `descriptionForValueSubtype`

`rusty_v8::inspector::V8InspectorClientImpl` does not currently expose these
callbacks.

As a result, an embedder cannot tell the Inspector that a host-backed V8 value
should use a specific `Runtime.RemoteObject` subtype and description. Such
values are presented as ordinary JavaScript objects instead.

## Impact

These callbacks affect how values are represented by the Inspector and rendered
by DevTools.

For example, an embedder may need to identify a host-backed object as a `node`
and provide a short description. Without the callbacks, the object remains
inspectable, but DevTools cannot apply subtype-specific presentation or
interactions.

This does not affect basic Inspector functionality such as evaluation,
breakpoints or property inspection.

V8's implementation also treats the two callbacks as a pair: for custom
subtypes, `descriptionForValueSubtype` is needed for the client-provided mirror
to be used. See
[`value-mirror.cc`](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/src/inspector/value-mirror.cc).

## Proposed API

Add two default methods to `V8InspectorClientImpl`:

```rust
fn value_subtype<'s>(
    &self,
    scope: &mut PinScope<'s, '_>,
    value: Local<'s, Value>,
) -> Option<UniquePtr<StringBuffer>> {
    None
}

fn description_for_value_subtype<'s>(
    &self,
    scope: &mut PinScope<'s, '_>,
    value: Local<'s, Value>,
) -> Option<UniquePtr<StringBuffer>> {
    None
}
```

The default implementations return `None`, preserving the current behavior and
keeping existing `V8InspectorClientImpl` implementations source-compatible.

The C++ `V8InspectorClient` wrapper would override the corresponding public V8
virtual methods and forward them to Rust.

A Rust callback scope is proposed because implementations may need to inspect
the value, its prototype or its internal fields in order to determine the
subtype. Scope construction and other unsafe operations should remain inside
the binding.

Returning `None` maps to a null `std::unique_ptr<StringBuffer>`, allowing V8 to
use its built-in value classification.

## Safety

- The callback is invoked synchronously by V8.
- The supplied `Local<Value>` cannot outlive the callback scope.
- Returned `StringBuffer` ownership is transferred to V8.
- No private V8 symbols are required.
- Existing client implementations continue to compile because both methods
  have defaults.

## Tests

1. A marked object receives the subtype and description returned by the client.
2. An ordinary object retains V8's default representation.
3. Returning `None` from both callbacks preserves current behavior.
4. Returning a subtype without a description follows V8's existing fallback
   behavior.
5. The callback can safely use the supplied scope to inspect the value.

The result can be observed through a normal `Runtime.evaluate` response, so
these tests do not depend on any additional session APIs.

## Open question

The C++ `valueSubtype` callback receives a value but no context. Should the Rust
binding provide a `PinScope` constructed from the active Inspector context, as
proposed above, or mirror the C++ signature more literally and pass only the
value?

Providing a scope is more useful to Rust implementations, but I would like to
confirm that this is the preferred API shape before submitting a PR.

satouriko:
Closes #2032.

## Summary

Expose V8 Inspector's `valueSubtype` and `descriptionForValueSubtype`
callbacks to Rust embedders.

- add source-compatible default methods to `V8InspectorClientImpl`
- establish the required V8 and Rust callback scopes before invoking client code
- pass scoped `Local<Value>` handles to the Rust callbacks
- return optional descriptions through `UniquePtr<StringBuffer>`, following the existing ownership pattern
- add a focused Inspector test covering custom subtype and description metadata while verifying that ordinary objects retain V8's default representation

## Testing

- `cargo fmt --check`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=/usr/lib/llvm-19/lib cargo clippy --all-targets --locked -- -D clippy::all`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=/usr/lib/llvm-19/lib cargo nextest run --all-targets --locked`

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -3562,6 +3562,13 @@ void v8_inspector__V8InspectorClient__BASE__consoleAPIMessage(
     const v8_inspector::StringView& message,
     const v8_inspector::StringView& url, unsigned lineNumber,
     unsigned columnNumber, v8_inspector::V8StackTrace* stackTrace);
+v8_inspector::StringBuffer* v8_inspector__V8InspectorClient__BASE__valueSubtype(
+    v8_inspector::V8InspectorClient* self, v8::Context* context,
+    v8::Value* value);
+v8_inspector::StringBuffer*
+v8_inspector__V8InspectorClient__BASE__descriptionForValueSubtype(
+    v8_inspector::V8InspectorClient* self, v8::Context* context,
+    v8::Value* value);
 v8::Context* v8_inspector__V8InspectorClient__BASE__ensureDefaultContextInGroup(
     v8_inspector::V8InspectorClient* self, int context_group_id);
 v8_inspector::StringBuffer*
@@ -3598,6 +3605,24 @@ struct v8_inspector__V8InspectorClient__BASE
     v8_inspector__V8InspectorClient__BASE__consoleAPIMessage(
         this, contextGroupId, level, message, url, lineNumber, columnNumber,
         stackTrace);
+  }
+  std::unique_ptr<v8_inspector::StringBuffer> valueSubtype(
+      v8::Local<v8::Value> value) override {
+    v8::Isolate* isolate = v8::Isolate::GetCurrent();
+    if (isolate == nullptr) return nullptr;
+    v8::Local<v8::Context> context = isolate->GetCurrentContext();
+    if (context.IsEmpty()) return nullptr;
+    v8_inspector::StringBuffer* b =
+        v8_inspector__V8InspectorClient__BASE__valueSubtype(
+            this, local_to_ptr(context), local_to_ptr(value));
+    return std::unique_ptr<v8_inspector::StringBuffer>(b);
+  }
+  std::unique_ptr<v8_inspector::StringBuffer> descriptionForValueSubtype(
+      v8::Local<v8::Context> context, v8::Local<v8::Value> value) override {
+    v8_inspector::StringBuffer* b =
+        v8_inspector__V8InspectorClient__BASE__descriptionForValueSubtype(
+            this, local_to_ptr(context), local_to_ptr(value));
+    return std::unique_ptr<v8_inspector::StringBuffer>(b);
   }
   v8::Local<v8::Context> ensureDefaultContextInGroup(
       int context_group_id) override {
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -13,9 +13,11 @@
 //! https://github.com/nodejs/node/tree/v13.7.0/src/inspector
 //! https://github.com/denoland/deno/blob/v0.38.0/cli/inspector.rs
 
+use crate::CallbackScope;
 use crate::Context;
 use crate::Isolate;
 use crate::Local;
+use crate::PinScope;
 use crate::StackTrace;
 use crate::Value;
 use crate::isolate::RealIsolate;
@@ -27,6 +29,7 @@ use crate::support::int;
 use std::cell::UnsafeCell;
 use std::ffi::c_void;
 use std::fmt::{self, Debug, Formatter};
+use std::pin::pin;
 
 unsafe extern "C" {
   fn v8_inspector__V8Inspector__Channel__BASE__CONSTRUCT(
@@ -269,6 +272,42 @@ unsafe extern "C" fn v8_inspector__V8InspectorClient__BASE__consoleAPIMessage(
         column_number,
         stack_trace,
       );
+  }
+}
+
+#[unsafe(no_mangle)]
+unsafe extern "C" fn v8_inspector__V8InspectorClient__BASE__valueSubtype(
+  this: *mut RawV8InspectorClient,
+  context: Local<Context>,
+  value: Local<Value>,
+) -> *mut StringBuffer {
+  let scope = pin!(unsafe { CallbackScope::new(context) });
+  let mut scope = scope.init();
+  unsafe {
+    V8InspectorClientHeap::from_raw(this)
+      .imp
+      .value_subtype(&mut scope, value)
+      .and_then(|mut v| v.take())
+      .map(|r| r.into_raw())
+      .unwrap_or(std::ptr::null_mut())
+  }
+}
+
+#[unsafe(no_mangle)]
+unsafe extern "C" fn v8_inspector__V8InspectorClient__BASE__descriptionForValueSubtype(
+  this: *mut RawV8InspectorClient,
+  context: Local<Context>,
+  value: Local<Value>,
+) -> *mut StringBuffer {
+  let scope = pin!(unsafe { CallbackScope::new(context) });
+  let mut scope = scope.init();
+  unsafe {
+    V8InspectorClientHeap::from_raw(this)
+      .imp
+      .description_for_value_subtype(&mut scope, value)
+      .and_then(|mut v| v.take())
+      .map(|r| r.into_raw())
+      .unwrap_or(std::ptr::null_mut())
   }
 }
 
@@ -531,6 +570,42 @@ pub trait V8InspectorClientImpl {
     column_number: u32,
     stack_trace: &mut V8StackTrace,
   ) {
+  }
+
+  /// Returns a custom Chrome DevTools Protocol `Runtime.RemoteObject` subtype
+  /// for `value`. Use one of the protocol's defined subtype enum values.
+  /// Returning `Some` causes V8 to call
+  /// [`Self::description_for_value_subtype`].
+  ///
+  /// The callback scope uses the isolate's current context as a best-effort
+  /// context; it is not necessarily the context in which `value` originated.
+  /// If the isolate has no current context, V8 skips this callback entirely.
+  /// This callback runs while the inspector is constructing a value mirror.
+  /// Operations such as property access can execute JavaScript through getters
+  /// or proxies, so wrap them in a [`crate::TryCatch`] to avoid leaving a
+  /// pending exception in the inspector's mirror-building path.
+  fn value_subtype<'s>(
+    &self,
+    scope: &mut PinScope<'s, '_>,
+    value: Local<'s, Value>,
+  ) -> Option<UniquePtr<StringBuffer>> {
+    None
+  }
+
+  /// Returns the description for a value whose custom subtype was returned by
+  /// [`Self::value_subtype`]. Returning `None` makes V8 fall back to the default
+  /// object mirror, which also discards the custom subtype unless it is
+  /// `"error"` or `"array"`.
+  ///
+  /// Like [`Self::value_subtype`], this callback runs while the inspector is
+  /// constructing a value mirror. Wrap operations that can execute JavaScript
+  /// in a [`crate::TryCatch`] so an exception does not remain pending.
+  fn description_for_value_subtype<'s>(
+    &self,
+    scope: &mut PinScope<'s, '_>,
+    value: Local<'s, Value>,
+  ) -> Option<UniquePtr<StringBuffer>> {
+    None
   }
 
   fn ensure_default_context_in_group(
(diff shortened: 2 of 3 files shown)


bartlomieju (commented):
Nice, well-scoped addition — the C++ shims follow the existing `resourceNameToUrl` ownership pattern, the `pin!(CallbackScope::new(..))` idiom matches the rest of the codebase, and the test asserts the negative case too (plain `{}` keeps V8's default representation), which is exactly the part that would regress silently.

One correctness issue worth fixing before merge (null context in `valueSubtype`), plus some docs/test suggestions inline.

src/binding.cc
```diff
@@ -3566,6 +3573,24 @@ struct v8_inspector__V8InspectorClient__BASE
...
         stackTrace);
   }
+  std::unique_ptr<v8_inspector::StringBuffer> valueSubtype(
+      v8::Local<v8::Value> value) override {
+    v8::Isolate* isolate = v8::Isolate::GetCurrent();
+    v8::HandleScope scope(isolate);
+    v8_inspector::StringBuffer* b =
+        v8_inspector__V8InspectorClient__BASE__valueSubtype(
+            this, local_to_ptr(isolate->GetCurrentContext()),
+            local_to_ptr(value));
```
bartlomieju: **Potential UB: `GetCurrentContext()` may be empty here.**

V8's `valueSubtype` takes no context, so this substitutes `isolate->GetCurrentContext()`. `local_to_ptr` is `return *local`, which yields `nullptr` for an empty `Local`. On the Rust side the shim receives it as `Local<Context>`, and `Local` is `NonNull<T>` (`src/handle.rs:110`) — a null there is immediately UB, and `CallbackScope::new` will then dereference it.

In practice V8 calls this from `value-mirror.cc:1101` / `:1848` inside an `InjectedScript::Scope`, which does enter a `v8::Context::Scope`, so today it happens to be non-empty. But that's a load-bearing assumption about V8 internals with no guard. Cheap fix:

```cpp
v8::Local<v8::Context> context = isolate->GetCurrentContext();
if (context.IsEmpty()) return nullptr;
```

Related and worth documenting on the Rust trait method: the *current* context isn't guaranteed to be the context V8 is mirroring for. The call site uses `clientFor(context)` with a context the callback never sees; `descriptionForValueSubtype` gets the real one, `valueSubtype` only gets a best-effort one. Embedders that key off the context's global should know that.

src/binding.cc
```diff
@@ -3566,6 +3573,24 @@ struct v8_inspector__V8InspectorClient__BASE
...
+    return std::unique_ptr<v8_inspector::StringBuffer>(b);
+  }
+  std::unique_ptr<v8_inspector::StringBuffer> descriptionForValueSubtype(
+      v8::Local<v8::Context> context, v8::Local<v8::Value> value) override {
+    v8::HandleScope scope(v8::Isolate::GetCurrent());
```
bartlomieju: Minor: `v8::Isolate::GetCurrent()` is unnecessary here since the context is already in hand (`context->GetIsolate()`), and this `HandleScope` is redundant with the one the Rust `CallbackScope` opens. Harmless, just noise.

src/inspector.rs
```diff
@@ -529,6 +568,22 @@ pub trait V8InspectorClientImpl {
...
+    None
+  }
+
+  fn description_for_value_subtype<'s>(
+    &self,
+    scope: &mut PinScope<'s, '_>,
+    value: Local<'s, Value>,
+  ) -> Option<UniquePtr<StringBuffer>> {
+    None
+  }
```
bartlomieju, line 609: These are the only new public API surface in the PR and they have no doc comments, unlike most of the surrounding API. Worth documenting:

- Returning `Some` from `value_subtype` is what causes V8 to call `description_for_value_subtype` at all — that coupling isn't discoverable from the signatures.
- Subtype strings should come from the CDP `RemoteObject.subtype` enum (`node`, `array`, `error`, …); arbitrary strings render inconsistently in DevTools.
- **Reentrancy hazard:** these fire in the middle of V8's value-mirror walk. An embedder that touches the object (getters, Proxy traps — see the test below) can run arbitrary JS and leave a pending exception that `value-mirror.cc` isn't prepared for. A note recommending a `TryCatch` would save someone a nasty debugging session.

Also worth noting for reviewers: `value_subtype` is now on the hot path for every mirrored object during `Runtime.evaluate`/console logging, costing an FFI hop + `HandleScope` + `CallbackScope` per object even for embedders using the `None` default. Probably fine (inspector paths aren't hot in production), but it's a nonzero cost that didn't exist before.

tests/test_api.rs
```diff
@@ -7183,6 +7185,81 @@ fn inspector_dispatch_protocol_message() {
...
+
+  impl V8InspectorClientImpl for Client {
+    fn value_subtype<'s>(
+      &self,
+      scope: &mut v8::PinScope<'s, '_>,
+      value: v8::Local<'s, v8::Value>,
+    ) -> Option<v8::UniquePtr<StringBuffer>> {
+      let object = value.to_object(scope)?;
+      let key = v8::String::new(scope, "__rusty_v8_value_subtype")?;
+      if object.get(scope, key.into())?.is_true() {
```
bartlomieju, line 7261: `object.get(..)` can invoke a getter or Proxy trap and throw, which would leave an exception pending while V8 is mid-mirror-walk. Since this test doubles as the usage example people will copy, it'd be better to model the safe pattern with a `TryCatch` here rather than the bare `?`.

tests/test_api.rs
```diff
@@ -7183,6 +7185,81 @@ fn inspector_dispatch_protocol_message() {
...
+  session.dispatch_protocol_message(StringView::from(
+    &br#"{"id":2,"method":"Runtime.evaluate","params":{"expression":"({})","contextId":1}}"#[..],
+  ));
+
+  let state = channel.state.borrow();
+  assert_eq!(state.responses.len(), 2);
+  assert!(state.responses[0].contains(r#""subtype":"node""#));
+  assert!(state.responses[0].contains(r#""description":"marked object""#));
+  assert!(state.responses[1].contains(r#""description":"Object""#));
+  assert!(!state.responses[1].contains(r#""subtype""#));
```
bartlomieju: Two small gaps:

1. Missing `inspector.context_destroyed(context);` at the end — the neighbouring inspector tests all do this.
2. No coverage for `description_for_value_subtype` returning `None` (or a null `UniquePtr`) while `value_subtype` returns `Some`. That's the fallback path where V8 supplies its own description, and it's the one most likely to break unnoticed.

src/binding.cc
satouriko: Fixed in 86cc33e. valueSubtype now returns nullptr when GetCurrentContext() is empty, and the Rust docs note that this is only a best-effort context.

src/binding.cc
satouriko: Removed the redundant HandleScopes. descriptionForValueSubtype no longer calls Isolate::GetCurrent().

src/inspector.rs
satouriko: Added docs for the callback coupling, CDP subtype values, best-effort context, and the reentrancy/TryCatch requirement. The per-object FFI/scope cost remains as noted.

tests/test_api.rs
satouriko: Updated the test client to create a TryCatch with tc_scope! around to_object/get in both callbacks.

tests/test_api.rs
satouriko: Added context_destroyed and coverage for description_for_value_subtype returning None. The test verifies fallback to the default Object description and the resulting absence of a custom subtype.

bartlomieju (approved):
Nice, well-scoped addition. The C++ shims follow the existing `resourceNameToUrl` ownership pattern, the `pin!(CallbackScope::new(..))` idiom matches the rest of the file, and the test covers the negative case too.

One doc fix before/after merge, please:

The doc comment on `description_for_value_subtype` says returning `None` "lets V8 use its default description". That's not quite what happens — in `clientMirror` (`v8/src/inspector/value-mirror.cc`), a null description falls through to the two-argument `ObjectMirror(value, descriptionForObject(...))`, which **drops the custom subtype as well** (except for the special-cased `"error"` and `"array"` subtypes). Your own test encodes the correct behavior — `responses[1]` asserts there is no `"subtype"` even though `value_subtype` returned `"node"` — so it's only the doc that's misleading. Something like:

> Returning `None` makes V8 fall back to the default object mirror, which discards the custom subtype as well, unless the subtype is `error` or `array`.

Two optional nits:

- In `value_subtype`'s doc, it's worth saying that when the isolate has no current context the callback is **skipped entirely** (the shim returns `nullptr`), not just that the context is best-effort — that's the thing an embedder would be confused by when a subtype silently doesn't show up.
- The `defaultDescription` assertions in the test are the most valuable and least obvious part of it. A short comment (`// V8 drops the custom subtype when the client supplies no description.`) would save the next reader a trip through `value-mirror.cc`.

Approving — CI is green on all legs and none of this blocks.

satouriko:
Updated in `41a96a4`:

- corrected the `None` documentation to describe the default object-mirror fallback and custom subtype removal
- documented that `value_subtype` is skipped when the isolate has no current context
- added a test comment explaining the `defaultDescription` assertions

src/binding.cc
```diff
@@ -3599,6 +3606,23 @@ struct v8_inspector__V8InspectorClient__BASE
         this, contextGroupId, level, message, url, lineNumber, columnNumber,
         stackTrace);
   }
+  std::unique_ptr<v8_inspector::StringBuffer> valueSubtype(
+      v8::Local<v8::Value> value) override {
+    v8::Isolate* isolate = v8::Isolate::GetCurrent();
+    v8::Local<v8::Context> context = isolate->GetCurrentContext();
```
bartlomieju, line 3613: `context.IsEmpty()` is guarded on the next line but `isolate` isn't. In practice the inspector always runs with an entered isolate so this shouldn't fire, but the function is already defensive about the context and it's a one-liner:

```suggestion
    v8::Isolate* isolate = v8::Isolate::GetCurrent();
    if (isolate == nullptr) return nullptr;
    v8::Local<v8::Context> context = isolate->GetCurrentContext();
```

bartlomieju (approved):
Nice work, and thanks for the thorough test.

The ownership handling matches the existing `resource_name_to_url` pattern exactly, and the scope construction follows the `pin!(CallbackScope::new(..)).init()` shape used elsewhere in `isolate.rs` and `function.rs` — good that this didn't invent a new convention for the first inspector callback to hand out a scope.

Particularly appreciated:

- Bridging the signature gap sensibly. V8 gives `valueSubtype` no context but `descriptionForValueSubtype` one, and the doc comment states the best-effort-context caveat outright rather than hiding it.
- Documenting that these run during mirror construction, so JS-executing operations need a `TryCatch` to avoid leaving a pending exception. That's easy to get wrong and the test client models it correctly.
- Testing through real `Runtime.evaluate` traffic rather than calling the callbacks directly, and covering the negative cases — including asserting that V8 drops a custom subtype when no description is supplied.

Default trait methods keep this source-compatible for existing implementors, and CI is green.

One small robustness suggestion inline; happy for it to land either way.

src/binding.cc
satouriko: Added the null check before dereferencing the current isolate. If either the isolate or current context is unavailable, `valueSubtype` now returns `nullptr` and V8 falls back to its default mirror. The targeted `inspector_value_subtype` test still passes.

bartlomieju (approved):
Thanks!

PR #2039, opened 2026-08-01
feat: expose inspector inspected objects

### LINKED ISSUE #2034: Expose V8InspectorSession::Inspectable and addInspectedObject
## Summary

V8 exposes
[`V8InspectorSession::Inspectable`](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/include/v8-inspector.h)
and `V8InspectorSession::addInspectedObject` as public embedder APIs.

They are used to add embedder-selected values to the Inspector's inspected
object history, which is exposed by the command-line API as `$0`, `$1`, and so
on.

`rusty_v8` does not currently expose either API.

## Impact

An embedder may know that a host object has been selected by an
embedder-provided Inspector domain, but it cannot make that value available as
`$0` in the DevTools Console.

Consequently:

- selecting a host object cannot update `$0`;
- previously selected objects cannot be accessed through `$1` through `$4`;
- the frontend's selected-object state cannot be connected to the Inspector
  command-line API.

This does not affect normal JavaScript evaluation or Console output. It affects
the integration between an embedder-specific selection operation and the
Inspector Console.

## Proposed API

Expose a Rust trait corresponding to V8's abstract `Inspectable` class:

```rust
pub trait InspectableImpl {
    fn get<'s>(
        &self,
        scope: &mut PinScope<'s, '_>,
        context: Local<'s, Context>,
    ) -> Local<'s, Value>;
}
```

Expose an owning wrapper:

```rust
pub struct Inspectable {
    // Owns the C++ wrapper until transferred to a session.
}

impl Inspectable {
    pub fn new(implementation: Box<dyn InspectableImpl>) -> Self;
}
```

Add a consuming session method:

```rust
impl V8InspectorSession {
    pub fn add_inspected_object(
        &self,
        inspectable: Inspectable,
    );
}
```

`InspectableImpl::get` is called lazily by V8 when the corresponding `$n`
binding is evaluated. This allows an implementation to retain the underlying
value in a `Global<Value>` and create a fresh `Local<Value>` for the callback
scope.

`add_inspected_object` consumes the Rust wrapper because ownership is
transferred to V8 as a `std::unique_ptr<Inspectable>`.

## Ownership model

The C++ wrapper and Rust implementation should use separate allocations.

Proposed lifecycle:

1. `Inspectable::new` allocates a C++ `Inspectable` subclass.
2. A separate Rust allocation owns the boxed `InspectableImpl`.
3. The C++ wrapper stores an opaque pointer to that Rust allocation.
4. Before transfer, the Rust `Inspectable` owns the C++ wrapper.
5. `add_inspected_object` consumes the Rust wrapper and transfers the C++
   object to V8.
6. The C++ destructor invokes a Rust callback that drops the Rust allocation
   exactly once.
7. If `Inspectable` is dropped before transfer, its normal destructor performs
   the same cleanup.

Keeping the two allocations separate avoids depending on:

- a shared Rust/C++ object layout;
- allocator compatibility;
- trailing Rust fields inside a C++ allocation;
- manual reconstruction of mixed-language heap objects.

## Callback safety

The C++ `Inspectable::get` implementation receives a V8 context. The binding
can construct a callback scope from that context and pass both the scope and
context to Rust.

The returned `Local<Value>` is used synchronously by V8 and cannot outlive the
callback scope.

The implementation must remain valid until V8 destroys the Inspectable, which
is enforced by transferring ownership of the boxed trait object together with
the C++ wrapper.

## Tests

1. The most recently added object is available as `$0`.
2. Previously added objects move to `$1`, `$2`, and so on.
3. Adding more objects than V8's inspected-object history limit drops the
   oldest object.
4. An object removed because of history overflow is dropped exactly once.
5. Remaining objects are dropped exactly once when the session is destroyed.
6. Dropping an `Inspectable` before passing it to a session does not leak.
7. The `get` callback receives a usable scope and the expected context.
8. Context and session teardown do not cause a use-after-free.

A drop counter in the test implementation can verify overflow and session
destruction behavior.

## Open questions

1. Is a boxed trait object consistent with the preferred pattern for other
   Rust implementations of C++ virtual interfaces in this crate?
2. Should `Inspectable::new` accept only `Box<dyn InspectableImpl>`, or also
   provide a generic convenience constructor?
3. Should `add_inspected_object` take `Inspectable` by value, making ownership
   transfer explicit, as proposed?

satouriko:
Closes #2034.

## Summary

Expose V8 Inspector's `V8InspectorSession::Inspectable` and
`addInspectedObject` APIs to Rust embedders, allowing embedder-selected values
to appear in the Inspector Console as `$0` through `$4`.

- add `InspectableImpl` and an owning `Inspectable` wrapper
- allocate the C++ subclass separately from the boxed Rust implementation
- establish the required V8 handle/context scope and Rust callback scope around `get()`
- consume the Rust wrapper when transferring the C++ object into V8's `unique_ptr`
- drop the Rust implementation when V8 destroys or evicts the inspected object
- add focused tests for `$0` round-trip behavior and Rust-side cleanup on eviction

## Testing

- `cargo fmt --check`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=/usr/lib/llvm-19/lib cargo clippy --all-targets --locked -- -D clippy::all`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=/usr/lib/llvm-19/lib cargo nextest run --all-targets --locked` (288 passed)

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -3504,6 +3504,62 @@ void v8_inspector__V8InspectorSession__cancelPauseOnNextStatement(
 }
 }  // extern "C"
 
+extern "C" {
+const v8::Value* v8_inspector__V8InspectorSession__Inspectable__BASE__get(
+    void* rust_impl, const v8::Context* context);
+void v8_inspector__V8InspectorSession__Inspectable__BASE__DROP(void* rust_impl);
+}
+
+class v8_inspector__V8InspectorSession__Inspectable__BASE final
+    : public v8_inspector::V8InspectorSession::Inspectable {
+ public:
+  explicit v8_inspector__V8InspectorSession__Inspectable__BASE(void* rust_impl)
+      : rust_impl_(rust_impl) {}
+
+  ~v8_inspector__V8InspectorSession__Inspectable__BASE() override {
+    v8_inspector__V8InspectorSession__Inspectable__BASE__DROP(rust_impl_);
+  }
+
+  v8::Local<v8::Value> get(v8::Local<v8::Context> context) override {
+    // The Rust CallbackScope relies on `NewCallbackScope for Local<Context>`
+    // having `NEEDS_SCOPE == false` and must not open its own HandleScope.
+    // Handles created by the Rust implementation are allocated in this scope
+    // and must be escaped here. Opening a nested HandleScope in Rust would
+    // destroy that scope before Escape uses its handle, causing a
+    // use-after-free.
+    v8::Isolate* isolate = v8::Isolate::GetCurrent();
+    if (isolate == nullptr) return v8::Local<v8::Value>();
+    v8::EscapableHandleScope handle_scope(isolate);
+    v8::Context::Scope context_scope(context);
+    return handle_scope.Escape(
+        ptr_to_local(v8_inspector__V8InspectorSession__Inspectable__BASE__get(
+            rust_impl_, local_to_ptr(context))));
+  }
+
+ private:
+  void* rust_impl_;
+};
+
+extern "C" {
+v8_inspector::V8InspectorSession::Inspectable*
+v8_inspector__V8InspectorSession__Inspectable__NEW(void* rust_impl) {
+  return new v8_inspector__V8InspectorSession__Inspectable__BASE(rust_impl);
+}
+
+void v8_inspector__V8InspectorSession__Inspectable__DELETE(
+    v8_inspector::V8InspectorSession::Inspectable* inspectable) {
+  delete inspectable;
+}
+
+void v8_inspector__V8InspectorSession__addInspectedObject(
+    v8_inspector::V8InspectorSession* self,
+    v8_inspector::V8InspectorSession::Inspectable* inspectable) {
+  self->addInspectedObject(
+      std::unique_ptr<v8_inspector::V8InspectorSession::Inspectable>(
+          inspectable));
+}
+}
+
 struct v8_inspector__V8Inspector__Channel__BASE
     : public v8_inspector::V8Inspector::Channel {
   using v8_inspector::V8Inspector::Channel::Channel;
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -98,6 +98,16 @@ unsafe extern "C" {
   fn v8_inspector__V8InspectorSession__canDispatchMethod(
     method: StringView,
   ) -> bool;
+  fn v8_inspector__V8InspectorSession__Inspectable__NEW(
+    rust_impl: *mut c_void,
+  ) -> *mut RawInspectable;
+  fn v8_inspector__V8InspectorSession__Inspectable__DELETE(
+    inspectable: *mut RawInspectable,
+  );
+  fn v8_inspector__V8InspectorSession__addInspectedObject(
+    session: *mut RawV8InspectorSession,
+    inspectable: *mut RawInspectable,
+  );
 
   fn v8_inspector__StringBuffer__DELETE(this: *mut StringBuffer);
   fn v8_inspector__StringBuffer__string(this: &StringBuffer) -> StringView<'_>;
@@ -676,6 +686,89 @@ impl Debug for V8InspectorClient {
 
 #[repr(C)]
 #[derive(Debug)]
+struct RawInspectable(Opaque);
+
+impl Drop for RawInspectable {
+  fn drop(&mut self) {
+    unsafe {
+      v8_inspector__V8InspectorSession__Inspectable__DELETE(self);
+    }
+  }
+}
+
+// A trait object is a fat pointer, so box it once more before passing it through
+// the FFI as a thin `*mut c_void`.
+struct InspectableData {
+  imp: Box<dyn InspectableImpl>,
+}
+
+/// Supplies the value of an object added to the inspector's `$0` through `$4`
+/// history.
+pub trait InspectableImpl {
+  /// Called by the inspector from within a V8 callback when the console
+  /// dereferences one of `$0` through `$4`.
+  ///
+  /// There is no way to return an empty handle; if no value is available,
+  /// return `undefined`.
+  fn get<'s>(
+    &self,
+    scope: &mut PinScope<'s, '_>,
+    context: Local<'s, Context>,
+  ) -> Local<'s, Value>;
+}
+
+/// An object that can be added to an inspector session's `$0` through `$4`
+/// history.
+pub struct Inspectable {
+  raw: UniqueRef<RawInspectable>,
+}
+
+impl Inspectable {
+  pub fn new(imp: Box<dyn InspectableImpl>) -> Self {
+    let data = Box::into_raw(Box::new(InspectableData { imp })).cast();
+    let raw = unsafe {
+      UniqueRef::from_raw(v8_inspector__V8InspectorSession__Inspectable__NEW(
+        data,
+      ))
+    };
+    Self { raw }
+  }
+}
+
+impl Debug for Inspectable {
+  fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+    f.debug_struct("Inspectable").finish()
+  }
+}
+
+#[unsafe(no_mangle)]
+unsafe extern "C" fn v8_inspector__V8InspectorSession__Inspectable__BASE__get(
+  rust_impl: *mut c_void,
+  context: Local<Context>,
+) -> *const Value {
+  let data = unsafe { &*rust_impl.cast::<InspectableData>() };
+  // SAFETY: `CallbackScope::new(context)` must not open its own HandleScope.
+  // `NewCallbackScope for Local<Context>` has `NEEDS_SCOPE == false`, and
+  // `make_new_callback_scope` constructs it with `needs_scope == false`. The
+  // handle returned here is allocated in the EscapableHandleScope opened by
+  // the C++ shim and escaped there. If Rust opens its own HandleScope, that
+  // scope is destroyed before C++ escapes the handle, causing a use-after-free.
+  let scope = pin!(unsafe { CallbackScope::new(context) });
+  let mut scope = scope.init();
+  data.imp.get(&mut scope, context).as_non_null().as_ptr()
+}
+
+#[unsafe(no_mangle)]
+unsafe extern "C" fn v8_inspector__V8InspectorSession__Inspectable__BASE__DROP(
+  rust_impl: *mut c_void,
+) {
+  unsafe {
+    drop(Box::from_raw(rust_impl.cast::<InspectableData>()));
+  }
+}
+
+#[repr(C)]
+#[derive(Debug)]
 pub struct RawV8InspectorSession(Opaque);
 
 pub struct V8InspectorSession {
@@ -728,6 +821,15 @@ impl V8InspectorSession {
     unsafe {
       v8_inspector__V8InspectorSession__cancelPauseOnNextStatement(
         self.raw.as_ptr(),
+      );
+    }
+  }
+
+  pub fn add_inspected_object(&self, inspectable: Inspectable) {
+    unsafe {
+      v8_inspector__V8InspectorSession__addInspectedObject(
+        self.raw.as_ptr(),
+        inspectable.raw.into_raw(),
       );
     }
   }
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -7097,6 +7097,28 @@ impl v8::inspector::ChannelImpl for ChannelCounter {
   }
 }
 
+struct TestInspectable {
+  value: v8::Global<v8::Value>,
+  drop_count: Arc<AtomicUsize>,
+}
+
+impl v8::inspector::InspectableImpl for TestInspectable {
+  fn get<'s>(
+    &self,
+    scope: &mut v8::PinScope<'s, '_>,
+    context: v8::Local<'s, v8::Context>,
+  ) -> v8::Local<'s, v8::Value> {
+    assert_eq!(scope.get_current_context(), context);
+    v8::Local::new(scope, &self.value)
+  }
+}
+
+impl Drop for TestInspectable {
+  fn drop(&mut self) {
+    self.drop_count.fetch_add(1, Ordering::SeqCst);
+  }
+}
+
 #[test]
 fn inspector_can_dispatch_method() {
   use v8::inspector::*;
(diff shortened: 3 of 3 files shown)


src/binding.cc
```diff
@@ -3471,6 +3471,54 @@ void v8_inspector__V8InspectorSession__cancelPauseOnNextStatement(
...
+    : public v8_inspector::V8InspectorSession::Inspectable {
+ public:
+  explicit v8_inspector__V8InspectorSession__Inspectable__BASE(void* rust_impl)
+      : rust_impl_(rust_impl) {}
+
+  ~v8_inspector__V8InspectorSession__Inspectable__BASE() override {
+    v8_inspector__V8InspectorSession__Inspectable__BASE__DROP(rust_impl_);
+  }
+
+  v8::Local<v8::Value> get(v8::Local<v8::Context> context) override {
```
bartlomieju, line 3523: This `EscapableHandleScope` is load-bearing in a non-obvious way, and it's the only place in `binding.cc` that needs one. It works because `CallbackScope::new(Local<Context>)` on the Rust side has `needs_scope: false` and therefore does *not* open its own `HandleScope` — so the handle the Rust impl creates is allocated here and survives to be escaped.

Worth a short comment so a future change to `CallbackScope` doesn't silently turn this into a use-after-free:

```c++
// The Rust `CallbackScope` built from a `Local<Context>` does not open its
// own HandleScope, so handles created by the Rust impl are allocated in this
// scope and must be escaped here.
```

src/binding.cc
```diff
@@ -3471,6 +3471,54 @@ void v8_inspector__V8InspectorSession__cancelPauseOnNextStatement(
...
+    v8::EscapableHandleScope handle_scope(v8::Isolate::GetCurrent());
```
bartlomieju: `v8::Isolate::GetCurrent()` is the only use in `binding.cc` outside the explicit `v8__Isolate__GetCurrent` export (line 177). Prefer deriving it from the argument you already have — no dependency on an entered-isolate invariant:

```suggestion
    v8::EscapableHandleScope handle_scope(context->GetIsolate());
```

src/inspector.rs
```diff
@@ -595,6 +608,71 @@ impl Debug for V8InspectorClient {
   }
 }
 
+#[repr(C)]
+struct RawInspectable(Opaque);
```
bartlomieju, line 689: Other `Raw*` types in this file (`RawV8InspectorSession`, `RawChannel`) derive `Debug`:

```suggestion
#[repr(C)]
#[derive(Debug)]
struct RawInspectable(Opaque);
```

src/inspector.rs
```diff
@@ -595,6 +608,71 @@ impl Debug for V8InspectorClient {
...
+  fn drop(&mut self) {
+    unsafe {
+      v8_inspector__V8InspectorSession__Inspectable__DELETE(self);
+    }
+  }
+}
+
+struct InspectableData {
+  imp: Box<dyn InspectableImpl>,
+}
```
bartlomieju, line 703: This single-field wrapper looks redundant at a glance, but it's necessary — a `Box<dyn InspectableImpl>` is a fat pointer and can't be thinned to `*mut c_void` directly. One line of comment would save the next reader the detour.

src/inspector.rs
```diff
@@ -595,6 +608,71 @@ impl Debug for V8InspectorClient {
...
+  fn get<'s>(
+    &self,
+    scope: &mut PinScope<'s, '_>,
+    context: Local<'s, Context>,
+  ) -> Local<'s, Value>;
+}
+
+pub struct Inspectable {
+  raw: UniqueRef<RawInspectable>,
+}
```
bartlomieju, line 724: Both new public items are undocumented. Worth at least a sentence each on `InspectableImpl` (that `get` is called by the inspector when the console dereferences `$0`–`$4`, from within a V8 callback) and on `Inspectable`.

Design nit, take it or leave it: `Inspectable` only ever exists to be immediately consumed by `add_inspected_object`. `add_inspected_object(&self, imp: Box<dyn InspectableImpl>)` would be a smaller surface with one fewer moving part. Keeping `Inspectable` for symmetry with the C++ API is defensible too.

tests/test_api.rs
```diff
@@ -7183,6 +7207,88 @@ fn inspector_dispatch_protocol_message() {
...
+  let session = inspector.connect(
+    1,
+    Channel::new(Box::new(channel.clone())),
+    StringView::from(&b"{}"[..]),
+    V8InspectorClientTrustLevel::FullyTrusted,
+  );
+  session.add_inspected_object(Inspectable::new(Box::new(TestInspectable {
+    value,
+    drop_count: Arc::new(AtomicUsize::new(0)),
+  })));
```
bartlomieju, line 7397: The `drop_count` here is created and then never asserted on, which reads as an oversight. Either drop the field for this test, or better: hold onto the `Arc`, drop the session/inspector at the end, and assert the impl was released — that covers a destruction path the other test doesn't.

tests/test_api.rs
```diff
@@ -7183,6 +7207,88 @@ fn inspector_dispatch_protocol_message() {
...
+
+  session.dispatch_protocol_message(StringView::from(
+    &br#"{"id":1,"method":"Runtime.evaluate","params":{"expression":"$0.answer","contextId":1,"includeCommandLineAPI":true}}"#[..],
+  ));
+
+  let state = channel.state.borrow();
+  assert_eq!(state.responses.len(), 1);
+  assert!(state.responses[0].contains(r#""value":42"#));
+}
```
bartlomieju, line 7435: Two destruction paths aren't covered by either test, and both are distinct code:

1. An `Inspectable` that is constructed but never added — that exercises `RawInspectable::drop` → `Inspectable__DELETE`, which nothing currently hits.
2. Session/inspector teardown while inspected objects are still live.

Both are a few lines each with the `TestInspectable` you already have.

tests/test_api.rs
```diff
@@ -7183,6 +7207,88 @@ fn inspector_dispatch_protocol_message() {
...
+      value: v8::Global::new(scope, value),
+      drop_count: drop_count.clone(),
+    })));
+    drop_counts.push(drop_count);
+  }
+
+  assert_eq!(drop_counts[0].load(Ordering::SeqCst), 1);
+  for drop_count in &drop_counts[1..] {
+    assert_eq!(drop_count.load(Ordering::SeqCst), 0);
+  }
```
bartlomieju, line 7477: This encodes the assumption that V8 retains exactly 5 inspected objects (`$0`–`$4`) and evicts the oldest. That's a stable DevTools contract so I'm fine with it, but `6` and the `[1..]` slice are opaque — a comment naming the `$0`–`$4` limit would help whoever debugs this after a V8 roll.

bartlomieju (commented):
Nice work — the tricky part here is right. I traced the handle lifetime: `CallbackScope::new(Local<Context>)` sets `needs_scope: false` (`src/scope.rs:1131-1142` → `make_new_callback_scope`), so the Rust scope does *not* open a nested `HandleScope`, which means the `Local` returned by the impl lands in the C++ `EscapableHandleScope` and `Escape()` promotes it correctly. Had `CallbackScope` opened its own scope, that escaped handle would be dangling. Ownership is also sound: `UniqueRef` → `into_raw` on transfer, and every destruction path (unused wrapper, eviction, session teardown) funnels through `~BASE` → `__DROP`.

Allocating separately with `new` instead of the in-place `__CONSTRUCT` pattern used by `Channel`/`V8InspectorClient` is the right call, since V8 takes a `unique_ptr` and will `delete` it.

A handful of nits and a test-coverage gap inline.

src/binding.cc
satouriko: Added the scope-lifetime comment. It now explains that the Rust `CallbackScope` does not open its own `HandleScope`, so Rust-created handles must be escaped from the C++ scope.

src/binding.cc
satouriko: I tried this suggestion against the V8 revision on current main, but it does not compile: `v8::Context::GetIsolate()` was removed in V8 commit `98282b9b8fa` (`[api] Remove deprecated GetIsolate`). That removal explicitly directs embedders to `v8::Isolate::GetCurrent()`, which is guaranteed to return the same isolate here, so I kept `GetCurrent()`.

src/inspector.rs
satouriko: Added `Debug` to `RawInspectable`, matching the other raw inspector types.

src/inspector.rs
satouriko: Added a comment explaining that the extra allocation turns the trait-object fat pointer into a thin pointer that can cross the `*mut c_void` FFI boundary.

src/inspector.rs
satouriko: Documented `InspectableImpl`, its `get` callback, and `Inspectable`. I kept the owning `Inspectable` wrapper for symmetry with the C++ API.

tests/test_api.rs
satouriko: The round-trip test now retains the `Arc`, verifies the implementation is still live, explicitly drops the session and inspector, and asserts that it was released.

tests/test_api.rs
satouriko: Added coverage for an `Inspectable` dropped before being added. The round-trip test now also covers session/inspector teardown while an inspected object is live.

tests/test_api.rs
satouriko: Added a comment naming the five-object `$0`-through-`$4` history and explaining that the sixth object evicts the oldest.

bartlomieju (commented):
The ownership handling here is sound — I checked the load-bearing claim and it holds: a `CallbackScope` built from a `Local<Context>` goes through `make_new_callback_scope`, which hardcodes `needs_scope: false`, so no `HandleScope` is opened on the Rust side and the handle really does need escaping in `binding.cc`.

My one request before merge is about that invariant rather than the code itself. The comment explaining it lives in `binding.cc`, but the thing it depends on is `NEEDS_SCOPE` for `impl NewCallbackScope<'s> for Local<'s, Context>` in `src/scope.rs`. If anyone ever flips that to `true`, `Inspectable::get` would escape a handle out of an already-destroyed scope — a use-after-free with no compile error and, most likely, no test failure. Whoever makes that change will be editing `src/scope.rs` and has no reason to go looking in `binding.cc`.

Could you mirror the comment onto the Rust shim (`v8_inspector__V8InspectorSession__Inspectable__BASE__get` in `src/inspector.rs`) and name `NEEDS_SCOPE` explicitly in both places? Something like:

```rust
// SAFETY: `CallbackScope::new(context)` must not open its own HandleScope
// (`NewCallbackScope for Local<Context>` has `NEEDS_SCOPE == false`). The
// handle returned here is allocated in the EscapableHandleScope opened by
// the C++ shim and escaped there. If `NEEDS_SCOPE` ever becomes true for
// `Local<Context>`, this becomes a use-after-free.
```

That way the hazard is visible from the file where the breaking change would be made.

satouriko:
Updated in `9d59259`. The Rust shim now documents the `NEEDS_SCOPE == false` invariant alongside the existing C++ explanation, and both comments spell out the failure mode: a nested Rust `HandleScope` would be destroyed before the C++ shim escapes its handle, causing a use-after-free.

src/binding.cc
```diff
@@ -3504,6 +3504,60 @@ void v8_inspector__V8InspectorSession__cancelPauseOnNextStatement(
...
+  }
+
+  v8::Local<v8::Value> get(v8::Local<v8::Context> context) override {
+    // The Rust CallbackScope relies on `NewCallbackScope for Local<Context>`
+    // having `NEEDS_SCOPE == false` and must not open its own HandleScope.
+    // Handles created by the Rust implementation are allocated in this scope
+    // and must be escaped here. Opening a nested HandleScope in Rust would
+    // destroy that scope before Escape uses its handle, causing a
+    // use-after-free.
+    v8::EscapableHandleScope handle_scope(v8::Isolate::GetCurrent());
```
bartlomieju: `Isolate::GetCurrent()` isn't null-checked. Unreachable in practice since `get()` only runs from inside an inspector callback with an entered isolate, so lower priority than the same nit on #2038 — noting it for symmetry rather than as a real risk.

src/inspector.rs
```diff
@@ -599,6 +612,86 @@ impl Debug for V8InspectorClient {
...
+// the FFI as a thin `*mut c_void`.
+struct InspectableData {
+  imp: Box<dyn InspectableImpl>,
+}
+
+/// Supplies the value of an object added to the inspector's `$0` through `$4`
+/// history.
+pub trait InspectableImpl {
+  /// Called by the inspector from within a V8 callback when the console
+  /// dereferences one of `$0` through `$4`.
```
bartlomieju, line 709: V8's `Inspectable::get` may return an empty `Local<Value>`, but this trait returns a non-optional `Local<'s, Value>`. That's the right choice — Rust can't represent an empty `Local` safely — but it leaves an embedder with no documented way to say "I have nothing", and someone will eventually reach for `unsafe` to fake an empty handle. Worth stating the intended answer:

```suggestion
  /// Called by the inspector from within a V8 callback when the console
  /// dereferences one of `$0` through `$4`.
  ///
  /// There is no way to return an empty handle; if no value is available,
  /// return `undefined`.
```

tests/test_api.rs
```diff
@@ -7240,6 +7262,119 @@ fn inspector_release_object_group() {
...
+  {
+    let state = channel.state.borrow();
+    assert_eq!(state.responses.len(), 1);
+    assert!(state.responses[0].contains(r#""value":42"#));
+  }
+
+  assert_eq!(drop_count.load(Ordering::SeqCst), 0);
+  drop(session);
+  drop(inspector);
+  assert_eq!(drop_count.load(Ordering::SeqCst), 1);
```
bartlomieju, line 7413: Both drops happen before the assertion, so this doesn't pin down which one releases the inspected object. Asserting in between makes the ownership claim precise:

```suggestion
  drop(session);
  assert_eq!(drop_count.load(Ordering::SeqCst), 1);
  drop(inspector);
  assert_eq!(drop_count.load(Ordering::SeqCst), 1);
```

(If it turns out the drop actually happens on `inspector` rather than `session`, that's worth knowing too — either way the test would then document the real ownership.)

bartlomieju (approved):
Nice work — this is trickier than #2038 and it's handled carefully.

I checked the invariant the handle-scope reasoning rests on: `impl NewCallbackScope for Local<'s, Context>` doesn't override `NEEDS_SCOPE`, so it takes the trait default of `false`. Your comment is accurate, and documenting it on *both* sides of the FFI is the right call — that's a non-local coupling that would break silently if someone later flipped `NEEDS_SCOPE` for `Local<Context>`.

Ownership looks correct in all three paths: `UniqueRef::drop` runs `drop_in_place`, which reaches `RawInspectable::drop` -> `__DELETE` -> C++ `delete` -> destructor -> `__DROP`, matching the existing `StringBuffer` shape. And `add_inspected_object` consuming by value plus `into_raw()` hands ownership to V8's `unique_ptr` without a double free. Worth noting for future readers: `Inspectable` deliberately has no `Drop` impl, which is what allows the partial move of `inspectable.raw` — adding one later would break that line (the compiler will say so).

The eviction test is the standout. Adding six objects and asserting only the oldest is dropped covers a case most bindings PRs would skip.

Three optional suggestions inline. None block merge.

tests/test_api.rs
satouriko: Moved the ownership assertion between the two drops. The count reaches 1 immediately after `drop(session)` and remains 1 after `drop(inspector)`, documenting that the session owns inspected-object destruction.

src/inspector.rs
satouriko: Documented the intended Rust-side contract: `InspectableImpl::get` cannot return an empty handle, so implementations should return `undefined` when no value is available.

src/binding.cc
satouriko: Added an explicit current-isolate check before constructing the `EscapableHandleScope`. If no isolate is entered, `get()` now returns an empty `Local<Value>`; the three inspected-object tests still pass.

bartlomieju (approved):
Thanks!

bartlomieju:
Ooops, can you please resolve conflicts @satouriko?

satouriko:
Resolved in `978994f`. I rebased onto current `main` (`94dab52`, including #2038) and kept the merged `inspector_value_subtype` test alongside the three inspected-object tests.

Local formatting, C++ formatting, Clippy, the three `inspector_inspect` tests, and `inspector_value_subtype` all pass. The new CI run is [31109697644](https://github.com/denoland/rusty_v8/actions/runs/31109697644).

PR #2040, opened 2026-08-01
feat: expose inspector object group release

satouriko:
Part of #2033.

## Summary

Expose V8 Inspector's `V8InspectorSession::releaseObjectGroup` API to Rust embedders.

- add `V8InspectorSession::release_object_group` following the existing `StringView` session API pattern
- forward directly to V8 without generated protocol headers or extra ownership machinery
- add a focused Inspector test proving released object-group IDs become unavailable

This intentionally contains only `releaseObjectGroup`; generated-header plumbing, `wrapObject`, and `unwrapObject` remain separate follow-up work as requested in #2033.

## Testing

- `cargo fmt --check`
- `clang-format-19 --verbose --Werror --dry-run src/*.cc src/*.hpp src/*.h`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=/usr/lib/llvm-19/lib cargo clippy --all-targets --locked -- -D clippy::all`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=/usr/lib/llvm-19/lib cargo nextest run --all-targets --locked` (287 passed)

Issue #2033, opened 2026-07-26
Expose V8InspectorSession remote-object lifecycle APIs


satouriko:
## Summary

V8 exposes the following public methods on
[`V8InspectorSession`](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/include/v8-inspector.h):

- `wrapObject`
- `unwrapObject`
- `releaseObjectGroup`

These methods are not currently exposed by `rusty_v8`.

Without them, a Rust embedder cannot use the Inspector's own remote-object
identity system from an embedder-provided Inspector command:

- A `Local<Value>` cannot be converted into a `Runtime.RemoteObject` with an
  Inspector-generated `objectId`.
- An `objectId` received from the frontend cannot be resolved back to its V8
  value and context.
- An embedder cannot directly release an object group that it created.

## Impact

The built-in Runtime and Debugger domains continue to work because V8 uses
these mechanisms internally.

The missing bindings affect embedder-provided Inspector functionality that
needs to exchange object references with the frontend. Without them, an
embedder would have to invent a separate object-ID registry instead of using
the Inspector's injected-script and object-group state.

That produces object IDs which are not naturally interoperable with standard
CDP operations such as `Runtime.getProperties`, and duplicates lifecycle
management already implemented by V8.

## Proposed API

Expose an opaque owned representation of the Inspector RemoteObject:

```rust
pub struct RemoteObject {
    // Owns a protocol::Runtime::API::RemoteObject.
}

impl RemoteObject {
    /// Serialize using the Inspector's CRDTP/CBOR representation.
    pub fn to_bytes(&self) -> Vec<u8>;
}
```

Add the corresponding session operations:

```rust
impl V8InspectorSession {
    pub fn wrap_object<'s>(
        &self,
        scope: &mut PinScope<'s, '_>,
        context: Local<'s, Context>,
        value: Local<'s, Value>,
        object_group: StringView,
        generate_preview: bool,
    ) -> Option<RemoteObject>;

    pub fn unwrap_object<'s>(
        &self,
        scope: &mut PinScope<'s, '_>,
        object_id: StringView,
    ) -> Result<UnwrappedObject<'s>, String>;

    pub fn release_object_group(
        &self,
        object_group: StringView,
    );
}
```

The unwrap result would contain all values returned by V8:

```rust
pub struct UnwrappedObject<'s> {
    pub value: Local<'s, Value>,
    pub context: Local<'s, Context>,
    pub object_group: Option<String>,
}
[...]

PR #2041, opened 2026-08-01
feat: expose inspector object wrapping

satouriko:
Part of #2033.

## Summary

Expose V8 Inspector's `V8InspectorSession::wrapObject` API without exposing generated protocol classes through the public Rust API.

- add `V8InspectorSession::wrap_object`, returning an owned `RemoteObject` and modeling null results as `Option`
- serialize remote objects to CRDTP/CBOR through the shared byte-vector bridge
- keep the generated protocol type opaque at the Rust FFI boundary and add only the GN include path required by the C++ binding target
- document the `kIdOnly`/`kPreview` behavior, context lookup semantics, and the `HandleScope` requirement
- test the unregistered-context `None` path, preview generation, and interoperability of wrapped object IDs with `Runtime.getProperties`

This intentionally contains only `wrapObject`; `releaseObjectGroup` is handled separately in #2040, and `unwrapObject` remains follow-up work as requested in #2033.

## Testing

- `cargo fmt --check`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=third_party/rust-toolchain/lib cargo nextest run --all-targets --locked --no-fail-fast` (290 passed)

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -8,6 +8,7 @@
 #include <memory>
 #include <mutex>
 #include <thread>
+#include <vector>
 
 #include "cppgc/allocation.h"
 #include "cppgc/persistent.h"
@@ -18,6 +19,7 @@
 #include "v8-callbacks.h"
 #include "v8-cppgc.h"
 #include "v8-fast-api-calls.h"
+#include "v8-inspector-protocol.h"
 #include "v8-inspector.h"
 #include "v8-internal.h"
 #include "v8-platform.h"
@@ -3514,6 +3516,22 @@ void v8_inspector__V8InspectorSession__releaseObjectGroup(
     v8_inspector::V8InspectorSession* self,
     v8_inspector::StringView object_group) {
   self->releaseObjectGroup(object_group);
+}
+
+v8_inspector::protocol::Runtime::API::RemoteObject*
+v8_inspector__V8InspectorSession__wrapObject(
+    v8_inspector::V8InspectorSession* self, const v8::Context* context,
+    const v8::Value* value, v8_inspector::StringView object_group,
+    bool generate_preview) {
+  return self
+      ->wrapObject(ptr_to_local(context), ptr_to_local(value), object_group,
+                   generate_preview)
+      .release();
+}
+
+void v8_inspector__RemoteObject__DELETE(
+    v8_inspector::protocol::Runtime::API::RemoteObject* self) {
+  delete self;
 }
 
 void v8_inspector__V8InspectorSession__schedulePauseOnNextStatement(
--- a/src/crdtp.rs
+++ b/src/crdtp.rs
@@ -208,7 +208,19 @@ struct DispatchResponseWrapper(Opaque);
 pub struct UberDispatcher(Opaque);
 
 #[repr(C)]
-struct CppVecU8(Opaque);
+pub(crate) struct CppVecU8(Opaque);
+
+impl CppVecU8 {
+  pub(crate) unsafe fn take_from_raw(ptr: *mut Self) -> Vec<u8> {
+    unsafe {
+      let len = crdtp__vec_u8__size(ptr);
+      let mut result = vec![0u8; len];
+      crdtp__vec_u8__copy(ptr, result.as_mut_ptr());
+      crdtp__vec_u8__DELETE(ptr);
+      result
+    }
+  }
+}
 
 #[repr(C)]
 struct RawSerializable(Opaque);
@@ -225,11 +237,7 @@ impl Serializable {
     unsafe {
       let vec = crdtp__vec_u8__new();
       crdtp__Serializable__AppendSerialized(self.ptr, vec);
-      let len = crdtp__vec_u8__size(vec);
-      let mut result = vec![0u8; len];
-      crdtp__vec_u8__copy(vec, result.as_mut_ptr());
-      crdtp__vec_u8__DELETE(vec);
-      result
+      CppVecU8::take_from_raw(vec)
     }
   }
 
@@ -625,11 +633,7 @@ pub fn json_to_cbor(json: &[u8]) -> Option<Vec<u8>> {
     let vec = crdtp__vec_u8__new();
     let ok = crdtp__json__ConvertJSONToCBOR(json.as_ptr(), json.len(), vec);
     if ok {
-      let len = crdtp__vec_u8__size(vec);
-      let mut result = vec![0u8; len];
-      crdtp__vec_u8__copy(vec, result.as_mut_ptr());
-      crdtp__vec_u8__DELETE(vec);
-      Some(result)
+      Some(CppVecU8::take_from_raw(vec))
     } else {
       crdtp__vec_u8__DELETE(vec);
       None
@@ -643,11 +647,7 @@ pub fn cbor_to_json(cbor: &[u8]) -> Option<Vec<u8>> {
     let vec = crdtp__vec_u8__new();
     let ok = crdtp__json__ConvertCBORToJSON(cbor.as_ptr(), cbor.len(), vec);
     if ok {
-      let len = crdtp__vec_u8__size(vec);
-      let mut result = vec![0u8; len];
-      crdtp__vec_u8__copy(vec, result.as_mut_ptr());
-      crdtp__vec_u8__DELETE(vec);
-      Some(result)
+      Some(CppVecU8::take_from_raw(vec))
     } else {
       crdtp__vec_u8__DELETE(vec);
       None
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -20,6 +20,7 @@ use crate::Local;
 use crate::PinScope;
 use crate::StackTrace;
 use crate::Value;
+use crate::crdtp::CppVecU8;
 use crate::isolate::RealIsolate;
 use crate::support::CxxVTable;
 use crate::support::Opaque;
@@ -87,6 +88,17 @@ unsafe extern "C" {
     session: *mut RawV8InspectorSession,
     object_group: StringView,
   );
+  fn v8_inspector__V8InspectorSession__wrapObject(
+    session: *mut RawV8InspectorSession,
+    context: *const Context,
+    value: *const Value,
+    object_group: StringView,
+    generate_preview: bool,
+  ) -> *mut RawRemoteObject;
+  fn v8_inspector__RemoteObject__DELETE(this: *mut RawRemoteObject);
+  fn v8_inspector__RemoteObject__toBytes(
+    this: *const RawRemoteObject,
+  ) -> *mut CppVecU8;
   fn v8_inspector__V8InspectorSession__schedulePauseOnNextStatement(
     session: *mut RawV8InspectorSession,
     break_reason: StringView,
@@ -801,6 +813,41 @@ impl V8InspectorSession {
     }
   }
 
+  /// Wraps a V8 value in an Inspector `Runtime.RemoteObject`.
+  ///
+  /// With `generate_preview == false`, V8 uses `kIdOnly`: object results still
+  /// include metadata such as `className` and `description`, but no property
+  /// preview. With `true`, V8 uses `kPreview`, which additionally includes a
+  /// property preview.
+  ///
+  /// `context` is used only to obtain an execution-context ID and look up the
+  /// corresponding `InspectedContext` in this session's context group. The
+  /// selected `InjectedScript` wraps `value` in that inspected context's stored
+  /// V8 context. Returns `None` if no matching inspected context is registered,
+  /// for example before `V8Inspector::context_created`.
+  ///
+  /// `_scope` is intentionally unused by Rust; borrowing it keeps the caller's
+  /// V8 `HandleScope` alive while `ValueMirror::create` allocates local handles.
+  pub fn wrap_object<'s>(
+    &self,
+    _scope: &mut PinScope<'s, '_>,
+    context: Local<'s, Context>,
+    value: Local<'s, Value>,
+    object_group: StringView,
+    generate_preview: bool,
+  ) -> Option<RemoteObject> {
+    unsafe {
+      UniqueRef::try_from_raw(v8_inspector__V8InspectorSession__wrapObject(
+        self.raw.as_ptr(),
+        &*context,
+        &*value,
+        object_group,
+        generate_preview,
+      ))
+      .map(|raw| RemoteObject { raw })
+    }
+  }
+
   pub fn schedule_pause_on_next_statement(
     &self,
     reason: StringView,
@@ -838,6 +885,38 @@ impl V8InspectorSession {
 impl Drop for V8InspectorSession {
   fn drop(&mut self) {
     unsafe { v8_inspector__V8InspectorSession__DELETE(self.raw.as_ptr()) };
+  }
+}
+
+/// An opaque, owned Inspector `Runtime.RemoteObject`.
+pub struct RemoteObject {
+  raw: UniqueRef<RawRemoteObject>,
+}
+
+#[repr(C)]
+#[derive(Debug)]
+struct RawRemoteObject(Opaque);
+
+impl RemoteObject {
+  /// Serializes this remote object to its CRDTP/CBOR representation.
+  pub fn to_bytes(&self) -> Vec<u8> {
+    unsafe {
+      CppVecU8::take_from_raw(v8_inspector__RemoteObject__toBytes(
+        self.raw.as_ptr(),
+      ))
+    }
+  }
+}
+
+impl Debug for RemoteObject {
+  fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+    f.debug_struct("RemoteObject").finish()
+  }
+}
+
+impl Drop for RawRemoteObject {
+  fn drop(&mut self) {
+    unsafe { v8_inspector__RemoteObject__DELETE(self) }
   }
 }
 
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -7263,6 +7263,98 @@ fn inspector_release_object_group() {
 }
 
 #[test]
+fn inspector_wrap_object() {
+  let _setup_guard = setup::parallel_test();
+  let isolate = &mut v8::Isolate::new(Default::default());
+
+  use v8::inspector::*;
+
+  let inspector_client = V8InspectorClient::new(Box::new(ClientCounter::new()));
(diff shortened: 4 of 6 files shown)


bartlomieju (commented):
The Rust-side design here is good: the null return from `wrapObject` is correctly modeled as `Option`, ownership goes through `UniqueRef` + `impl Drop for RawRemoteObject` (same idiom as `RawSlot` in `isolate.rs:2746`), the vector allocated in `binding.cc` is freed by the matching `delete` in `crdtp_binding.cc:256`, and the generated protocol classes genuinely don't leak into the public API. I also confirmed the build ordering is safe — `build_binding()` runs after `build_v8()` inside the `V8_FROM_SOURCE` branch, and the prebuilt path never invokes bindgen, so no consumer breaks.

My comments are mostly about the build/CI half, which I suspect is larger than it needs to be — see the note on `RawRemoteObject`, which may let the entire build.rs delta (and the CI cache change) disappear. The one thing I'd want resolved before merge is the cache-key restructure.

**Merge coordination:** this is the third PR in the series (#2038, #2040, #2041) making byte-identical edits to `ChannelCounterState` — adding `responses: Vec<String>` and rewriting `send_response`. Whichever two land later will conflict in `tests/test_api.rs`. Landing that shared test-harness change as its own small PR first would make all three trivially rebasable.

src/inspector.rs
```diff
@@ -651,6 +686,31 @@ impl Drop for V8InspectorSession {
   }
 }
 
+/// An opaque, owned Inspector `Runtime.RemoteObject`.
+pub struct RemoteObject {
+  raw: UniqueRef<RawRemoteObject>,
+}
+
+#[repr(transparent)]
+struct RawRemoteObject(RawRemoteObjectLayout);
```
bartlomieju: **This may be the change that lets most of the build plumbing go away.**

`RawRemoteObject` wraps a bindgen *opaque* type whose pointee is never dereferenced from Rust — the only operations are pass-the-pointer, `toBytes`, and `DELETE`. Every other raw inspector type in this file uses the crate's own marker instead:

```rust
pub struct RawV8InspectorSession(Opaque);
```

If `struct RawRemoteObject(Opaque);` works here — and I can't see what would break — then this whole chain disappears with it:

- `binding.hpp`'s `#include <v8-inspector-protocol.h>` and the `using v8_inspector__RemoteObject = ...` alias
- the `allowlist_item` / `opaque_type` entries in build.rs
- the `-I{gn_out}/gen/v8/include` clang arg in build.rs
- and, most likely, the CI cache change entirely

`binding.cc` still needs the header (it calls `wrapObject` and `AppendSerialized`), so the `BUILD.gn` include-dir addition stays — that part is real, and `$target_gen_dir/v8/include` correctly matches the `gn_out/gen/v8/include` path build.rs computes. But the *bindgen* side is paying real complexity for a pointer target. Worth trying; if there's a reason `Opaque` doesn't work, that reason belongs in a comment here.

.github/workflows/ci.yml
```diff
@@ -405,8 +405,9 @@ jobs:
             target/*/.*
             target/*/build
             target/*/deps
-          key: cargo1-${{ matrix.config.target }}-${{ matrix.config.variant }}-${{ matrix.config.v8_enable_pointer_compression }}-${{ matrix.config.simdutf }}-${{ hashFiles('Cargo.lock', 'build.rs', 'git_submodule_status.txt') }}
-          restore-keys: cargo1-${{ matrix.config.target }}-${{ matrix.config.variant }}-${{ matrix.config.v8_enable_pointer_compression }}-
+          # Never restore a build script compiled from a different build.rs.
+          key: cargo1-${{ matrix.config.target }}-${{ matrix.config.variant }}-${{ matrix.config.v8_enable_pointer_compression }}-${{ hashFiles('build.rs') }}-${{ matrix.config.simdutf }}-${{ hashFiles('Cargo.lock', 'git_submodule_status.txt') }}
+          restore-keys: cargo1-${{ matrix.config.target }}-${{ matrix.config.variant }}-${{ matrix.config.v8_enable_pointer_compression }}-${{ hashFiles('build.rs') }}-
```
bartlomieju: **This is the part I'd most want justified before merge.**

`hashFiles('build.rs')` was already inside `key`, so an exact-match restore already required an identical build.rs. This change only affects `restore-keys` — the partial-match fallback. The consequence is that **any** PR touching build.rs now restores nothing at all, including `target/sccache`. On the jobs that also cache `gn_out`, that's a full cold V8 build across every matrix entry, on every push to that branch. That's an expensive permanent default to pay for a stale-artifact case.

The comment says "never restore a build script compiled from a different build.rs," but Cargo already fingerprints build.rs and reruns it when it changes — so what was the actual failure? If it was stale bindgen output, I'd bet the real bug is that **`src/binding.hpp` is not in the rerun-if-changed list**: build.rs:49-51 declares only `.gn`, `BUILD.gn`, and `src/binding.cc`. This PR adds an include to `binding.hpp`, which makes that gap newly load-bearing. Adding

```rust
println!("cargo:rerun-if-changed=src/binding.hpp");
```

is a targeted fix; invalidating all CI caches isn't.

Either way, this is an unrelated infrastructure change riding along with a feature PR — easier to review, and to revert, on its own.

src/inspector.rs
```diff
@@ -620,6 +634,27 @@ impl V8InspectorSession {
...
   }
 
+  /// Wraps a V8 value in an Inspector `Runtime.RemoteObject`.
+  pub fn wrap_object<'s>(
+    &self,
+    _scope: &mut PinScope<'s, '_>,
+    context: Local<'s, Context>,
+    value: Local<'s, Value>,
+    object_group: StringView,
+    generate_preview: bool,
```
bartlomieju, line 837: `generate_preview: false` doesn't mean "full object, minus the preview" — V8 maps it to `WrapMode::kIdOnly` vs `kPreview` (`v8-inspector-session-impl.cc:331`), so `false` yields a `RemoteObject` with essentially just `type` and `objectId`, no `description` or `className`. That's a genuinely surprising API shape and worth a line in the doc comment; embedders will otherwise expect the richer wrap.

src/inspector.rs
```diff
@@ -651,6 +686,31 @@ impl Drop for V8InspectorSession {
   }
 }
 
+/// An opaque, owned Inspector `Runtime.RemoteObject`.
+pub struct RemoteObject {
+  raw: UniqueRef<RawRemoteObject>,
+}
```
bartlomieju, line 894: Nit: no `Debug` impl. Most public types in this crate have one, and `#[derive(Debug)]` won't work through `UniqueRef<RawRemoteObject>` without one on the raw type — so it'd need a manual `f.debug_struct("RemoteObject").finish()`, like `Channel` and `V8InspectorClient` have in this same file.

build.rs
```diff
@@ -208,12 +208,18 @@ fn build_binding() {
...
+    .join("gen")
+    .join("v8")
+    .join("include");
   let mut clang_args = vec![
     "-x".to_string(),
     "c++".to_string(),
     "-std=c++20".to_string(),
     "-nostdinc++".to_string(),
     "-Iv8/include".to_string(),
+    format!("-I{}", inspector_protocol_dir.display()),
```
bartlomieju: This `-I` (and the allowlist/`opaque_type` pair below) only exists so bindgen can see `v8-inspector-protocol.h` for the opaque `RemoteObject` type. If `RawRemoteObject(Opaque)` works — see my comment on `src/inspector.rs:694` — the entire build.rs delta can be dropped, along with the `binding.hpp` include that pulls the generated headers into the bindgen TU.

build.rs
```diff
@@ -316,12 +322,14 @@ fn build_binding() {
...
     .allowlist_item("v8__.*")
     .allowlist_item("cppgc__.*")
+    .allowlist_item("v8_inspector__RemoteObject")
     .allowlist_item("RustObj")
     .allowlist_item("memory_span_t")
     .allowlist_item("const_memory_span_t")
     .allowlist_item("ExternalConstOneByteStringResource")
     .blocklist_item("cppgc.*Visitor")
     .blocklist_item("RustObj.*Trace")
+    .opaque_type("v8_inspector__RemoteObject")
```
bartlomieju: Same as above: these two entries exist only to produce a Rust type that is never dereferenced. `Opaque` would make them unnecessary.

src/binding.hpp
```diff
@@ -52,6 +53,8 @@ using v8__HeapCodeStatistics = v8::HeapCodeStatistics;
 using v8__GCType = v8::GCType;
 using v8__GCCallbackFlags = v8::GCCallbackFlags;
 using v8__Intercepted = v8::Intercepted;
+using v8_inspector__RemoteObject =
+    v8_inspector::protocol::Runtime::API::RemoteObject;
```
bartlomieju: This alias exists purely for bindgen — `binding.cc` uses the fully-qualified `v8_inspector::protocol::Runtime::API::RemoteObject` directly. It's the include above it (`v8-inspector-protocol.h` in the bindgen TU) that forces the generated-header include dir into the bindgen invocation. Both go away if `RawRemoteObject` uses `Opaque`.

src/binding.cc
```diff
@@ -3486,6 +3487,29 @@ void v8_inspector__V8InspectorSession__dispatchProtocolMessage(
...
+    v8_inspector::protocol::Runtime::API::RemoteObject* self) {
+  delete self;
+}
+
+std::vector<uint8_t>* v8_inspector__RemoteObject__toBytes(
+    const v8_inspector::protocol::Runtime::API::RemoteObject* self) {
+  auto bytes = std::make_unique<std::vector<uint8_t>>();
+  self->AppendSerialized(bytes.get());
+  return bytes.release();
+}
```
bartlomieju: Cosmetic: this open-codes `make_unique<std::vector<uint8_t>>()` where `crdtp__vec_u8__new()` (`crdtp_binding.cc:252`) already does exactly that, and is what the Rust side pairs `crdtp__vec_u8__DELETE` with. Reusing it would keep the allocation in one place. (The types and `new`/`delete` pairing are correct as written — just duplicated.)

src/crdtp.rs
```diff
@@ -208,7 +208,19 @@ struct DispatchResponseWrapper(Opaque);
...
+  pub(crate) unsafe fn take_from_raw(ptr: *mut Self) -> Vec<u8> {
+    unsafe {
+      let len = crdtp__vec_u8__size(ptr);
+      let mut result = vec![0u8; len];
+      crdtp__vec_u8__copy(ptr, result.as_mut_ptr());
+      crdtp__vec_u8__DELETE(ptr);
+      result
+    }
+  }
+}
```
bartlomieju, line 223: Nice extraction. Since you're already touching this type: `crdtp.rs` still has three hand-inlined copies of this exact size/copy/delete sequence — `Serializable::to_bytes` (~line 224), `json_to_cbor` (~623), and `cbor_to_json` (~641). Collapsing them onto `take_from_raw` would leave the file cleaner than you found it.

tests/test_api.rs
```diff
@@ -7183,6 +7185,66 @@ fn inspector_dispatch_protocol_message() {
...
+  let get_properties = format!(
+    r#"{{"id":1,"method":"Runtime.getProperties","params":{{"objectId":"{object_id}","ownProperties":true}}}}"#,
+  );
+  session
+    .dispatch_protocol_message(StringView::from(get_properties.as_bytes()));
+
+  let state = channel.state.borrow();
+  assert_eq!(state.responses.len(), 1);
+  assert!(state.responses[0].contains(r#""name":"answer""#));
+  assert!(state.responses[0].contains(r#""value":42"#));
```
bartlomieju: The round-trip through `Runtime.getProperties` is the right shape — it proves the objectId is genuinely registered in the session rather than just that bytes came back. Three gaps:

1. **The `None` path is never exercised.** `wrapObject` returning null is the one branch this PR deliberately models, and nothing tests it. Wrapping a value whose context isn't tracked by the session (before `context_created`, or a context in a different group) should yield `None`.
2. **`generate_preview: true` is untested** — and given the kIdOnly/kPreview split noted on `wrap_object`, that flag is the one most likely to be misunderstood.
3. No `inspector.context_destroyed(context)` at the end, unlike the neighbouring inspector tests.

The `split_once(r#""objectId":""#)` parsing is brittle, though consistent with #2040.

src/inspector.rs
satouriko: Added a manual `Debug` implementation for `RemoteObject`, matching the surrounding public wrapper types.

build.rs
satouriko: Removed both bindgen entries; the Rust side now uses `RawRemoteObject(Opaque)`.

src/inspector.rs
satouriko: Switched `RawRemoteObject` to the crate-local `Opaque` marker. The source build and wrap-object tests pass without exposing the generated protocol type to bindgen.

src/inspector.rs
satouriko: Documented that `false` selects the V8 ID-only wrap mode, while `true` requests a property preview.

.github/workflows/ci.yml
satouriko: Removed the cache-key commit from this PR. With `RawRemoteObject` using `Opaque`, this feature no longer changes `build.rs` or `binding.hpp`, so no cache workaround or `rerun-if-changed` addition is needed here.

src/binding.hpp
satouriko: Removed the generated-header include and type alias from `binding.hpp`; bindgen no longer needs to see the protocol type.

src/binding.cc
satouriko: Reused `crdtp__vec_u8__new()` from the C++ bridge so allocation and deletion stay paired with the existing CRDTP vector helpers.

src/crdtp.rs
satouriko: Updated `Serializable::to_bytes`, `json_to_cbor`, and `cbor_to_json` to use `CppVecU8::take_from_raw`.

tests/test_api.rs
satouriko: Added coverage for the `None` path before `context_created`, `generate_preview: true`, and context teardown. The existing round trip through `Runtime.getProperties` remains intact; I kept the `split_once` extraction consistent with #2040.

build.rs
satouriko: Removed the entire `build.rs` delta. Only the GN include path needed by `binding.cc` remains.

satouriko:
Rebased onto current main after #2040 landed. The shared `ChannelCounterState` response capture now comes from main, so this PR no longer carries a competing test-harness edit. The branch is reduced to one feature commit; the unrelated CI cache commit and bindgen plumbing have been removed.

bartlomieju (commented):
Nicely scoped, and the parts that could have gone wrong all check out: `API::RemoteObject` inherits `Exported`, which has a virtual destructor, so `delete` through the base pointer is fine; and not entering the context on the Rust side is correct because `InjectedScript::wrapObject` opens its own `v8::Context::Scope`. The `CppVecU8::take_from_raw` extraction is a nice drive-by. The `Runtime.getProperties` round-trip is the right test — it proves the object ID is real rather than just that CBOR came back.

One thing I'd like fixed before merge, plus a few nits.

### `generate_preview` doc comment

The doc on `wrap_object` is garbled and reads backwards:

> `false` uses V8's ID-only wrap mode rather than producing a full remote object without a preview; `true` includes a property preview.

The intent is right — `v8-inspector-session-impl.cc` maps `false` to `WrapOptions({WrapMode::kIdOnly})` and `true` to `kPreview` — but as written it's hard to parse and easy to read as the opposite. Suggestion:

```rust
/// With `generate_preview == false`, V8 uses `kIdOnly`: the result carries the
/// type and object ID only — not a full remote object minus the preview. With
/// `true` it uses `kPreview`, which additionally includes a property preview.
```

While you're there, two more doc lines worth adding:

- **When `None` is returned.** `wrapObject` returns `nullptr` when no `InjectedScript` is found for the context, i.e. when the context was never registered via `context_created`. The test covers this, but nothing documents it.
- **What `context` actually selects.** `InjectedScript::wrapObject` wraps against `m_context->context()`; the `context` argument is only used as a lookup key via `InspectedContext::contextId`. The signature reads as if the value is wrapped "in" that context, which isn't quite what happens.

### Nits

- `_scope` is unused and unexplained. It's doing real work — V8 allocates handles inside `ValueMirror::create` and needs a live `HandleScope` — but neither a reader nor rustdoc (which will render it as `_scope`) can tell. A one-line comment would cover it.
- `extern "C" std::vector<uint8_t>* crdtp__vec_u8__new();` is hand-declared at the top of `binding.cc` while the definition lives in `src/crdtp_binding.cc`, so nothing checks that the two agree. Since `v8_inspector__RemoteObject__toBytes` is the only user, moving that function into `crdtp_binding.cc` next to the definition seems cleanest.
- The `kIdOnly` assertion `!json.contains(r#""preview":"#)` is weaker than it looks: `kIdOnly` also drops `description`/`className`, so asserting the absence of `"description"` would catch a silent switch to `kNoPreview` where the current check wouldn't.
- The PR description says the include directory was added "to GN and bindgen", but only `BUILD.gn` changed — which is correct, since bindgen parses `src/binding.hpp` and that doesn't include `v8-inspector-protocol.h`. Just worth fixing the description.

Unrelated to the change: the `release aarch64-unknown-linux-gnu simdutf` leg failed at a 3h timeout. That job took ~1h13m on one of your other PRs, so it looks like runner slowness — worth a re-run to confirm.

satouriko:
Addressed this review in `7951b33`:

- rewrote the `generate_preview` documentation and documented both the unregistered-context `None` case and what the `context` argument selects;
- documented why the otherwise-unused `_scope` must keep the caller's V8 `HandleScope` alive;
- moved `v8_inspector__RemoteObject__toBytes` into `crdtp_binding.cc` next to `crdtp__vec_u8__new`;
- corrected the PR description to describe the GN-only generated-header plumbing and the current test coverage;
- reran formatting, clang-format, clippy, and the complete nextest suite locally (290/290 passed).

I did not add the suggested assertion that `description` is absent. With the V8 revision in this repository, the actual `kIdOnly` result is `{"type":"object","className":"Object","description":"Object","objectId":"..."}`. `ObjectMirror::buildRemoteObject` adds `className` and `description` for non-JSON object modes, while `kPreview` additionally adds the property preview. The test therefore continues to assert the behavior that distinguishes these two modes here: `preview` is absent for `kIdOnly` and present for `kPreview`.

The force-push also starts a fresh CI run, superseding the previously timed-out leg.

bartlomieju (approved):
LGTM

PR #2044, opened 2026-08-06
feat: support sharing isolates across threads via v8::Locker

ry:


src/isolate.rs
```diff
@@ -2336,6 +2406,34 @@ impl OwnedIsolate {
...
+  /// Convert this isolate into a [`crate::SharedIsolate`] that can be
+  /// locked and used from any thread. The isolate is exited on the
+  /// current thread; all further access goes through
+  /// [`crate::SharedIsolate::lock`].
+  ///
+  /// # Panics
+  ///
+  /// Panics if this is a snapshot-creator isolate, or if another isolate
+  /// is entered on top of this one on the current thread.
+  pub fn into_shared(self) -> crate::SharedIsolate {
```
bartlomieju: This is the one I'd like to resolve before merging. `SharedIsolate` is `Send + Sync` and `Locker: DerefMut<Target = Isolate>`, so everything hanging off the annex now moves between threads with no `Send` bound:

- `Isolate::set_slot<T: 'static>` / `get_slot` have no `Send`/`Sync` bound. `shared_isolate_concurrent_use` demonstrates the pattern deliberately with a `Global`; swap that for an `Rc<Cell<u32>>` and you have a data race reachable from entirely safe code.
- The annex's `finalizer_map` (`Box<dyn FnOnce>`), `create_param_allocations: Box<dyn Any>`, and every slot value get *dropped* on whichever thread drops the `SharedIsolate`. `shared_isolate_moves_between_threads` tears down on a thread that never created any of it.

I don't see a cheap sound fix — a `Send`-bounded slot API for shared isolates is the real answer and it's invasive. My suggestion is to make this an `unsafe fn` with the contract spelled out (no non-`Send` data in isolate slots, `set_data`, finalizers, or create params) rather than shipping it as a safe constructor.

Separately, this should probably also reject an isolate that already has live `Weak`s / a non-empty `finalizer_map`, and one with a cppgc heap attached — see the comments on `Weak::new_raw` and the `SharedIsolate` docs.

src/handle.rs
```diff
@@ -742,6 +756,13 @@ impl<T> Weak<T> {
     data: NonNull<T>,
     finalizer_id: Option<FinalizerId>,
   ) -> Self {
+    // Weak callbacks fire during GC on whichever thread holds a shared
+    // isolate's lock, racing the `WeakData` owned by this (non-Send)
+    // handle on its home thread.
+    assert!(
+      !unsafe { (*isolate).global_liveness().as_ref() }.is_shared(),
+      "v8::Weak is not supported on shared isolates"
```
bartlomieju: This guard is creation-time only: a `Weak` created *before* `into_shared` sails right through, and the unsoundness described in the comment above applies to it in full. The check probably wants to live in `into_shared` as well (reject a non-empty `finalizer_map` / live weaks), since that's the point where the isolate actually becomes shared.

src/isolate.rs
```diff
@@ -2140,6 +2156,60 @@ impl IsolateLiveness {
...
+  /// isolate's lock; otherwise queue it for the next lock acquisition.
+  ///
+  /// While the queue is `Some`, teardown has not started: its final drain
+  /// closes the queue under this mutex before the isolate is disposed, so
+  /// the isolate pointer read below stays valid for the duration of this
+  /// call.
+  pub(crate) fn reset_or_defer_global(&self, data: *const Data) {
+    let mut q = self.deferred_global_drops.lock().unwrap();
+    let Some(v) = q.as_mut() else { return };
+    let isolate = self.get_isolate_ptr();
```
bartlomieju, line 2249: `get_isolate_ptr()` is re-read here, so it can be null even though `Global::drop`'s null check passed a moment ago. `v8::Locker::IsLocked` dereferences the isolate to reach its thread manager, so that's a null deref.

Narrow window, cheap fix: treat null as "not on the isolate thread", or just return early — a disposed isolate means the cell is gone anyway.

src/isolate.rs
```diff
@@ -2140,6 +2156,60 @@ impl IsolateLiveness {
   pub(crate) fn get_isolate_ptr(&self) -> *mut RealIsolate {
     self.isolate.load(std::sync::atomic::Ordering::Relaxed)
   }
+
+  pub(crate) fn mark_shared(&self) {
```
bartlomieju, line 2198: The `Relaxed` ordering here is load-bearing in a way that isn't written down. It works out only because `into_shared` consumes an `OwnedIsolate`, which is `!Send` (it holds a `NonNull`), so `mark_shared` always runs on the home thread — meaning the only thread that could observe a stale `false` and take the immediate-reset path in `Global::drop` is the same thread that performed the store.

That argument gets more load-bearing in #2045, where `Global` becomes `Send`. Either use `Release`/`Acquire` here and in `is_shared`, or leave the reasoning in a comment so it survives the next refactor.

src/locker.rs
```diff
@@ -0,0 +1,172 @@
...
+/// `v8::Locker`, enters the isolate on the current thread, and yields a
+/// [`Locker`] guard that dereferences to [`Isolate`].
+///
+/// Limitations (enforced by panics):
+/// - [`crate::Weak`] handles are not supported on shared isolates.
+/// - Snapshot-creator isolates cannot be shared.
+/// - Cloning a [`crate::Global`] belonging to a shared isolate requires
+///   holding the lock on the current thread.
+///
+/// cppgc heaps attached to shared isolates are unsupported and unsound.
```
bartlomieju: The cppgc case is called out as unsound but isn't in the "enforced by panics" list above and isn't actually checked anywhere. An assert in `into_shared` that no cppgc heap is attached would be cheap and would keep this from being a footgun that only shows up as a crash.

src/locker.rs
```diff
@@ -0,0 +1,172 @@
...
+      locker
+        .global_liveness()
+        .as_ref()
+        .drain_deferred_global_drops();
+      locker
+    }
+  }
+}
+
+impl Drop for Locker<'_> {
```
bartlomieju, line 287: The queue is drained on acquire but not on release, so the final unlock leaves whatever accumulated during it sitting there until teardown. Draining here too (while we still hold the lock and are still entered) would keep the steady state tighter — the cells hold their JS objects alive in the meantime.

src/locker.rs
```diff
@@ -0,0 +1,172 @@
...
+
+impl Deref for Locker<'_> {
+  type Target = Isolate;
+  fn deref(&self) -> &Isolate {
+    unsafe { Isolate::from_raw_ref(&self.cxx_isolate) }
+  }
+}
+
+impl DerefMut for Locker<'_> {
+  fn deref_mut(&mut self) -> &mut Isolate {
```
bartlomieju, line 317: `Isolate::from_raw_ref_mut` already exists in isolate.rs and does exactly this with the `repr(transparent)` safety comment attached — worth using it here instead of an open-coded `transmute`.

bartlomieju (commented):
The core mechanism looks good. The deferred-drop queue with a `Some`/`None` "closed" state is a nice way to make late `Global` drops race-free against teardown, and the `close`-then-`dispose` ordering is correct in both interleavings I could construct. Test coverage is genuinely strong — suspended generator/async frames resumed across threads, JS→Rust→JS reentrancy, nested LIFO lockers.

One thing I'd like to settle before this lands, plus a handful of smaller items inline.

The blocking question is whether `into_shared` should be safe (see the inline comment on it): `SharedIsolate` is `Send + Sync` and `Locker` derefs to `Isolate`, so every piece of embedder state hanging off the annex now crosses threads with no `Send` bound — isolate slots, finalizer closures, create-param allocations. I don't think there's a cheap sound fix, so my suggestion is `unsafe fn into_shared` with the contract documented rather than a safe constructor.

Also missing tests, if you want them: `into_shared` panicking on a snapshot-creator isolate, and the LIFO assertion in `Locker::drop`.

Note I reviewed this by reading only — I didn't build or run the test suite.

src/handle.rs
```diff
@@ -992,6 +1014,17 @@ impl<T> Clone for Weak<T> {
 
 impl<T> Drop for Weak<T> {
   fn drop(&mut self) {
+    // `data` is `Some` iff this handle was created through `new_raw` and
+    // thus counted. SAFETY: we're in the isolate's thread because `Weak`
+    // isn't Send or Sync.
+    if self.data.is_some() {
```
bartlomieju, line 1096: `Weak::into_raw` unbalances this counter. It does `self.data.take()` and then lets `self` drop normally, so `data.is_some()` is already false by the time this runs and the decrement is skipped — but `new_raw` counted the handle on the way in.

When `into_raw` returns `Some`, it works out: `from_raw` doesn't increment either, and the eventual drop of the reconstructed `Weak` decrements once, so the pair nets to zero. When `into_raw` returns `None` — the value was GC'd and there's no pending finalizer, so the box is just dropped — the count never comes back down.

Effect is a false-positive panic in `into_shared` on an isolate with no live weaks at all, and it's permanent once it happens. Cleanest fix is probably to decrement in `into_raw` and increment in `from_raw`, so each counted `+1` is paired with the handle that actually owns it, rather than inferring ownership from `data.is_some()`.

src/isolate.rs
```diff
@@ -2336,6 +2429,57 @@ impl OwnedIsolate {
...
+  ///
+  /// A shared isolate migrates between threads together with everything
+  /// attached to it, without `Send` bounds the type system can check.
+  /// The caller must ensure that all embedder state hanging off the
+  /// isolate is `Send`: isolate slot values ([`Isolate::set_slot`]),
+  /// embedder data ([`Isolate::set_data`]), callbacks and their captured
+  /// state, and the allocations referenced by its `CreateParams`. All of
+  /// it may be accessed and eventually dropped on whichever thread holds
+  /// the lock or drops the `SharedIsolate`.
+  pub unsafe fn into_shared(self) -> crate::SharedIsolate {
```
bartlomieju: The contract as worded reads like a point-in-time check on the state the isolate has *right now*, but it's really an obligation for the isolate's whole remaining lifetime: `Locker` derefs to `&mut Isolate`, so a caller can `set_slot` a non-`Send` value long after `into_shared` returned and nothing catches it. Worth saying so explicitly — something like "this obligation extends to state added later through a `Locker`, not just state present at this call."

Minor: rustdoc convention puts `# Safety` before `# Panics`.

tests/test_api.rs
```diff
@@ -14195,3 +14195,436 @@ fn crdtp_dispatcher_cleanup_on_drop() {
...
+    .map(|s| s.as_str())
+    .or_else(|| err.downcast_ref::<&str>().copied())
+    .unwrap();
+  assert!(msg.contains("reverse order"));
+  // The failed drop never exited isolate A or released its v8::Locker,
+  // so A is wedged on this thread; unwind B cleanly and leak A rather
+  // than tearing it down in a state V8 would abort on.
+  drop(lb);
+  drop(shared_b);
+  std::mem::forget(shared_a);
```
bartlomieju: Worth being explicit in the comment about what state this leaves behind, since it's more than a leaked isolate: A's `v8::Locker` is still held by this thread and A is still entered, and `la`'s `Box<RawLocker>` is freed by the unwind without `~Locker()` ever running. That's benign in practice — V8's `ThreadManager` tracks the lock, not the `Locker` object, and libtest gives each test its own thread that then exits — but it's worth a line so nobody later "fixes" the `forget` into a `drop`.

The test itself is good; it's exactly the assertion I wanted covered.

bartlomieju (commented):
Thanks — this addresses everything from the last round. `unsafe fn into_shared` with the contract written out is the right call, and rejecting live weaks / pending finalizers / embedder cppgc heaps up front closes the gap where a `Weak` created before sharing slipped past the `Weak::new_raw` assert. The `Release`/`Acquire` pair, the null check in `reset_or_defer_global`, `from_raw_ref_mut`, and the drain in `Locker::drop` all look right.

I checked the new `live_weak_count` accounting against every path that constructs or consumes a `Weak` — `clone_raw` routes through `new_raw` so clones are counted, `empty()` leaves `data: None` so it's correctly skipped, and `from_raw`/`drop` balance out. One path doesn't, though; see the inline comment on the decrement.

I also confirmed `has_embedder_cpp_heap` can't go stale: `CreateParams::cpp_heap` is the only way to attach one (there's no `AttachCppHeap` binding, only `get_cpp_heap`), and `new_impl` is the only isolate-creation path that isn't already rejected as a snapshot creator.

Still read-only review — I haven't built or run the suite.

bartlomieju (approved):
LGTM. The deferred-drop design holds up under the interleavings I could construct, the teardown ordering is right, and the guardrails around what can be shared (snapshot creators, live weaks, pending finalizers, embedder cppgc heaps) are in the right place now that `into_shared` is `unsafe`.

One thing to pick up as a follow-up rather than a blocker: the `Weak::into_raw` / `live_weak_count` imbalance I described above. `into_raw` takes `data` and then lets `self` drop, so the decrement is skipped; the `into_raw` → `from_raw` round trip nets to zero, but `into_raw` returning `None` leaks a count permanently, and after that `into_shared` panics on an isolate with no live weaks. It's a false-positive panic rather than anything unsound, and it only bites embedders using the raw `Weak` API, so it doesn't need to hold this up — but it'll be a confusing bug report if it lands as-is.

The smaller notes (safety-doc wording, the comment on the drop-order test) are take-them-or-leave-them.

Reviewed by reading only — I haven't built this or run the test suite, so the usual caveat applies about anything that would only show up under a debug V8 build.

bartlomieju (approved):
LGTM. The deferred-drop design holds up under the interleavings I could construct, the teardown ordering is right, and the guardrails around what can be shared (snapshot creators, live weaks, pending finalizers, embedder cppgc heaps) are in the right place now that `into_shared` is `unsafe`.

One thing to pick up as a follow-up rather than a blocker: the `Weak::into_raw` / `live_weak_count` imbalance I described above. `into_raw` takes `data` and then lets `self` drop, so the decrement is skipped; the `into_raw` → `from_raw` round trip nets to zero, but `into_raw` returning `None` leaks a count permanently, and after that `into_shared` panics on an isolate with no live weaks. It's a false-positive panic rather than anything unsound, and it only bites embedders using the raw `Weak` API, so it doesn't need to hold this up — but it'll be a confusing bug report if it lands as-is.

The smaller notes (safety-doc wording, the comment on the drop-order test) are take-them-or-leave-them.

Reviewed by reading only — I haven't built this or run the test suite, so the usual caveat applies about anything that would only show up under a debug V8 build.

PR #2045, opened 2026-08-06
feat: make Global<T> Send + Sync

ry:
## Summary

- Make `Global<T>` `Send + Sync`, allowing global handles to be moved or shared between threads.
- Permit dropping a `Global` on any thread. When its isolate cannot be accessed, resetting the V8 handle cell is deferred to the next safe lock boundary or isolate teardown.
- Gate operations that touch V8 (`clone`, `hash`, same-isolate equality, and conversion to `Local`) on the isolate home thread or the current thread holding the shared isolate's `Locker`.
- Keep the non-shared `Global` hot path lock-free: it uses the shared flag, liveness pointer, and cached thread ID without taking the deferred-reset mutex.

## Safety model

- `Global<T>` being `Send + Sync` does not make the underlying V8 heap object concurrently accessible.
- `Global::open` is now unsafe because its returned reference can otherwise outlive the isolate or the `Locker` authorizing access.
- V8 heap-object wrapper types are `!Sync`, preventing shared references obtained from `Local` or `Global::open` from crossing threads.
- `Borrow<T> for Global<T>` is removed because a safe `Borrow` implementation cannot tie the returned reference to an isolate or `Locker` guard.
- Conversion to `SharedIsolate` asserts that the `OwnedIsolate` is still on its creation thread, in addition to documenting that requirement for unsafe callers.
- Thread identity uses a destructor-free TLS cache backed by never-reused process-local IDs. Its cold path does not call `std::thread::current()`, allocate, or panic during TLS teardown.

## Breaking changes

- Calls to `Global::open` require an unsafe block and must uphold its documented lifetime and threading invariants.
- `Global<T>` no longer implements `Borrow<T>`. Use `Local::new(scope, &global)` for V8 access. A `HashMap<Global<T>, _>` can no longer be queried with `&T`; code requiring allocation-free borrowed lookup should use an embedder-owned stable key.
- `clone`, `hash`, and comparisons that may access the same isolate panic when called away from the non-shared isolate's home thread or without the shared isolate's `Locker`. Comparisons between handles known to belong to different live isolates still return `false` without touching V8.
- Shared references to V8 heap-object wrapper types can no longer be sent between threads.

## Validation

- `cargo nextest run --release --lib --bins --tests --locked --no-fail-fast` — 343 passed
- `cargo clippy --release --all-targets --locked -- -D clippy::all`
- `cargo fmt --all -- --check`
- `cargo doc --no-deps --locked`


src/handle.rs
```diff
@@ -350,18 +350,23 @@ impl<T> Global<T> {
   }
 }
 
+// A `Global` only touches V8 through methods that either take a scope or
+// `&Isolate` argument (obtainable only on the isolate's thread or under
+// its Locker), or that are guarded through `IsolateLiveness` (`clone`,
+// `drop`).
+unsafe impl<T> Send for Global<T> {}
```
bartlomieju, line 422: This justification isn't quite true yet — `PartialEq` and `Hash` on `Global<T>` (handle.rs:534-566, just below) take neither a scope nor an `&Isolate`, and both dereference `self.data` and call straight into V8:

- `Global<Value> == Global<Value>` → `v8__Value__SameValue` (value.rs:518)
- `Hash for Global<Object>` → `v8__Object__GetIdentityHash` (object.rs:660), which can *mutate* the object to install the hash

Before this PR they were reachable only on the owning thread by construction. With `Send + Sync`, a `HashSet<Global<v8::Object>>` moved to a worker thread calls into V8 while another thread is running JS — a data race from safe code.

I think they need the same `on_isolate_thread()` assert `clone` got. Once they do, this comment becomes accurate as written.

src/handle.rs
```diff
@@ -374,12 +379,13 @@ impl<T> Drop for Global<T> {
...
+      } else if !liveness.is_shared() && liveness.on_isolate_thread() {
         // Destroy the storage cell that contains the contents of this Global.
         v8__Global__Reset(self.data.cast().as_ptr());
       } else {
-        // A shared isolate's lock may be held by another thread; release
-        // the cell now if we hold it, otherwise defer to the next lock.
+        // Another thread may own the isolate right now; release the cell
+        // immediately if we may touch it, otherwise defer to the next
+        // lock acquisition or isolate teardown.
         liveness.reset_or_defer_global(self.data.cast().as_ptr());
```
bartlomieju: For a non-shared isolate this branch is now reachable (that's the point of `Send`), but nothing ever drains the queue in that case: `drain_deferred_global_drops` is called from exactly one place, `Locker::new` in locker.rs. So a `Global` dropped off the home thread sits in the queue pinning its JS object until the isolate is disposed.

That's the headline use case of this PR. `global_send_across_threads` exercises precisely this path and passes only because it drops the isolate right after. For a long-lived isolate handing `Global`s to a thread pool it's unbounded heap growth.

Worth draining at a cheap isolate-thread checkpoint — the `on_isolate_thread()` branch just above, or `HandleScope::new`. Gate it on an `AtomicUsize`/`AtomicBool` length hint so the common case stays a relaxed load rather than a mutex acquisition.

src/isolate.rs
```diff
@@ -2168,8 +2183,20 @@ impl IsolateLiveness {
...
   }
 
-  /// Release a `Global`'s cell now if the current thread holds the
-  /// isolate's lock; otherwise queue it for the next lock acquisition.
+  /// True when the current thread may touch the isolate's handle
+  /// storage: it holds the isolate's `v8::Locker` (shared), or it is the
+  /// isolate's home thread (non-shared).
+  pub(crate) fn on_isolate_thread(&self) -> bool {
+    if self.is_shared() {
+      unsafe { crate::locker::v8__Locker__IsLocked(self.get_isolate_ptr()) }
```
bartlomieju: `get_isolate_ptr()` can be null here — callers like `Global::clone` reach this without a null check, and in `Global::drop` the null check is a separate earlier read. `v8::Locker::IsLocked` dereferences the isolate to reach its thread manager, so this is a null deref.

Same note I left on `reset_or_defer_global` in #2044; fixing it here covers both, since that function now delegates to this one.

bartlomieju (commented):
Reviewed on top of #2044. Moving the thread check into `IsolateLiveness::on_isolate_thread()` and using it uniformly from `clone`, `drop` and `reset_or_defer_global` is a real simplification over the shared-only check in the base PR, and hoisting `close_deferred_global_drops` into `prepare_annex_for_dispose` so both teardown paths get it is right.

Two things I think block this, both inline: `PartialEq`/`Hash` on `Global` call into V8 with no scope and no lock (so `Send` makes them reachable off-thread in safe code), and the deferred queue is never drained for non-shared isolates, which turns the headline use case of this PR into a leak. The rest is smaller.

One more that I couldn't anchor inline since it's outside the diff — `Global::open`:

```rust
pub fn open<'a>(&'a self, scope: &mut Isolate) -> &'a T
```

The returned `&'a T` is tied to `&'a self`, not to the `&mut Isolate` borrow, so under a `Locker` you can hold the reference past unlock and then read it while another thread runs a GC that relocates the object. This is pre-existing, but `Send + Sync` is what makes it cross-thread reachable, so it's at least worth a doc note here.

Small API question: is `Sync` needed, or is `Send` the actual ask? Dropping `Sync` narrows the surface a bit, though not for the eq/hash problem below, which is reachable by moving alone.

As with #2044, I read-reviewed this without building or running the tests.

src/isolate.rs
```diff
@@ -2120,8 +2122,20 @@ impl IsolateAnnex {
   }
 }
 
+thread_local! {
+  static CURRENT_THREAD_ID: std::thread::ThreadId =
```
bartlomieju: Non-`const` thread-local initializer, so every `on_isolate_thread()` pays a lazy-init check — and `Global::clone` is a hot path in deno_core, where this assert is now on every clone for ordinary non-shared isolates too. Probably fine, but worth a quick measurement before landing. `ThreadId` wraps a `NonZeroU64`, so a `Cell<u64>` with a const initializer would sidestep it if it does show up.

src/locker.rs
```diff
@@ -79,19 +79,12 @@ impl SharedIsolate {
 impl Drop for SharedIsolate {
   fn drop(&mut self) {
     // Ownership guarantees no outstanding `Locker` (they borrow `self`),
-    // but other threads may still be dropping `Global`s concurrently:
-    // drain the deferred queue and close it under the lock, then tear
+    // so no thread can touch the isolate concurrently: `Global` droppers
+    // on other threads only push onto the mutex-protected deferred
+    // queue, which `prepare_annex_for_dispose` drains and closes. Tear
     // down the same way `OwnedIsolate::drop` does.
     unsafe {
```
bartlomieju, line 174: Dropping the `Locker` from this path looks correct to me — `GlobalHandles::Destroy` has no isolate-entry requirement, and exclusive ownership of `SharedIsolate` means no other thread can be inside V8 — but this is exactly the kind of change a V8 debug build with DCHECKs enabled would catch if the reasoning is wrong. Worth confirming CI covers `is_debug=true` before landing, given the base PR is the first thing in the crate to use `v8::Locker` at all.

src/locker.rs
```diff
@@ -14,12 +15,27 @@ use crate::isolate::RealIsolate;
...
+  /// read instead of an FFI call into `v8::Locker::IsLocked` — it sits
+  /// on the hot path of every `Global` clone/drop/eq/hash for shared
+  /// isolates.
+  static LOCKED_ISOLATES: RefCell<Vec<*mut RealIsolate>> =
+    const { RefCell::new(Vec::new()) };
+}
+
+/// Whether the current thread holds the `v8::Locker` for `isolate` (via
+/// [`SharedIsolate::lock`]).
+pub(crate) fn thread_holds_lock(isolate: *mut RealIsolate) -> bool {
```
bartlomieju, line 40: With the `v8__Locker__IsLocked` extern removed here, the C++ side in binding.cc (added by #2044) no longer has a Rust caller. Worth dropping it in the same PR rather than leaving an orphaned binding, or keeping it with a comment saying why the TLS shadow is preferred so the next person doesn't "restore" the call.

src/locker.rs
```diff
@@ -149,14 +175,16 @@ impl Drop for Locker<'_> {
...
+        .maybe_drain_deferred_global_drops();
       assert!(
         std::ptr::eq(self.cxx_isolate.as_ptr(), v8__Isolate__GetCurrent()),
         "Locker dropped while its isolate was not the entered one; lockers \
          must be dropped in reverse order of creation"
       );
       v8__Isolate__Exit(self.cxx_isolate.as_ptr());
       v8__Locker__DESTRUCT(&mut *self.raw);
+      let popped = LOCKED_ISOLATES.with(|v| v.borrow_mut().pop());
+      debug_assert_eq!(popped, Some(self.cxx_isolate.as_ptr()));
```
bartlomieju: `debug_assert_eq!` means a mismatch silently corrupts the shadow in release, and a corrupt shadow is a soundness problem rather than a diagnostic one — `thread_holds_lock` returning a false positive lets `Global::drop` reset a cell without the lock.

The LIFO assert above does guarantee this can't happen today, so this is belt-and-braces. But given the consequence, either make it a real `assert!`, or make the removal position-independent (find `self.cxx_isolate` and `swap_remove` it) so the shadow self-heals.

bartlomieju (commented):
This addresses everything from the last round, and the two blocking items look properly fixed.

The `eq`/`hash` asserts are in the right place. I checked the disposed-isolate case too: `match_host` panics on `DisposedIsolate` before `eq` reaches the assert, so there's no path where a use-after-free slips past — `hash` keeps its own explicit check because it has no host comparison to lean on. Good.

The drain story is much better. `deferred_len` as a relaxed length hint with all writes under the mutex is the right shape, and the two new checkpoints (`Global::new_raw`, and the home-thread branch of `Global::drop`) cover the realistic patterns. One residual gap, not worth blocking on: a home thread that neither creates nor drops any further `Global`s — just runs JS on ones it already holds — still never drains. `HandleScope::new` would close that if it ever shows up in practice; the hint makes it cheap enough to consider.

Swapping `v8::Locker::IsLocked` for the `LOCKED_ISOLATES` TLS shadow is a good call for the hot path, and I convinced myself it can't diverge from V8's view: `Locker::new`/`Drop` are the only mutators, and the pop sits *after* the LIFO assert, so a wedged out-of-order drop correctly leaves the entry in place (the lock genuinely is still held at that point). Two small notes on it inline.

The `# Cost` section on `SharedIsolate` is a genuinely useful thing to have written down — the work-stealing-executor warning in particular.

As before, read-only review; not built or run.

src/isolate.rs
```diff
@@ -2136,14 +2138,40 @@ impl IsolateAnnex {
...
+  // lazy-init check; `Global::clone`/`drop` hit this on every call.
+  static CURRENT_THREAD_ID: std::cell::Cell<Option<std::thread::ThreadId>> =
+    const { std::cell::Cell::new(None) };
+}
+
+fn current_thread_id() -> std::thread::ThreadId {
+  CURRENT_THREAD_ID.with(|c| match c.get() {
+    Some(id) => id,
+    None => {
+      let id = std::thread::current().id();
```
bartlomieju: `std::thread::current()` is reachable here during TLS destruction: a `Global` that was sent to this thread and parked in a `thread_local!` gets dropped at thread exit, hits `on_isolate_thread()`, and this cell is still `None` because nothing else on this thread ever touched it.

On older rustc that panics with "use of std::thread::current() is not possible after the thread's local data has been destroyed" — it was relaxed fairly recently to return an unnamed `Thread` instead. Worth a check against the crate's MSRV; if it's a concern, `thread::current()` can be swapped for an address-based thread identity that has no such dependency.

(The `const`-init on `CURRENT_THREAD_ID` is doing real work here beyond perf, incidentally — `Cell<Option<ThreadId>>` has no destructor, so the thread-local itself is never torn down and can't panic on access. `LOCKED_ISOLATES` does have one, but it's only reachable under a live `Locker`, so that's fine.)

src/handle.rs
```diff
@@ -374,12 +391,14 @@ impl<T> Drop for Global<T> {
       if liveness.get_isolate_ptr().is_null() {
         // This `Global` handle is associated with an `Isolate` that has already
         // been disposed.
-      } else if !liveness.is_shared() {
+      } else if !liveness.is_shared() && liveness.on_isolate_thread() {
```
bartlomieju: Nit: `is_shared()` is now an `Acquire` load (from #2044) and this evaluates it twice per drop — once here, and again inside `on_isolate_thread()`. Free on x86, an `ldar` each on aarch64, on what is a very hot path in deno_core. Hoisting it into a local and threading it down would collapse it to one.

bartlomieju (approved):
LGTM. Both of the blocking items from my first pass are properly closed: `eq`/`hash` now assert the same thread condition as `clone` before they call into V8, and the drain checkpoints mean a non-shared isolate no longer accumulates deferred cells until teardown, which was the leak that would have bitten the headline use case.

The remaining notes are all optional — the orphaned `v8__Locker__IsLocked` binding, the `debug_assert_eq!` on the `LOCKED_ISOLATES` pop, the `thread::current()`-during-TLS-teardown question, and the double `Acquire` load in `Global::drop`. None of them need to hold this up; the `debug_assert!` is the one I'd most like to see promoted, purely because its failure mode is soundness rather than diagnostics.

Same caveat as on #2044: read-only review, not built or run.

bartlomieju (approved):
LGTM. Both of the blocking items from my first pass are properly closed: `eq`/`hash` now assert the same thread condition as `clone` before they call into V8, and the drain checkpoints mean a non-shared isolate no longer accumulates deferred cells until teardown, which was the leak that would have bitten the headline use case.

The remaining notes are all optional — the orphaned `v8__Locker__IsLocked` binding, the `debug_assert_eq!` on the `LOCKED_ISOLATES` pop, the `thread::current()`-during-TLS-teardown question, and the double `Acquire` load in `Global::drop`. None of them need to hold this up; the `debug_assert!` is the one I'd most like to see promoted, purely because its failure mode is soundness rather than diagnostics.

Same caveat as on #2044: read-only review, not built or run.

nathanwhit (changes requested):
I re-reviewed the rebased head `b81b1cc7bdff` against the updated #2044 design. The rebase fixed several stale-parent problems, and the full `cargo nextest run --no-fail-fast` suite passes (315/315), but two safe-code soundness holes still block the unconditional `Send + Sync` impl.

1. **`Borrow<T>` exports unguarded V8 access from a non-shared `Global`.**

   [`Borrow::borrow`](https://github.com/denoland/rusty_v8/blob/b81b1cc7bdff138260b0eadd72bff5f00f19cc4a/src/handle.rs#L584-L593) rejects shared isolates, but it does not check the non-shared isolate's home thread before returning `&T`. A safe program can leave the `OwnedIsolate` on thread A, move its `Global<String>` to thread B, and call `Borrow::borrow(&global).length()` there.

   Adding a home-thread assertion alone is not a complete fix: the V8 heap-object wrappers currently auto-implement `Sync`, so a `&T` obtained on the home thread can subsequently be sent to another scoped thread and used there without another check. My preferred fix is to remove `Borrow<T> for Global<T>` and require `Local::new(scope, &global)`. I tried that locally: it caused exactly three test compile errors (two heterogeneous `HashMap<Global<_>, _>` lookups and the explicit shared-Borrow rejection test); after adapting those tests, all 315 nextest tests pass. The larger alternative is to make the V8 heap-object types `!Sync`, keep shared borrowing prohibited, and assert the home thread here.

2. **`Local == Global` still bypasses the non-shared thread check.**

   The explicit `on_isolate_thread()` assertion exists only in the [`PartialEq` implementation where `Global` is the left operand](https://github.com/denoland/rusty_v8/blob/b81b1cc7bdff138260b0eadd72bff5f00f19cc4a/src/handle.rs#L638-L659). The generic [`Local`-left implementation](https://github.com/denoland/rusty_v8/blob/b81b1cc7bdff138260b0eadd72bff5f00f19cc4a/src/handle.rs#L624-L635) calls the Global's `Handle::assert_safe_to_access`, but [that hook only checks shared-isolate locking](https://github.com/denoland/rusty_v8/blob/b81b1cc7bdff138260b0eadd72bff5f00f19cc4a/src/handle.rs#L532-L555); it does nothing for a non-shared Global used off its home thread.

   This is exploitable through safe code by moving Global A to a worker, creating a Local in isolate B there, and evaluating `local_b == global_a`. [`HandleHost::match_host(..., None)` deliberately treats the unknown `Scope`/`Isolate` pairing as a match](https://github.com/denoland/rusty_v8/blob/b81b1cc7bdff138260b0eadd72bff5f00f19cc4a/src/handle.rs#L712-L735), so the implementation dereferences both values and calls V8 across isolates/off A's thread. The full non-shared/shared access assertion needs to live in the common Global handle hook (for both `Global` and `&Global`), or the Local path must otherwise prove the hosts before touching either value. Please add a reversed-operand regression test; the existing off-thread equality test only exercises `global == ...`.

3. **The soundness-critical Locker shadow is destructible TLS.**

   [`LOCKED_ISOLATES` is a `RefCell<Vec<_>>` TLS value and every access uses `LocalKey::with`](https://github.com/denoland/rusty_v8/blob/b81b1cc7bdff138260b0eadd72bff5f00f19cc4a/src/locker.rs#L29-L43), including [`Locker`/`RelockGuard` cleanup](https://github.com/denoland/rusty_v8/blob/b81b1cc7bdff138260b0eadd72bff5f00f19cc4a/src/locker.rs#L315-L360). TLS values are destroyed in reverse initialization order. A TLS container initialized before this key can later hold a `Locker` or shared `Global`; at thread exit, its destructor can call this key after the key has already been destroyed. `LocalKey::with` then panics, and a panic from a TLS destructor aborts the process.

   Since nested Lockers are prohibited, this can be a non-destructible `Cell<*mut RealIsolate>` rather than a `Vec`, or the implementation can retain authoritative `v8::Locker::IsLocked`. At minimum, query paths need teardown-safe access, but the mutation/cleanup paths also need a design that cannot panic after acquiring the real V8 lock.

I am not re-raising the older `Global::open`, shared teardown, forgotten-Locker, or Unlocker transition findings: the rebased code correctly retains #2044's shared-open rejection, real Locker around queued teardown resets, Arc-backed leak safety, and explicit Exit/Enter plus relock draining.

bartlomieju:
All three fixed, and rebased onto main now that #2044 landed. I'm an author on
this branch now, so please disregard my earlier approval — your re-review gates
it.

**1. `Borrow`** — agreed, no check can rescue it: `fn borrow(&self) -> &T` has
nowhere to take proof the caller may touch the isolate, and nowhere to bind the
reference's lifetime to that proof, so any guard expires while the `&T` lives
on. Removed. Your measurement matched exactly — the two map lookups now key on
`Global`, and the rejection test became a `compile_fail` case. On the break: deno
has 1392 `v8::Global<` uses and four `Global`-keyed maps, all looked up by
`&Global`, so zero breakage there.

**2. `Local == Global`** — moved the full check into
`IsolateLiveness::assert_access_allowed`, now the single gate (shared: Locker
held; non-shared: home thread) behind `assert_safe_to_access` for both `Global`
and `&Global`. Dropped the duplicated inline asserts in `clone`/`hash`/`eq`, so
`match_host(.., None)` no longer matters for soundness — both operands are gated
before any deref. Added the reversed-operand test; two tests that asserted on the
old per-operation wording now assert on the unified message.

**3. `LOCKED_ISOLATES`** — you're right, and I had this wrong when it went in; I
claimed it was safe because it was only reachable under a live `Locker`, which
misses exactly your case. Went with retaining `v8::Locker::IsLocked` over the
`Cell`: nested `Locker`s for *different* isolates are supported (there's a LIFO
test), so it needs set membership rather than one slot, and a non-destructible
stack means leaking the storage per thread. Asking V8 is authoritative and also
drops the shadow's other obligation — staying in sync across `unlock` windows.
Costs an FFI call back on the hot path; separate change if it ever profiles.

**Open:** left `Sync` on `Global` since you didn't re-raise it. With `Borrow`
gone the surface is just the gated `clone`/`eq`/`hash` — happy to narrow to
`Send` only if you'd prefer.

bartlomieju:
Design looks sound to me — `OwnedIsolate` being `!Send` (it holds a `NonNull`) makes
the home-thread tracking well-founded, `ThreadId` is never reused, and `open` /
`Local::new` still go through `assert_match_isolate` with an `&mut Isolate`, so `Sync`
doesn't open a hole there. Four things I'd like other eyes on before this lands.

### 1. `assert_access_allowed` takes a mutex on every clone / hash / eq, including non-shared isolates

This is my main concern. The old `assert_shared_access` started with
`if !self.is_shared() { return }` — one relaxed load. Now every `Global::clone`,
`Hash::hash`, and *both operands* of `PartialEq` acquire the per-isolate
`deferred_global_resets` mutex. `Global::clone` is very hot in Deno, and
`HashMap<Global<T>, V>` pays one lock per hash plus one per eq per probe. It's also a
real contention point, since off-thread droppers take the same mutex.

I don't think the mutex is needed for the non-shared case: teardown of a non-shared
isolate happens on the home thread (`OwnedIsolate: !Send`), so being on the home thread
already excludes racing teardown. Something like:

```rust
pub(crate) fn assert_access_allowed(&self) {
  if !self.is_shared() {
    assert!(self.on_home_thread(),
      "accessing a Global requires being on its isolate's thread");
    assert!(!self.get_isolate_ptr().is_null(),
      "attempt to access Handle hosted by disposed Isolate");
    return;
  }
  // ... existing mutex-guarded shared path
}
```

Compounding this: dropping `Borrow<T> for Global<T>` means `HashMap<Global<T>, V>` can
now only be looked up with an *owned* `Global`, i.e. a real V8 handle allocation plus a
drop per lookup — as the test change itself shows (`map[&*obj]` →
`map[&v8::Global::new(scope, obj)]`).

### 2. The behavioral break isn't documented anywhere user-facing

This is semver-major twice over: the `Borrow` removal, and `clone`/`hash`/`eq` on a
*non-shared* isolate's `Global` now panicking off-thread where they previously worked.
The public rustdoc on `Global` says nothing about `Send`/`Sync` or the new panic
conditions — only the internal comments do. The PR body is empty too. Worth writing
both up.

Two smaller observable changes in the same vein: `Hash` lost its dedicated
"can't hash Global after its host Isolate has been disposed" message, and `PartialEq`
now panics rather than returning `false` for cross-isolate/disposed comparisons
(`assert_safe_to_access` runs before `match_host`).

While here, a few stale doc references to the now-removed impl: the `Global` rustdoc
("`Global::open` and `Borrow::borrow` are not supported for Globals that belong to a
shared isolate"), the `try_into_shared` docs ("obtained from a `Global` through
`Global::open` or `std::borrow::Borrow`"), and `assert_access_allowed` has two stacked
doc comments where the old shared-only paragraph was left above the new one.

### 3. `mark_shared` should assert the home thread

`Drop`'s fast path (`!is_shared() && on_home_thread()` → immediate `v8__Global__Reset`
with no mutex) is only safe because `try_into_shared` can't run concurrently on another
thread. That holds today via `OwnedIsolate: !Send`, but an `OwnedIsolate` reconstructed
from a raw pointer on a different thread would break it silently. A
`debug_assert!(self.on_home_thread())` in `mark_shared`, or a documented safety clause
on `try_into_shared`, would make the invariant enforceable rather than incidental.

### 4. TLS-destructor abort risk is acknowledged but not mitigated

`current_thread_id()`'s cold path calls `std::thread::current()`, which historically
panics after TLS teardown — and a panic in a TLS destructor aborts. The comment says
"on current rustc it returns an unnamed handle there rather than panicking". Since the
failure mode is an abort rather than a panic, I'd rather see that pinned as an MSRV
note in rust-toolchain/CI than asserted in a code comment. The realistic trigger is a
thread that receives a `Global` via `Send`, never touches it, and drops it from a TLS
destructor — `CURRENT_THREAD_ID` is cold there.

nathanwhit (approved):

PR #2049, opened 2026-08-07
feat: expose inspector object unwrapping

### LINKED ISSUE #2033: Expose V8InspectorSession remote-object lifecycle APIs

satouriko:
Closes #2033.

## Summary

Expose V8 Inspector `V8InspectorSession::unwrapObject` to Rust embedders, completing the remote-object lifecycle API discussed in #2033.

- add `V8InspectorSession::unwrap_object` with a scope-bound `Result<(Local<Value>, Local<Context>, UniquePtr<StringBuffer>), UniquePtr<StringBuffer>>` API
- keep the C++ bool and out-parameters inside the FFI bridge, constructing Rust `Local` handles only after V8 reports success
- transfer error and object-group `StringBuffer` ownership directly through `UniquePtr` without copying into Rust strings
- model the successful object-group output without an outer `Option`, because V8 always supplies a non-null buffer, including an empty buffer for an empty group name
- document invalid and stale object IDs together with the caller `HandleScope` requirement
- test wrap/unwrap identity, context and object-group round trips, plus the invalid-object-ID error path

This intentionally contains only `unwrapObject`; `releaseObjectGroup` landed in #2040 and `wrapObject` landed in #2041.

## Testing

- `cargo fmt --check`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=third_party/rust-toolchain/lib cargo nextest run --all-targets --locked --no-fail-fast`

### DIFF
--- a/src/inspector.rs
+++ b/src/inspector.rs
@@ -95,6 +95,14 @@ unsafe extern "C" {
     object_group: StringView,
     generate_preview: bool,
   ) -> *mut RawRemoteObject;
+  fn v8_inspector__V8InspectorSession__unwrapObject(
+    session: *mut RawV8InspectorSession,
+    error: *mut *mut StringBuffer,
+    object_id: StringView,
+    value: *mut *const Value,
+    context: *mut *const Context,
+    object_group: *mut *mut StringBuffer,
+  ) -> bool;
   fn v8_inspector__RemoteObject__DELETE(this: *mut RawRemoteObject);
   fn v8_inspector__RemoteObject__toBytes(
     this: *const RawRemoteObject,
@@ -846,6 +854,55 @@ impl V8InspectorSession {
       ))
       .map(|raw| RemoteObject { raw })
     }
+  }
+
+  /// Resolves an Inspector-generated object ID to its V8 value and context.
+  ///
+  /// On success, the returned object group contains the group name supplied
+  /// when the object was wrapped. An empty group name is returned as a
+  /// non-null buffer containing an empty string. On failure, the error buffer
+  /// is non-null and contains the Inspector's error message. This includes an
+  /// invalid ID or an ID made stale by releasing its object group.
+  ///
+  /// `_scope` is intentionally unused by Rust; borrowing it keeps the caller's
+  /// V8 `HandleScope` alive and ties the returned local handles to that scope.
+  #[allow(clippy::type_complexity)]
+  pub fn unwrap_object<'s>(
+    &self,
+    _scope: &mut PinScope<'s, '_>,
+    object_id: StringView,
+  ) -> Result<
+    (
+      Local<'s, Value>,
+      Local<'s, Context>,
+      UniquePtr<StringBuffer>,
+    ),
+    UniquePtr<StringBuffer>,
+  > {
+    let mut error = std::ptr::null_mut();
+    let mut value = std::ptr::null();
+    let mut context = std::ptr::null();
+    let mut object_group = std::ptr::null_mut();
+    let success = unsafe {
+      v8_inspector__V8InspectorSession__unwrapObject(
+        self.raw.as_ptr(),
+        &mut error,
+        object_id,
+        &mut value,
+        &mut context,
+        &mut object_group,
+      )
+    };
+
+    if !success {
+      return Err(unsafe { UniquePtr::from_raw(error) });
+    }
+
+    Ok((
+      unsafe { Local::from_raw(value).unwrap() },
+      unsafe { Local::from_raw(context).unwrap() },
+      unsafe { UniquePtr::from_raw(object_group) },
+    ))
   }
 
   pub fn schedule_pause_on_next_statement(
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -7369,6 +7369,70 @@ fn inspector_wrap_object() {
 }
 
 #[test]
+fn inspector_unwrap_object() {
+  let _setup_guard = setup::parallel_test();
+  let isolate = &mut v8::Isolate::new(Default::default());
+
+  use v8::inspector::*;
+
+  let inspector_client = V8InspectorClient::new(Box::new(ClientCounter::new()));
+  let inspector = V8Inspector::create(isolate, inspector_client);
+
+  v8::scope!(let scope, isolate);
+  let context = v8::Context::new(scope, Default::default());
+  let scope = &mut v8::ContextScope::new(scope, context);
+
+  let name = StringView::from(&b""[..]);
+  inspector.context_created(context, 1, name, name);
+
+  let session = inspector.connect(
+    1,
+    Channel::new(Box::new(ChannelCounter::new())),
+    StringView::from(&b"{}"[..]),
+    V8InspectorClientTrustLevel::Untrusted,
+  );
+
+  let value = eval(scope, "({ answer: 42 })").unwrap();
+  let remote_object = session
+    .wrap_object(
+      scope,
+      context,
+      value,
+      StringView::from(&b"rusty-v8-test"[..]),
+      false,
+    )
+    .unwrap();
+  let json = String::from_utf8(
+    v8::crdtp::cbor_to_json(&remote_object.to_bytes()).unwrap(),
+  )
+  .unwrap();
+  let object_id = json
+    .split_once(r#""objectId":""#)
+    .unwrap()
+    .1
+    .split_once('"')
+    .unwrap()
+    .0;
+
+  let (unwrapped_value, unwrapped_context, object_group) = session
+    .unwrap_object(scope, StringView::from(object_id.as_bytes()))
+    .unwrap();
+  assert!(unwrapped_value.strict_equals(value));
+  assert_eq!(unwrapped_context, context);
+  assert_eq!(object_group.unwrap().string().to_string(), "rusty-v8-test");
+
+  let error = session
+    .unwrap_object(scope, StringView::from(&b"invalid"[..]))
+    .unwrap_err();
+  assert_eq!(
+    error.unwrap().string().to_string(),
+    "Invalid remote object id"
+  );
+
+  inspector.context_destroyed(context);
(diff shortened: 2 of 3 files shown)


src/inspector.rs
```diff
@@ -848,6 +856,57 @@ impl V8InspectorSession {
...
+    _scope: &mut PinScope<'s, '_>,
+    object_id: StringView,
+  ) -> Result<
+    (
+      Local<'s, Value>,
+      Local<'s, Context>,
+      Option<UniquePtr<StringBuffer>>,
+    ),
+    UniquePtr<StringBuffer>,
+  > {
```
bartlomieju, line 881: `Option<UniquePtr<StringBuffer>>` is doubly-nullable. `UniquePtr` is already the nullable variant — it has `is_null()`, `as_ref() -> Option<&UniqueRef<T>>`, and `UniquePtr::from_raw` (`src/support.rs:66`) maps a null pointer to the empty `UniquePtr` for you. Wrapping it in an `Option` means callers write `.unwrap().unwrap()` to reach the string (see the test), and it suggests a `None` case that can't actually occur (see my comment further down).

I'd drop the outer layer:

```suggestion
  ) -> Result<
    (Local<'s, Value>, Local<'s, Context>, UniquePtr<StringBuffer>),
    UniquePtr<StringBuffer>,
  > {
```

This may also let you drop the `#[allow(clippy::type_complexity)]` on the line above — worth re-running clippy to check. There's only one other instance of that allow in the repo (`src/wasm.rs:359`), so avoiding a second is nice if it works out.

I realize the `Option` shape was what #2033 sketched, so push back if you'd rather keep it — but it's much cheaper to settle now than after release.

src/inspector.rs
```diff
@@ -848,6 +856,57 @@ impl V8InspectorSession {
...
+    }
+
+    Ok((
+      unsafe { Local::from_raw(value).unwrap() },
+      unsafe { Local::from_raw(context).unwrap() },
+      if object_group.is_null() {
+        None
+      } else {
+        Some(unsafe { UniquePtr::from_raw(object_group) })
+      },
```
bartlomieju: This branch is dead code, and it's also exactly what `UniquePtr::from_raw` already does internally.

Because the shim always passes a non-null `objectGroup` out-param, `v8-inspector-session-impl.cc:288-290` runs `*objectGroup = StringBufferFrom(std::move(objectGroupString))` unconditionally on success. So `object_group` is never null here — an object wrapped with an empty group name comes back as a non-null buffer holding `""`, not as a null pointer.

With the return type change above:

```suggestion
      unsafe { UniquePtr::from_raw(object_group) },
```

src/inspector.rs
```diff
@@ -848,6 +856,57 @@ impl V8InspectorSession {
     }
   }
 
+  /// Resolves an Inspector-generated object ID to its V8 value and context.
+  ///
+  /// On success, the returned object group identifies the group supplied when
+  /// the object was wrapped. An invalid ID, or an ID made stale by releasing
+  /// its object group, returns the Inspector's error message.
```
bartlomieju: Could you note here that an object wrapped with an *empty* group name resolves to an empty string rather than an absent group? As written, "the returned object group identifies the group supplied when the object was wrapped" reads as though absence is representable, which it isn't (see my comment on the `is_null()` branch).

Same reasoning applies to the error side: on failure V8 always populates `*error` (`v8-inspector-session-impl.cc:281-287`), so the `Err` payload is never a null `UniquePtr` — worth stating, since the type doesn't guarantee it.

tests/test_api.rs
```diff
@@ -7368,6 +7368,73 @@ fn inspector_wrap_object() {
...
+    .unwrap()
+    .0;
+
+  let (unwrapped_value, unwrapped_context, object_group) = session
+    .unwrap_object(scope, StringView::from(object_id.as_bytes()))
+    .unwrap();
+  assert!(unwrapped_value.strict_equals(value));
+  assert_eq!(unwrapped_context, context);
+  assert_eq!(
+    object_group.unwrap().unwrap().string().to_string(),
```
bartlomieju: The `.unwrap().unwrap()` here is the double-`Option` showing through at the call site. With the return-type change:

```suggestion
    object_group.unwrap().string().to_string(),
```

bartlomieju (commented):
Checked the shim against `v8/include/v8-inspector.h:208` and `v8/src/inspector/v8-inspector-session-impl.cc:274` — ownership transfer and the success/failure ordering are correct, and I don't see a memory-safety issue. One API-shape request before this lands, left inline: the `Option<UniquePtr<StringBuffer>>` has two redundant layers of nullability, and the `None` arm is unreachable.

src/inspector.rs
satouriko: Updated the docs to state both invariants explicitly: an empty group is a non-null buffer containing an empty string, and failures carry a non-null error buffer.

src/inspector.rs
satouriko: Done — I removed the dead null branch and now construct the `UniquePtr` directly from `object_group`.

src/inspector.rs
satouriko: Done — the success payload now returns `UniquePtr<StringBuffer>` directly. I also tried removing the `clippy::type_complexity` allow and re-ran Clippy, but the `Result` tuple still triggers the lint, so I kept the narrowly scoped allow.

tests/test_api.rs
satouriko: Done — the test now uses a single `unwrap()`.

bartlomieju (approved):
Thanks!

PR #2058, opened 2026-08-11
build: support explicit Linux bindgen sysroots

dills122:
## Summary

- Allow Linux source builds to provide bindgen's Clang resource directory and target libc headers explicitly.
- Preserve existing automatic discovery and musl behavior when the new overrides are unset.

## Motivation

`rusty_v8` expects a system `libclang`, but its Linux bindgen setup currently assumes that `LIBCLANG_PATH` has a sibling `bin/clang` and that target libc headers are installed in host search paths. Split or hermetic cross-toolchains do not necessarily use either layout.

The generic bindgen extra-argument variables can work around this, but the build already has target-aware Linux setup and a dedicated musl sysroot input. Keeping these inputs in `build.rs` makes the GNU and musl paths explicit, target-scoped, documented, and tracked by Cargo.

## What Changed

- Add `RUSTY_V8_BINDGEN_RESOURCE_DIR` for installations where matching Clang resources cannot be discovered beside `LIBCLANG_PATH`.
- Add `RUSTY_V8_GLIBC_PREFIX` for GNU cross-toolchains whose target headers are outside host search paths.
- Keep `RUSTY_V8_GLIBC_PREFIX` and `RUSTY_V8_MUSL_SYSROOT` scoped to their respective target environments; bindgen continues to derive the Clang target from Cargo's `TARGET`.
- Register all three explicit bindgen inputs with Cargo's environment-change tracking. This also fixes the existing omission of `RUSTY_V8_MUSL_SYSROOT`, so changing it now retriggers the build script.
- Document why explicit resources use `-resource-dir`, musl uses `--sysroot`, and GNU prefixes use `-isystem<prefix>/include`.
- Add focused tests for GNU prefix handling, GNU resource-only handling, musl resource-plus-sysroot handling, and target-environment scoping.
- Document the variables and their expected directory layouts.

## Validation

- `cargo fmt --all -- --check`
- `git diff --check`
- `cargo clippy --locked --test build -- -D clippy::all`
- `cargo test --locked --test build -- --nocapture` — 12 passed
- The pre-rebase full local suite passed, including 292 API tests, UI tests, and doctests; the exact rebased head is covered by the upstream CI matrix now running on this PR.
- Locked bindgen 0.72 source verification confirms that, absent an explicit target argument, it reads Cargo's `TARGET` and injects the matching Clang `--target` for cross-builds.
- The original contribution passed upstream's complete CI matrix: [31548574780](https://github.com/denoland/rusty_v8/actions/runs/31548574780).
- Review-staging CI passed GNU x86_64 release with and without pointer compression, x86_64 and ARM64 musl releases, iOS device/simulator, and an available Apple ARM64 release. Four paid Intel macOS jobs could not start because paid Actions are not configured on the fork: [31700715004](https://github.com/Shrimpworks/rusty_v8/actions/runs/31700715004).
- An end-to-end x64-to-ARM64 builder run using the equivalent explicit inputs passed a network-disabled source build, bindgen generation, and QEMU test: [30925045754](https://github.com/Shrimpworks/rusty_v8/actions/runs/30925045754).

## Related Work

- #1595 documents the system-`libclang` requirement that this extends to split and hermetic toolchains. This PR does not reopen or close that issue.
- #2018 established the existing explicit musl sysroot path that this change preserves and keeps separate from GNU prefixes.

## Scope Notes

- The branch is based on exact upstream commit `00fce34ac8a6eb7cb2da7d8aa51fe1f0bca897c7` and changes only `README.md` and `build.rs`.
- All new behavior is opt-in; unset inputs retain current behavior.
- No CI workflow, V8 build configuration, or artifact publication changes are included.
- The end-to-end hermetic build used the equivalent inputs on v150.2.0; current v152 validation covers the focused build-script behavior and upstream's platform matrix.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -68,7 +68,7 @@ fn main() {
     "OUT_DIR",
     "RUSTY_V8_ARCHIVE",
     "RUSTY_V8_BINDGEN_RESOURCE_DIR",
-    "RUSTY_V8_GLIBC_SYSROOT",
+    "RUSTY_V8_GLIBC_PREFIX",
     "RUSTY_V8_MIRROR",
     "RUSTY_V8_MIRROR_TAG",
     "RUSTY_V8_MIRROR_FALLBACK",
@@ -252,20 +252,17 @@ fn build_binding() {
         clang_args.push(format!("-isystem{}/include", resource_dir.trim()));
       }
     }
-    // Parse the V8 headers against the musl sysroot. bindgen already targets
-    // the musl triple (from $TARGET), so without this it looks for the target
-    // arch's glibc multiarch headers, which aren't installed when cross-
-    // compiling (e.g. aarch64 glibc headers on an x86_64 runner).
+    // Add target libc headers for Linux cross-builds. bindgen already targets
+    // Cargo's triple (from $TARGET), but target headers may not be installed in
+    // the host's search paths.
     let target_env = env::var("CARGO_CFG_TARGET_ENV").unwrap_or_default();
-    let target_triple = env::var("TARGET").unwrap();
     let musl_sysroot = env::var("RUSTY_V8_MUSL_SYSROOT").ok();
-    let glibc_sysroot = env::var("RUSTY_V8_GLIBC_SYSROOT").ok();
+    let glibc_prefix = env::var("RUSTY_V8_GLIBC_PREFIX").ok();
     clang_args.extend(explicit_linux_bindgen_args(
       &target_env,
-      &target_triple,
       resource_dir.as_deref(),
       musl_sysroot.as_deref(),
-      glibc_sysroot.as_deref(),
+      glibc_prefix.as_deref(),
     ));
   } else if target_os == "windows" {
     // libclang otherwise discovers the runner's system Clang resource
@@ -348,12 +345,22 @@ fn build_binding() {
     .expect("Couldn't write bindings!");
 }
 
+/// Builds the explicit Linux header-search arguments for bindgen.
+///
+/// An explicit Clang resource directory uses `-resource-dir` so Clang resolves
+/// the complete builtin resource layout. The auto-discovery path only adds the
+/// discovered `include` directory because it is supplementing libclang's
+/// existing resource setup.
+///
+/// Musl receives `--sysroot` because its input is a complete sysroot. A GNU
+/// cross prefix instead stores target headers directly under `<prefix>/include`;
+/// treating it as a sysroot would make Clang search `<prefix>/usr/include`.
+/// bindgen supplies the Clang target from Cargo's `TARGET` environment variable.
 fn explicit_linux_bindgen_args(
   target_env: &str,
-  target_triple: &str,
   resource_dir: Option<&str>,
   musl_sysroot: Option<&str>,
-  glibc_sysroot: Option<&str>,
+  glibc_prefix: Option<&str>,
 ) -> Vec<String> {
   let mut args = Vec::new();
   if let Some(resource_dir) = resource_dir {
@@ -367,9 +374,8 @@ fn explicit_linux_bindgen_args(
       }
     }
     "gnu" => {
-      if let Some(sysroot) = glibc_sysroot {
-        args.push(format!("--target={target_triple}"));
-        args.push(format!("-isystem{sysroot}/include"));
+      if let Some(prefix) = glibc_prefix {
+        args.push(format!("-isystem{prefix}/include"));
       }
     }
     _ => {}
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -335,6 +346,35 @@ fn build_binding() {
     .expect("Couldn't write bindings!");
 }
 
+fn explicit_linux_bindgen_args(
+  target_env: &str,
+  target_triple: &str,
+  resource_dir: Option<&str>,
+  musl_sysroot: Option<&str>,
+  glibc_sysroot: Option<&str>,
+) -> Vec<String> {
```
bartlomieju, line 364: Could this get a `///` block? The function encodes some genuinely non-obvious toolchain policy that a future reader will otherwise have to reverse-engineer — specifically why musl gets `--sysroot` while GNU gets `-isystem<prefix>/include`.

For what it's worth I think the asymmetry is *correct*: a GNU cross prefix like `/usr/aarch64-linux-gnu` holds the headers at `include/` directly, whereas `--sysroot` would make Clang look under `<root>/usr/include`. But that reasoning currently lives only in the README, and the README isn't what someone editing this function will be reading.

Same for the resource dir: the explicit input is passed as `-resource-dir=` here while the auto-discovery path a few lines up uses `-isystem{dir}/include` (build.rs:244). I read that as deliberate — it matches what the Windows branch does at build.rs:267 — but the two Linux paths using different flags for the same concept will trip someone up.

build.rs
```diff
@@ -335,6 +346,35 @@ fn build_binding() {
...
+  match target_env {
+    "musl" => {
+      if let Some(sysroot) = musl_sysroot {
+        args.push(format!("--sysroot={sysroot}"));
+      }
+    }
+    "gnu" => {
+      if let Some(sysroot) = glibc_sysroot {
+        args.push(format!("--target={target_triple}"));
+        args.push(format!("-isystem{sysroot}/include"));
```
bartlomieju: I think this `--target` push is redundant and should come out. bindgen's `find_effective_target` already reads the `TARGET` env var and injects `--target=` when the caller hasn't supplied one, and `rust_to_clang_target` is the identity for `*-unknown-linux-gnu` — so this passes a duplicate of what bindgen would pass anyway.

That's also what the existing comment at build.rs:250-253 asserts ("bindgen already targets the musl triple (from `$TARGET`)"), which is why the musl arm above doesn't push one. As written, the two arms disagree with each other and the gnu arm disagrees with the comment sitting just above the call site.

```suggestion
        args.push(format!("-isystem{sysroot}/include"));
```

If it turns out this *did* matter in your ARM64 hermetic run, I'd rather keep it with a comment explaining what broke without it — that's a useful thing to have recorded. But then `target_triple` should probably be dropped from the signature if it goes, since it's the only use.

bartlomieju: Non-blocking, but worth settling now since env var names are public API and awkward to rename later: `RUSTY_V8_GLIBC_SYSROOT` isn't a sysroot — it's never passed to `--sysroot`, unlike its musl counterpart. `RUSTY_V8_GLIBC_PREFIX` would describe what it actually takes.

The counter-argument is symmetry with `RUSTY_V8_MUSL_SYSROOT`, which is fair. Happy either way, just don't want it to go unconsidered.

build.rs
```diff
@@ -1500,4 +1540,46 @@ edge [fontsize=10]
...
+        None,
+        Some("/opt/aarch64-linux-gnu"),
+      ),
+      vec![
+        "-resource-dir=/opt/llvm/lib/clang/21",
+        "--target=aarch64-unknown-linux-gnu",
+        "-isystem/opt/aarch64-linux-gnu/include",
+      ]
+    );
+  }
```
bartlomieju, line 1880: Two cases I'd like added here — both a couple of lines:

1. **gnu with `resource_dir` only, no sysroot.** That's a realistic configuration (split toolchain, host libc headers available), and it's the one where the `--target` question above actually has consequences.
2. **musl together with `resource_dir`,** pinning that `-resource-dir` is emitted regardless of `target_env` — right now the resource-dir arg and the musl arm are only ever exercised separately.

The scoping test below is good and covers the case I'd most worry about (glibc sysroot silently ignored on musl), so this is just filling in the grid.

bartlomieju (commented):
Thanks for this — and for the unusually thorough writeup and validation. Pulling the env-scoped logic out into a pure function so it can actually be unit-tested under the `[[test]] name = "build"` target is exactly the right shape for build-script code, and the opt-in design means the blast radius is close to zero for existing users.

Worth calling out that this also fixes a latent bug that isn't mentioned in the description: `RUSTY_V8_MUSL_SYSROOT` was never in the `rerun-if-env-changed` list, so changing it didn't retrigger a rebuild. Good to have that closed.

A few changes requested inline before this lands — mostly about recording *why* the flag choices differ, plus one line I think is redundant.

bartlomieju:
@dills122 this now looks good to me, could you please re-format the code? I did resolve conflicts, but introduced some formatting problem and I can no longer push to your fork.

dills122:
@bartlomieju should be all cleaned up now.

bartlomieju (approved):
Thanks!

PR #2060, opened 2026-08-13
feat: isolate groups

bartlomieju:
V8 partitions isolates into groups, each with its own pointer compression
cage. Embedders that want more than the 4GB of heap a single cage allows
have to spread their isolates across several groups, and rusty_v8 has had
no way to express that. This exposes `v8::IsolateGroup` along with
`Isolate::new_with_group`, leaving `Isolate::new` on V8's group-less
overload so the common path is unchanged.

Creating groups beyond the default one only works when V8 is built with
pointer compression in multi-cage mode, which is why #1695 stalled: no
configuration this crate ships enables it, so the feature would be inert
everywhere. The new `v8_enable_multi_cage` feature sets that up. Nothing
is published for it, so the build script refuses to proceed without
`V8_FROM_SOURCE` rather than quietly linking the shared-cage `_ptrcomp`
prebuilt and leaving `IsolateGroup::can_create_new_groups()` returning
false at run time. One from-source CI job covers the configuration
without publishing artifacts.

Supersedes #1695, whose branch had fallen 246 commits behind and predated
the scope rewrite; the code is reapplied rather than rebased. Original
work by @piscisaureus.


nathanwhit (approved):
LGTM

PR #2061, opened 2026-08-13
build: relax RUSTY_V8_MIRROR release-tag assumptions

bartlomieju:
This PR relaxes the assumptions that `RUSTY_V8_MIRROR` makes about the release
tag. Today the build script looks for artifacts only under
`<mirror>/v<crate version>/<file>`. That layout fails for development checkouts
whose version is not published yet, and for plain directories of downloaded
artifacts. This PR is the first of a series; a later PR relaxes the check-only
and rust-analyzer path.

Changes are in `build.rs`, `README.md`, and `.github/workflows/ci.yml`.

## Tag override

The new `RUSTY_V8_MIRROR_TAG` variable overrides the tag path segment. The
build script uses the value verbatim and does not prepend a `v`, so directories
such as `nightly` or `pr-1234` work:

```
RUSTY_V8_MIRROR_TAG=v152.1.0 cargo build
```

Because this pairs the checkout's Rust sources with a `librusty_v8.a` and
`src_binding_*.rs` generated from a possibly-different V8, the build emits a
`cargo:warning=` whenever the resolved tag differs from `v<crate version>`, and
the README calls out the ABI-mismatch risk.

## Templated mirror URLs

If the `RUSTY_V8_MIRROR` value contains a `{` character it is treated as a full
template instead of a base. Placeholders: `{tag}`, `{version}`, `{target}`,
`{profile}`, `{features}`, `{file}`. No placeholder ⇒ current
`{base}/{tag}/{file}` behaviour, so existing mirrors are unaffected.

## Candidate list

`static_lib_url() -> String` becomes `artifact_urls(file) -> Vec<Candidate>`.
Each candidate is tried in order and the first that exists is used: the mirror
URL (tag layout or template) if `RUSTY_V8_MIRROR` is set; a flat `{base}/{file}`
layout for a plain filesystem mirror used as a cache; and the upstream default
base. `RUSTY_V8_ARCHIVE` still short-circuits, for the static library only.

Each candidate carries whether it is a *deliberate source* (a pinned mirror, the
upstream release) or a best-effort *cache* (the flat filesystem layout, the
`~/.cargo/.rusty_v8` download cache). A `FetchError::Miss` (absent artifact —
404, missing file, no downloader) falls through to the next candidate; a
`FetchError::Fatal` (present but corrupt/unusable) aborts — **except** for cache
candidates, where corruption is demoted to a miss (and the bad cache entry
evicted) so the build falls through rather than hard-failing.

`RUSTY_V8_MIRROR_STRICT=1` stops the list after the mirror entries, for hermetic
CI that must not reach the network. It is honored unconditionally: with no
mirror set the list is empty and the build fails with an explicit error rather
than silently reaching upstream.

## Bindings: new capability

Previously a bare git checkout (which ships only `gen/.gitkeep`) could not build
in prebuilt mode at all — `cargo build` without `V8_FROM_SOURCE` or a mirror
always failed at `include!(env!("RUSTY_V8_SRC_BINDING_PATH"))`. This PR makes
that path work: the binding is fetched through the same candidate list and
written into `OUT_DIR` (never into the registry-owned source tree). A published
crate that ships `gen/<name>.rs` still uses it as-is with no fetch, so the
common case is unchanged. The `DOCS_RS`/RLS early-exit does not fetch a missing
binding (no network there), leaving the `include!` diagnostic in place.

A new `prebuilt-git-checkout` CI job exercises this path end to end; it
self-skips when the current crate version has no published release yet (e.g. a
version-bump PR).

## Errors, safety, and the checksum fix

- One panic at the end of the candidate list naming every URL tried and the
  escape hatches (`RUSTY_V8_ARCHIVE`, `RUSTY_V8_MIRROR`, `RUSTY_V8_MIRROR_TAG`,
  `V8_FROM_SOURCE=1`); a distinct message for the empty-list (strict) and Fatal
  cases.
- `copy_archive` stages to a sibling temp and renames on success, so a corrupt
  gzip or full disk can't truncate a previously-good `librusty_v8.a`.
- The checksum is written only after the copy fully succeeds, and the
  redundant-download check now compares against the requested URL rather than
  `static_lib_url()` (bindings were re-fetched every build; this also makes tag
  changes invalidate the cache correctly).
- The downloader is split into `download_with_{deno,python,curl}`; a missing
  `curl` is a miss, not a panic.

## Tests / housekeeping

`RUSTY_V8_MIRROR_TAG` and `RUSTY_V8_MIRROR_STRICT` are registered for
`rerun-if-env-changed`. `artifact_urls_from` is a pure function with an
ordering-matrix table test (no mirror; http; fs + flat cache; template; strict
with and without a mirror), plus a template-expansion test. The README
`RUSTY_V8_MIRROR` section is rewritten (tag override, template, fallback order,
strict mode) and the stale `v0.13.0`/pre-`.gz` cache example refreshed.

Behaviour is unchanged for a published crate with none of the new variables set.
Existing `<base>/v<version>/<file>` mirrors keep working. No public API change,
no change to the `V8_FROM_SOURCE` path, no new dependency.

Closes bartlomieju/orchid-inbox#148

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,9 +65,68 @@ We default to release builds of `v8` due to performance & CI reasons in `deno`.
 
 ## The `RUSTY_V8_MIRROR` environment variable
 
-Tells the build script where to get binary builds from. Understands `http://`
-and `https://` URLs, and file paths. The default is
-https://github.com/denoland/rusty_v8/releases.
+Tells the build script where to get the prebuilt static library and the
+generated bindings from. Understands `http://` and `https://` URLs, and file
+paths. The default is
+https://github.com/denoland/rusty_v8/releases/download.
+
+### Resolution order
+
+By default the build script downloads two artifacts — the static library
+(`librusty_v8_<features>_<profile>_<target>.a.gz`) and the bindings
+(`src_binding_<features>_<profile>_<target>.rs`) — from
+`<base>/<tag>/<file>`, where `<tag>` is `v<crate version>` (e.g. `v152.1.0`).
+For each artifact the following candidates are tried in order, and the first
+one that exists is used:
+
+1. `RUSTY_V8_ARCHIVE`, if set — a specific static library, static library only
+   (see below). Short-circuits everything else.
+2. The `RUSTY_V8_MIRROR` value, if set (see the layouts below).
+3. For a **filesystem** mirror laid out as a plain directory of artifacts, a
+   flat `<mirror>/<file>` lookup (no tag subdirectory).
+4. The upstream default releases, `<default base>/<tag>/<file>`.
+
+If nothing matches, the build fails with a message listing every location that
+was tried.
+
+### Overriding the tag
+
+`RUSTY_V8_MIRROR_TAG` overrides the `<tag>` path segment. The value is used
+**verbatim** (no `v` is prepended), so non-tag directories such as `nightly` or
+`pr-1234` also work:
+
+```bash
+# build a checkout whose Cargo.toml version isn't published yet using the
+# artifacts from the last published tag
+RUSTY_V8_MIRROR_TAG=v152.1.0 cargo build
+```
+
+### Templated mirrors
+
+If the `RUSTY_V8_MIRROR` value contains a `{` placeholder it is treated as a
+full URL/path template instead of a base. The following placeholders are
+substituted:
+
+| Placeholder  | Value                                                        |
+| ------------ | ------------------------------------------------------------ |
+| `{tag}`      | resolved tag (`RUSTY_V8_MIRROR_TAG` or `v<crate version>`)   |
+| `{version}`  | raw crate version, no `v` prefix                             |
+| `{target}`   | Rust target triple                                           |
+| `{profile}`  | `release` or `debug`                                         |
+| `{features}` | ` `, `_ptrcomp`, `_sandbox`, or `_ptrcomp_sandbox`           |
+| `{file}`     | full artifact filename                                       |
+
+```bash
+RUSTY_V8_MIRROR='https://my-cache.example.com/rusty_v8/{tag}/{file}' cargo build
+```
+
+### Strict mode
+
+`RUSTY_V8_MIRROR_STRICT=1` stops the candidate list after the mirror entries,
+so the build never falls back to the upstream releases. Use this for hermetic
+CI that must never reach the network.
+
+### File-based mirrors
 
 File-based mirrors are good for using cached downloads. First, point the
 environment variable to a suitable location:
@@ -75,26 +134,31 @@ environment variable to a suitable location:
     # you might want to add this to your .bashrc
     $ export RUSTY_V8_MIRROR=$HOME/.cache/rusty_v8
 
-Then populate the cache:
+Then populate the cache. The artifacts are gzip-compressed (`.gz`) and named per
+target and profile:
 
 ```bash
 #!/bin/bash
 
 # see https://github.com/denoland/rusty_v8/releases
 
-for REL in v0.13.0 v0.12.0; do
-  mkdir -p $RUSTY_V8_MIRROR/$REL
+for REL in v152.1.0 v152.0.0; do
+  mkdir -p "$RUSTY_V8_MIRROR/$REL"
   for FILE in \
-    librusty_v8_debug_x86_64-unknown-linux-gnu.a \
-    librusty_v8_release_x86_64-unknown-linux-gnu.a \
+    librusty_v8_release_x86_64-unknown-linux-gnu.a.gz \
+    src_binding_release_x86_64-unknown-linux-gnu.rs \
   ; do
-    if [ ! -f $RUSTY_V8_MIRROR/$REL/$FILE ]; then
-      wget -O $RUSTY_V8_MIRROR/$REL/$FILE \
-        https://github.com/denoland/rusty_v8/releases/download/$REL/$FILE
+    if [ ! -f "$RUSTY_V8_MIRROR/$REL/$FILE" ]; then
+      wget -O "$RUSTY_V8_MIRROR/$REL/$FILE" \
+        "https://github.com/denoland/rusty_v8/releases/download/$REL/$FILE"
     fi
   done
 done
 ```
+
+Artifacts may also be dropped straight into `$RUSTY_V8_MIRROR` with no tag
+subdirectory (the flat layout from step 3 above), which makes a directory of
+downloaded files usable as a cache without knowing the tag.
 
 ## The `RUSTY_V8_ARCHIVE` environment variable
 
--- a/README.md
+++ b/README.md
@@ -85,9 +85,20 @@ one that exists is used:
 3. For a **filesystem** mirror laid out as a plain directory of artifacts, a
    flat `<mirror>/<file>` lookup (no tag subdirectory).
 4. The upstream default releases, `<default base>/<tag>/<file>`.
+5. For the bindings only: the copy shipped inside the crate at
+   `gen/<file>`, if present. A published crate always ships it, so a mirror that
+   carries only the static library does not force a network fetch — and strict
+   mode does not fail — when the correct bindings are already on disk.
 
 If nothing matches, the build fails with a message listing every location that
 was tried.
+
+> ⚠️ **Untagged layout.** Candidate 3 has no tag in the path, and artifact
+> filenames do not encode a version, so an artifact left over from another
+> release is indistinguishable from the right one. When a build resolves through
+> the flat layout it emits a `cargo:warning=`; see the ABI mismatch note below
+> for why that matters. Use the `<mirror>/<tag>/<file>` layout if you keep more
+> than one version around.
 
 ### Overriding the tag
 
@@ -128,6 +139,10 @@ substituted:
 RUSTY_V8_MIRROR='https://my-cache.example.com/rusty_v8/{tag}/{file}' cargo build
 ```
 
+Only the placeholders above are recognised. A misspelled one would otherwise
+survive expansion and surface as an opaque 404, so the build fails immediately
+and names the offenders.
+
 ### Strict mode
 
 `RUSTY_V8_MIRROR_STRICT=1` stops the candidate list after the mirror entries,
@@ -137,6 +152,10 @@ CI that must never reach the network. It is honored unconditionally: if no
 fails with an explicit error rather than silently reaching upstream. A hermetic
 build that also needs the bindings without a mirror can point
 `RUSTY_V8_SRC_BINDING_PATH` at them directly.
+
+An artifact already fetched by an earlier build is reused, so the usual
+workflow of populating a cache once with network access and then building
+offline works without a mirror configured.
 
 ### File-based mirrors
 
--- a/README.md
+++ b/README.md
@@ -107,14 +107,14 @@ If the `RUSTY_V8_MIRROR` value contains a `{` placeholder it is treated as a
 full URL/path template instead of a base. The following placeholders are
 substituted:
 
-| Placeholder  | Value                                                        |
-| ------------ | ------------------------------------------------------------ |
-| `{tag}`      | resolved tag (`RUSTY_V8_MIRROR_TAG` or `v<crate version>`)   |
-| `{version}`  | raw crate version, no `v` prefix                             |
-| `{target}`   | Rust target triple                                           |
-| `{profile}`  | `release` or `debug`                                         |
-| `{features}` | ` `, `_ptrcomp`, `_sandbox`, or `_ptrcomp_sandbox`           |
-| `{file}`     | full artifact filename                                       |
+| Placeholder  | Value                                                                        |
+| ------------ | ---------------------------------------------------------------------------- |
+| `{tag}`      | resolved tag (`RUSTY_V8_MIRROR_TAG` or `v<crate version>`)                    |
+| `{version}`  | raw crate version, no `v` prefix                                             |
+| `{target}`   | Rust target triple                                                           |
+| `{profile}`  | `release`, or `debug` only when `V8_FORCE_DEBUG=true` on a non-Windows target |
+| `{features}` | empty, or `_ptrcomp`, `_sandbox`, `_ptrcomp_sandbox`                          |
+| `{file}`     | full artifact filename                                                       |
 
 ```bash
 RUSTY_V8_MIRROR='https://my-cache.example.com/rusty_v8/{tag}/{file}' cargo build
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -1064,10 +1201,15 @@ fn print_prebuilt_src_binding_path() {
...
-    let version = env::var("CARGO_PKG_VERSION").unwrap();
-    let url = format!("{base}/v{version}/{name}");
-    download_file(&url, &src_binding_path);
+  // Fetch the binding when a mirror is configured (existing behaviour), or when
+  // it is missing from this checkout. A git checkout has no `gen/<name>.rs`, so
+  // fetching it from the mirror/upstream turns a missing binding into a build
+  // script diagnostic rather than a confusing `include!` failure in
+  // `src/binding.rs`. A published crate ships `gen/<name>.rs`, so with no env
+  // vars set this is skipped and behaves exactly as before.
+  if env::var_os("RUSTY_V8_MIRROR").is_some() || !src_binding_path.exists() {
```
bartlomieju: `main()` calls `print_prebuilt_src_binding_path()` on the `DOCS_RS`/RLS early-exit path (build.rs:103) before returning. With the new `|| !src_binding_path.exists()` condition, those builds now attempt a network fetch, and `download_artifact` panics when it fails. docs.rs has no network, so any published crate that lacks `gen/src_binding_<features>_release_<target>.rs` for the exact feature combination docs.rs resolves (e.g. a docs build with `v8_enable_pointer_compression`) turns a previously-working docs build into a hard build-script panic.

Suggest skipping the fetch on the doc/RLS path, or making failure non-fatal there and letting the existing `include!` diagnostic stand.

Also note this makes the "zero behaviour change when no new env vars are set" claim not quite true: a plain `cargo build` in a git checkout now performs a network fetch it previously did not.

While here: `download_static_lib_binaries` does `fs::create_dir_all(&dir)` before downloading, but this path does not. `gen/.gitkeep` covers a git checkout, but a published crate that ships no binding for the target may have no `gen/` directory at all, and both the tmpfile write and `File::create` would then fail. One `fs::create_dir_all(parent)` closes that gap.

build.rs
```diff
@@ -888,42 +1005,49 @@ fn download_file(url: &str, filename: &Path) {
...
-  assert!(tmpfile.exists());
 
   // Write checksum (i.e url) & move file
-  fs::write(static_checksum_path(filename), url).unwrap();
-  copy_archive(&tmpfile.to_string_lossy(), filename);
-  fs::remove_file(&tmpfile).unwrap();
+  fs::write(static_checksum_path(filename), url).map_err(|e| {
+    format!("could not write checksum for {}: {e}", filename.display())
+  })?;
+  copy_archive(&tmpfile.to_string_lossy(), filename)?;
```
bartlomieju: The checksum is written *before* the copy succeeds. If `copy_archive` fails midway (corrupt gzip, disk full), `filename` has already been truncated/partially written by `File::create` + `io::copy`, but the `.sum` file already records `url`. The new `filename.exists() && sum == url` fast path above then reports `Already downloaded` on the next build and links a truncated `.a`.

This was survivable before because the failure panicked; now the candidate loop continues and leaves the bad state on disk. Please write the checksum only after `copy_archive` returns `Ok`, and remove any stale `.sum` on the error path.

build.rs
```diff
@@ -806,33 +881,73 @@ fn replace_non_alphanumeric(url: &str) -> String {
...
 }
 
-fn download_file(url: &str, filename: &Path) {
-  if !url.starts_with("http:") && !url.starts_with("https:") {
-    copy_archive(url, filename);
-    return;
+/// Fetch each candidate in `urls` into `filename`, stopping at the first that
+/// succeeds. If they all fail, panic with the full list of what was tried and
+/// the available escape hatches.
+fn download_artifact(urls: &[String], filename: &Path) {
```
bartlomieju: Every failure from `try_download_file` is treated as recoverable — "not found", "corrupt archive", "disk full" and "could not create destination" all just advance to the next candidate. A user with a deliberately pinned `RUSTY_V8_MIRROR` whose artifact is corrupt silently gets an upstream download instead of an error, which is the opposite of what pinning is for.

Suggest distinguishing *miss* (fall through to the next candidate) from *found but unusable* (abort). At minimum, `copy_archive`/decompress failures should be fatal rather than falling through.

build.rs
```diff
@@ -747,21 +749,94 @@ fn static_lib_name(suffix: &str) -> String {
...
+      // Flat layout `<base>/<file>` for a plain directory of downloaded
+      // artifacts used as a cache. Only meaningful for filesystem mirrors.
+      if !is_http_url(&mirror) {
+        candidates.push(format!("{mirror}/{file}"));
+      }
+    }
+  }
+
+  // Fall back to the upstream releases, unless strict mode forbids the network.
+  if !env_bool("RUSTY_V8_MIRROR_STRICT") {
```
bartlomieju: `RUSTY_V8_MIRROR_STRICT=1` with no `RUSTY_V8_MIRROR` set produces an empty candidate list, and `download_artifact` then panics with `Tried the following candidates, in order:` followed by nothing — a confusing dead end.

Either error early with an explicit "strict mode requires RUSTY_V8_MIRROR" message, or make strict a no-op when no mirror is configured.

build.rs
```diff
@@ -806,33 +881,73 @@ fn replace_non_alphanumeric(url: &str) -> String {
...
+/// Try to fetch a single `url` (an `http(s)://` URL or a filesystem path) into
+/// `filename`. Returns `Err` with a human-readable reason on a recoverable
+/// miss so the caller can fall through to the next candidate.
+fn try_download_file(url: &str, filename: &Path) -> Result<(), String> {
+  if !is_http_url(url) {
+    // A filesystem path: an explicit archive, a file mirror, or a flat cache.
+    if !Path::new(url).exists() {
+      return Err(format!("file not found: {url}"));
+    }
+    return copy_archive(url, filename);
```
bartlomieju: Filesystem candidates never write a `.sum`, so a file mirror re-copies (and re-decompresses) the ~100MB archive on every build. This is pre-existing behaviour, but the new flat-cache layout makes file mirrors much more prominent, so it's worth extending the checksum short-circuit to cover the filesystem path too.

build.rs
```diff
@@ -979,22 +1103,35 @@ where
...
 /// This is necessary because the V8 archive could live inside a read-only
 /// filesystem, and subsequent builds would fail to overwrite it.
-fn copy_archive(url: &str, filename: &Path) {
+fn copy_archive(url: &str, filename: &Path) -> Result<(), String> {
   println!("Copying {url} to {filename:?}");
-  let mut src = fs::File::open(url).unwrap();
-  let mut dst = fs::File::create(filename).unwrap();
+  let mut src = fs::File::open(url)
+    .map_err(|e| format!("could not open source archive {url}: {e}"))?;
+  let mut dst = fs::File::create(filename).map_err(|e| {
```
bartlomieju: `File::create(filename)` truncates the destination before the source is validated, so a failed candidate destroys a previously good artifact. The retry loop usually papers over this, but writing to a temp path and renaming on success would be safer — especially now that a failure here is non-fatal.

build.rs
```diff
@@ -747,21 +749,94 @@ fn static_lib_name(suffix: &str) -> String {
   }
 }
 
-fn static_lib_url() -> String {
-  if let Ok(custom_archive) = env::var("RUSTY_V8_ARCHIVE") {
-    return custom_archive;
+const DEFAULT_MIRROR_BASE: &str =
+  "https://github.com/denoland/rusty_v8/releases/download";
+
+fn is_http_url(s: &str) -> bool {
```
bartlomieju, line 792: Case-sensitive: `HTTP://…` would be treated as a filesystem path and get a bogus flat-layout candidate. `eq_ignore_ascii_case` on the prefix (or `starts_with` on a lowercased copy) is free.

Separately — `expand_mirror_template` is the one genuinely unit-testable piece added here, but it reads `TARGET`/`CARGO_PKG_VERSION`/profile/features from the environment internally. Threading those in as parameters would let you add a case to the existing `mod test` — cheap insurance for a string-substitution function that six env vars now feed.

README.md
```diff
@@ -65,37 +65,101 @@ We default to release builds of `v8` due to performance & CI reasons in `deno`.
...
+full URL/path template instead of a base. The following placeholders are
+substituted:
+
+| Placeholder  | Value                                                        |
+| ------------ | ------------------------------------------------------------ |
+| `{tag}`      | resolved tag (`RUSTY_V8_MIRROR_TAG` or `v<crate version>`)   |
+| `{version}`  | raw crate version, no `v` prefix                             |
+| `{target}`   | Rust target triple                                           |
+| `{profile}`  | `release` or `debug`                                         |
+| `{features}` | ` `, `_ptrcomp`, `_sandbox`, or `_ptrcomp_sandbox`           |
```
bartlomieju: The `{features}` value renders as a lone backtick-space. Better to spell it out: empty, or `_ptrcomp` / `_sandbox` / `_ptrcomp_sandbox`.

The `{profile}` row above is also slightly optimistic — `prebuilt_profile()` returns `release` unless `V8_FORCE_DEBUG=true` **and** the target is not Windows.

bartlomieju (commented):
Direction is good and the refactor reads well. A few things to fix before merge — details inline. The most important are (1) the binding fetch now runs on the `DOCS_RS`/RLS early-exit path, (2) the checksum is written before the copy succeeds, which can cache a truncated artifact, (3) every failure is recoverable so a broken pinned mirror silently falls back to the network, and (4) `RUSTY_V8_MIRROR_STRICT=1` with no mirror set panics with an empty candidate list.

build.rs
bartlomieju: Fixed in 2490aac. The DOCS_RS/RLS early-exit now calls `print_prebuilt_src_binding_path(false)`, which never fetches a missing binding — those contexts fall through to the existing `include!` diagnostic, so docs.rs builds for an unshipped feature combo no longer turn into a hard panic. Added `fs::create_dir_all(parent)` before the binding download to cover a published crate with no `gen/` dir. And you're right about the wording — a plain `cargo build` in a *git checkout* now fetches the binding (that's the intended fix for the confusing `include!` failure); the "zero behaviour change" guarantee only holds for a published crate that ships `gen/<name>.rs`. I've reworded the code comment accordingly.

build.rs
bartlomieju: Fixed in 2490aac. `write_checksum` now runs only after `copy_archive` returns `Ok`, and `copy_archive` stages to a sibling temp + renames on success, so a failed copy leaves the old artifact (and its `.sum`) untouched rather than recording a truncated one. Verified: a corrupt mirror artifact aborts and the previously-good 157MB `.a` is left intact.

build.rs
bartlomieju: Fixed in 2490aac via `FetchError::Miss` vs `FetchError::Fatal`. Missing file / 404 / network / no-downloader are `Miss` (fall through); `copy_archive`/decompress/destination-write failures are `Fatal` (abort). Verified end to end: a present-but-corrupt pinned mirror artifact now aborts with `Found a V8 prebuilt artifact at <url> but could not use it: ...` instead of silently downloading from upstream.

build.rs
bartlomieju: Fixed in 2490aac. Strict mode now only suppresses the upstream fallback when a mirror is actually configured (`mirror.is_some() && env_bool(...)`), so `RUSTY_V8_MIRROR_STRICT=1` with no mirror is a no-op and the list is never empty. Verified: strict + no mirror builds normally via upstream. Documented the nuance in the README strict-mode section.

build.rs
bartlomieju: Fixed in 2490aac. Moved the checksum short-circuit above the http/file split and now write the `.sum` after a filesystem copy too, so a file mirror no longer re-copies/re-decompresses the archive every build. Verified: second build against a valid file mirror reports `Already downloaded` for both artifacts.

build.rs
bartlomieju: Fixed in 2490aac. `copy_archive` writes to `filename.with_extension("copytmp")` and `fs::rename`s into place only after the copy/decompress fully succeeds; the staging file is removed on any error path. A failed candidate can no longer destroy a good artifact.

build.rs
bartlomieju: Fixed in 2490aac. `is_http_url` now lowercases before the prefix check, so `HTTP://…` is treated as a URL. `expand_mirror_template` takes tag/version/target/profile/features/file as params (no env reads) and has a `#[test]` in `mod test` covering full substitution, empty features, and a repeated placeholder.

README.md
bartlomieju: Fixed in 2490aac. `{features}` is now spelled out as "empty, or `_ptrcomp`, `_sandbox`, `_ptrcomp_sandbox`", and `{profile}` reads "`release`, or `debug` only when `V8_FORCE_DEBUG=true` on a non-Windows target".

bartlomieju:
Thanks for the thorough review — all points addressed in 2490aac.

**Summary of changes**
- **Miss vs. Fatal** — new `FetchError::{Miss,Fatal}`. A candidate that is *absent* (missing file, 404, network, no downloader) is a `Miss` and falls through; one that is *found but unusable* (corrupt archive, disk full, unwritable dest) is `Fatal` and aborts. A broken pinned mirror no longer silently escapes to the network.
- **Safe copies** — `copy_archive` stages to a sibling temp and renames on success, so a failed copy/decompress can’t truncate a good artifact. The `.sum` is written only after the copy fully succeeds.
- **Filesystem checksum short-circuit** — moved ahead of the http/file split and now written for filesystem copies too, so file mirrors stop re-copying the ~100MB archive every build.
- **docs.rs/RLS** — that early-exit path no longer fetches a missing binding (passes `fetch_when_missing = false`); `gen/` parent dir is created before a binding download.
- **Strict mode** — no-op when no mirror is set, so the candidate list is never empty.
- **Polish** — case-insensitive `is_http_url`; `expand_mirror_template` is now a pure, parameterised, unit-tested function; README `{features}`/`{profile}`/strict-mode wording corrected.

You were also right that the "zero behaviour change" claim is too strong: a plain `cargo build` in a **git checkout** now fetches the binding by design (that’s the fix for the confusing `include!` failure). It stays a true no-op for a **published crate**, which ships `gen/<name>.rs`.

**Verified locally** (`aarch64-apple-darwin`): `cargo fmt --check` + clippy clean; corrupt flat-mirror artifact aborts `Fatal` with the good `.a` left intact and no `.copytmp` litter; strict + no mirror builds via upstream; file mirror reports `Already downloaded` on rebuild; full `cargo build` links.

bartlomieju:
Addressed the handoff review in `adea640` — decisions resolved as **keep binding auto-fetch here** + **write to OUT_DIR**. Highlights: strict mode now errors explicitly when it leaves no candidates; `cargo:warning=` on any tag/version mismatch; bindings fetched into OUT_DIR (never the source tree) with a new `prebuilt-git-checkout` CI job; cache-origin corruption falls through (`Miss`) while pinned-source corruption aborts (`Fatal`); `artifact_urls` refactored to a pure, table-tested function. Full breakdown in bartlomieju/orchid-inbox#148.

PR #2062, opened 2026-08-14
feat: expose `Module::HasTopLevelAwait`

ursm:
`Module::IsGraphAsync` is already bound, but it answers a different question: it walks the whole linked graph, and it requires the module to be instantiated. `HasTopLevelAwait` is the module's own `[[HasTLA]]` — no dependencies, and no status precondition (`api.cc` reads the bit straight off the `SourceTextModule` and returns `false` for anything else), so it can be called on a module that has only been compiled.

That distinction matters to an embedder implementing a rule the spec states in terms of `[[HasTLA]]`. Disallowing top-level await in a service worker's module script, for instance, is currently only approximable:

- instantiate the whole graph first — which means resolving and fetching every import just to answer the question — and then over-reject on a dependency's await; or
- guess from the source, by compiling it as a classic script and then again inside an `(async () => { ... })` wrapper. That cannot tell `import`-plus-`await` apart from `import` alone, since import declarations fail both probes.

The test covers both halves of the distinction: the answer is available while the module is still `Uninstantiated`, and a module that only *imports* an awaiting one is `is_graph_async()` but not `has_top_level_await()`.

Verified locally with `cargo check --lib` and `cargo fmt --check`. I could not run the test suite here — `binding.cc` is compiled by GN, so a new wrapper needs a from-source V8 build, and the checkout has no `third_party/icu` data for the test targets either. Relying on CI for that.

### DIFF
--- a/src/binding.cc
+++ b/src/binding.cc
@@ -3901,6 +3901,10 @@ bool v8__Module__IsGraphAsync(const v8::Module& self) {
   return ptr_to_local(&self)->IsGraphAsync();
 }
 
+bool v8__Module__HasTopLevelAwait(const v8::Module& self) {
+  return ptr_to_local(&self)->HasTopLevelAwait();
+}
+
 bool v8__Module__IsSourceTextModule(const v8::Module& self) {
   return ptr_to_local(&self)->IsSourceTextModule();
 }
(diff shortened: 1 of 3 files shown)


bartlomieju (approved):
Thanks!

PR #2068, opened 2026-08-18
feat: expose Context::DetachGlobal

### LINKED ISSUE #2059: Expose Context::DetachGlobal()
## Summary

V8 exposes [`v8::Context::DetachGlobal()`](https://chromium.googlesource.com/v8/v8/+/refs/heads/main/include/v8-context.h), but `rusty_v8` currently does not provide the corresponding Rust API.

## Motivation

V8 allows multiple contexts to share a `MicrotaskQueue` and recommends using the same queue for contexts that may access each other synchronously.

An embedder may need to retire one context while keeping the shared queue and the other contexts alive. Dropping the embedder's `Global<Context>` handle is not sufficient to express this lifecycle transition: promises, functions, or other cross-context references may keep the context reachable.

V8 provides `Context::DetachGlobal()` for this purpose. It detaches the context's global proxy and disassociates the context from its microtask queue. Microtasks associated with the detached context are no longer executed, while the shared queue remains usable by other contexts.

This is useful for embedders implementing independently managed realms, reloadable execution environments, or worker-like lifecycles.

The `rusty_v8` documentation for `MicrotaskQueue` already refers to `Context::DetachGlobal()`, but the method itself is not currently exposed.

## Proposed API

```rust
impl Context {
  /// Detaches this context's global proxy and disassociates the context
  /// from its microtask queue.
  pub fn detach_global(&self);
}
```

This would be a thin binding over the existing public V8 API:

```cpp
void v8__Context__DetachGlobal(v8::Context& self) {
  ptr_to_local(&self)->DetachGlobal();
}
```

## Tests

A binding test could create two contexts that share an explicit microtask queue, enqueue work associated with both contexts, detach one context, and verify that:

- work associated with the detached context does not run;
- work associated with the remaining context still runs;
- the shared microtask queue remains usable.

If this API shape is acceptable, I would be happy to prepare a PR.

satouriko:
## Summary

- expose `Context::detach_global(&self)` as a thin binding over V8 `Context::DetachGlobal()`
- document global proxy reuse, idempotence, microtask disassociation, and the lifetime constraint on previously returned queue references
- test two contexts sharing a microtask queue, with tasks enqueued under their corresponding context scopes
- verify that detached-context work is skipped, active-context work still runs, and the shared queue remains reusable

Closes #2059.

## Testing

- `cargo fmt --check`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=third_party/rust-toolchain/lib cargo doc --no-deps --locked`
- `V8_FROM_SOURCE=1 LIBCLANG_PATH=third_party/rust-toolchain/lib cargo nextest run --all-targets --locked --no-fail-fast` (361 passed)

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -28,6 +28,7 @@ unsafe extern "C" {
   ) -> *const Context;
   fn v8__Isolate__GetCurrent() -> *mut RealIsolate;
   fn v8__Context__Global(this: *const Context) -> *const Object;
+  fn v8__Context__DetachGlobal(this: *const Context);
   fn v8__Context__GetExtrasBindingObject(this: *const Context)
   -> *const Object;
   fn v8__Context__GetNumberOfEmbedderDataFields(this: *const Context) -> u32;
@@ -143,11 +144,40 @@ impl Context {
     unsafe { scope.cast_local(|_| v8__Context__Global(self)) }.unwrap()
   }
 
+  /// Detaches this context's global proxy so the global object can be reused
+  /// to create a new context through [`ContextOptions::global_object`]. The new
+  /// context must use the same global template as this context. Calling this
+  /// method more than once has no additional effect.
+  ///
+  /// Detaching resets the context's security token to its default unique state,
+  /// revoking cross-context access previously granted through a shared token.
+  ///
+  /// This also disassociates the context from its microtask queue. Pending
+  /// microtasks associated with this context will not run.
+  ///
+  /// A reference previously returned by [`Self::get_microtask_queue`] does not
+  /// keep the queue alive. With the default `v8_cppgc_microtask_queue=true`,
+  /// detaching removes this context's contribution to keeping the queue alive
+  /// after its [`crate::MicrotaskQueueHandle`] has been dropped. The handle or
+  /// another associated context must keep the queue alive while using any such
+  /// reference.
+  #[inline(always)]
+  pub fn detach_global(&self) {
+    unsafe { v8__Context__DetachGlobal(self) }
+  }
+
   /// Returns the microtask queue associated with this context.
   ///
   /// Returns `None` if the context has no microtask queue, which is the case
   /// after its global object has been detached — detaching clears the native
   /// context's microtask queue pointer.
+  ///
+  /// The returned reference does not own or root the queue. If the context is
+  /// later detached with the default `v8_cppgc_microtask_queue=true`, the
+  /// context no longer keeps the queue alive after its
+  /// [`crate::MicrotaskQueueHandle`] has been dropped. Do not use a reference
+  /// obtained before detachment unless the handle or another associated
+  /// context still keeps the queue alive.
   #[inline(always)]
   pub fn get_microtask_queue(&self) -> Option<&MicrotaskQueue> {
     let queue = unsafe { v8__Context__GetMicrotaskQueue(self) };
--- a/tests/test_api.rs
+++ b/tests/test_api.rs
@@ -13459,6 +13459,121 @@ fn microtask_queue_new() {
 }
 
 #[test]
+fn context_detach_global_reuses_global_object() {
+  let _setup_guard = setup::parallel_test();
+  let mut isolate = v8::Isolate::new(Default::default());
+
+  let scope = pin!(v8::HandleScope::new(&mut isolate));
+  let scope = scope.init();
+  let global_template = v8::ObjectTemplate::new(&scope);
+  let context = v8::Context::new(
+    &scope,
+    v8::ContextOptions {
+      global_template: Some(global_template),
+      ..Default::default()
+    },
+  );
+  let global_object = context.global(&scope);
+
+  context.detach_global();
+  let reused_context = v8::Context::new(
+    &scope,
+    v8::ContextOptions {
+      global_template: Some(global_template),
+      global_object: Some(global_object.into()),
+      ..Default::default()
+    },
+  );
+
+  assert_eq!(global_object, reused_context.global(&scope));
+}
+
+#[test]
+#[should_panic(
+  expected = "cannot set a microtask queue on a context whose global object has been detached"
+)]
+fn context_set_microtask_queue_panics_after_detach() {
+  let _setup_guard = setup::parallel_test();
+  let mut isolate = v8::Isolate::new(Default::default());
+
+  let scope = pin!(v8::HandleScope::new(&mut isolate));
+  let mut scope = scope.init();
+  let queue =
+    v8::MicrotaskQueue::new(&mut scope, v8::MicrotasksPolicy::Explicit);
+  let context = v8::Context::new(&scope, Default::default());
+
+  context.detach_global();
+  context.set_microtask_queue(queue.as_ref());
+}
+
+#[test]
+fn context_detach_global_cancels_its_microtasks() {
+  let _setup_guard = setup::parallel_test();
+  let mut isolate = v8::Isolate::new(Default::default());
+
+  let scope = pin!(v8::HandleScope::new(&mut isolate));
+  let mut scope = scope.init();
+  let queue =
+    v8::MicrotaskQueue::new(&mut scope, v8::MicrotasksPolicy::Explicit);
+
+  let detached_context = v8::Context::new(&scope, Default::default());
+  detached_context.set_microtask_queue(queue.as_ref());
+  let active_context = v8::Context::new(&scope, Default::default());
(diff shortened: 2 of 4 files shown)


src/context.rs
```diff
@@ -143,11 +144,33 @@ impl Context {
...
+  /// to create a new context through [`ContextOptions::global_object`].
+  /// Calling this method more than once has no additional effect.
+  ///
+  /// This also disassociates the context from its microtask queue. Pending
+  /// microtasks associated with this context will not run.
+  ///
+  /// A reference previously returned by [`Self::get_microtask_queue`] does not
+  /// keep the queue alive after detachment. In particular, when
+  /// `v8_cppgc_microtask_queue=false`, the owning
+  /// [`crate::MicrotaskQueueHandle`] must outlive every such reference.
```
bartlomieju: This names the wrong configuration, and it's worth getting right because it inverts the advice.

`v8/gni/v8.gni:98` has `v8_cppgc_microtask_queue = true` as the default, and `src/microtask.rs:114-117` already spells out the two regimes:

- **`=false`** (non-default): the handle owns the queue via `unique_ptr` and "must outlive every associated context." That rule predates this PR, and detaching doesn't change it — no new hazard.
- **`=true`** (the default): the queue is cppgc-managed and *associated contexts keep it alive after the handle is dropped*. `microtask_queue_new` relies on exactly that.

`detach_global` breaks the second invariant. Drop the handle, detach the last associated context, and nothing roots the queue any more — an earlier `&MicrotaskQueue` dangles. So the newly introduced hazard lives precisely in the config this note exempts.

Suggest pointing it at the default cppgc case instead, and amending `MicrotaskQueueHandle`'s doc at `src/microtask.rs:114` to say that detaching a context revokes its contribution to keeping the queue alive — otherwise that doc still reads as an unconditional guarantee.

bartlomieju (commented):
Thanks — the binding is right and the test premise holds up. I checked the skip-detached behavior against the builtin and it's guaranteed rather than incidental (`PrepareForContext` bails when the context's queue pointer is null, under the comment "Skip the microtask execution if the associated context is shutdown"), and the idempotency claim matches `Isolate::DetachGlobal`'s `env->IsDetached()` early return. Nice work getting the microtask association right in the test — enqueueing under each context's own scope is the subtle part, and holding the `MicrotaskQueueHandle` alive for the whole test is the other one.

One thing I'd like fixed before merge: the lifetime note names the wrong GN configuration, and as written it tells readers the default build is the safe one when it's actually the only one this PR changes. Details inline.

Beyond that, a behavioral side effect worth documenting and a few assertions worth adding — none of them large.

src/context.rs
```diff
@@ -143,11 +144,33 @@ impl Context {
...
   ///
   /// Returns `None` if the context has no microtask queue, which is the case
   /// after its global object has been detached — detaching clears the native
   /// context's microtask queue pointer.
+  ///
+  /// The returned reference does not own or root the queue. If the context is
+  /// later detached, do not use the reference after the queue's owning
+  /// [`crate::MicrotaskQueueHandle`] has been dropped. When
+  /// `v8_cppgc_microtask_queue=false`, the handle owns the queue and must
+  /// outlive every returned reference.
```
bartlomieju: Same inversion here — with the default `v8_cppgc_microtask_queue=true`, dropping the handle is explicitly supported (contexts root the queue), which is what makes a later detach able to strand this reference. The `=false` case is the one where the pre-existing handle-outlives-everything rule already covered you.

src/context.rs
```diff
@@ -143,11 +144,33 @@ impl Context {
...
+  ///
+  /// This also disassociates the context from its microtask queue. Pending
+  /// microtasks associated with this context will not run.
+  ///
+  /// A reference previously returned by [`Self::get_microtask_queue`] does not
+  /// keep the queue alive after detachment. In particular, when
+  /// `v8_cppgc_microtask_queue=false`, the owning
+  /// [`crate::MicrotaskQueueHandle`] must outlive every such reference.
+  #[inline(always)]
+  pub fn detach_global(&self) {
```
bartlomieju, line 165: Undocumented side effect worth a line here: `Isolate::DetachGlobal` ends with `env->set_security_token(env->global_object())` — "Set security token back to default (unique) state making sure that only accesses from the same native context are allowed."

Since rusty_v8 exposes `set_security_token`/`use_default_security_token`, an embedder that had granted cross-context access will silently lose it on detach. That's security-relevant behavior, not just trivia, so I'd call it out alongside the microtask disassociation.

tests/test_api.rs
```diff
@@ -13458,6 +13458,71 @@ fn microtask_queue_new() {
...
+  };
+
+  detached_context.detach_global();
+  {
+    let mut context_scope = v8::ContextScope::new(&mut scope, active_context);
+    queue.perform_checkpoint(&mut context_scope);
+  }
+
+  assert_eq!(DETACHED_CALL_COUNT.load(Ordering::SeqCst), 0);
+  assert_eq!(ACTIVE_CALL_COUNT.load(Ordering::SeqCst), 1);
```
bartlomieju, line 13565: Worth adding `assert!(detached_context.get_microtask_queue().is_none());` here. That's the observable contract #2065 introduced and the whole reason the getter returns `Option`, so it should be pinned down by the PR that makes the state reachable.

tests/test_api.rs
```diff
@@ -13458,6 +13458,71 @@ fn microtask_queue_new() {
...
+
+  {
+    let mut context_scope = v8::ContextScope::new(&mut scope, active_context);
+    queue.enqueue_microtask(&mut context_scope, active_task);
+    queue.perform_checkpoint(&mut context_scope);
+  }
+  assert_eq!(ACTIVE_CALL_COUNT.load(Ordering::SeqCst), 2);
```
bartlomieju, line 13573: One more line here — `assert_eq!(DETACHED_CALL_COUNT.load(Ordering::SeqCst), 0);` — proves the detached task was discarded rather than merely deferred past the first checkpoint. Cheap, and it's the assertion that actually distinguishes the two behaviors.

tests/test_api.rs
```diff
@@ -13458,6 +13458,71 @@ fn microtask_queue_new() {
   assert_eq!(CALL_COUNT.load(Ordering::SeqCst), 1);
 }
 
+#[test]
+fn context_detach_global_cancels_its_microtasks() {
```
bartlomieju, line 13510: Two cases I'd like covered in addition to this one:

1. A `#[should_panic]` test for `set_microtask_queue` on a detached context. #2065 added that assert specifically because V8's `Context::SetMicrotaskQueue` unconditionally dereferences the *existing* queue in two ApiChecks, so without the guard this sequence segfaults inside V8 from safe Rust. Now that `detach_global` makes it reachable, a test should lock the guard in.

2. Global-object reuse via `ContextOptions::global_object`. The doc on `detach_global` leads with that as the purpose of the API (as does V8's own header), but nothing exercises it — right now it's an unverified claim. Related: `ContextOptions::global_object` requires the new context to use the same global template, which the detach doc doesn't mention and which a test would make concrete.

src/context.rs
satouriko: Updated the lifetime note to cover the default `v8_cppgc_microtask_queue=true` case, where detachment removes the root contributed by the context. I also amended the `MicrotaskQueueHandle` documentation to make that guarantee end at detachment while retaining the existing `=false` handle-ownership rule.

src/context.rs
satouriko: Updated the getter documentation accordingly. After detachment, the handle or another associated context must continue keeping the queue alive before a previously returned reference is used.

src/context.rs
satouriko: Documented that detachment resets the context security token to its default unique state, revoking cross-context access previously granted through a shared token.

tests/test_api.rs
satouriko: Added the `None` assertion immediately after detachment.

tests/test_api.rs
satouriko: Added another detached-call-count assertion after reusing the queue and performing the second checkpoint.

tests/test_api.rs
satouriko: Added both focused tests: a `#[should_panic]` test for setting a microtask queue after detachment, and a global-object reuse test using `ContextOptions::global_object`. The reuse test passes the same explicit `ObjectTemplate` to both contexts and verifies that the global proxy identity is preserved.

bartlomieju (approved):
LGTM, thanks

PR #2072, opened 2026-09-02
fix: return Value from Function::get_name

ldm0:
Match V8's `Function::GetName` return type instead of assuming every function name is a `String`.

https://v8.github.io/api/head/classv8_1_1Function.html#a145a91e74d5adb78c8b710618253b487


bartlomieju:
Please add a test that exercises this change

ldm0:
@bartlomieju test added.

PR #2081, opened 2026-09-21
feat: add bindings for templates, array buffers and script compilation

bartlomieju:
This expands V8 API coverage across three areas that had no bindings yet.
Everything here is a straight wrapper over an existing V8 entry point; no
existing behaviour changes.

## Templates

`FunctionTemplate::HasInstance` is now available as `has_instance`.

`Template::SetNativeDataProperty` and `Template::SetLazyDataProperty` are
bound on `Template` itself, mirroring how `Template::set` works, so both
`ObjectTemplate` and `FunctionTemplate` can reach them through their
`Deref`. Previously only the `ObjectTemplate` overload of
`SetNativeDataProperty` was bound and there was no binding at all for
`SetLazyDataProperty`.

`v8::DictionaryTemplate` is exposed as a new `Data` subtype with the usual
`Deref`/`TryFrom`/equality impls, plus `new(scope, names)` and
`new_instance(scope, values)`. A `None` entry in `values` is the
`MaybeLocal` hole, i.e. the property is left out of the instance.

The callback-less `FunctionTemplate::New(isolate)` is reachable via
`FunctionTemplate::new_without_callback` and
`FunctionBuilder::new_without_callback`. `FunctionBuilder` now stores an
`Option<FunctionCallback>` internally; that field is private, so this is
not a breaking change.

## ArrayBuffers

`BackingStoreInitializationMode` and `BackingStoreOnFailureMode` are bound
as Rust enums, and the four-argument `NewBackingStore` is available on both
`ArrayBuffer` and `SharedArrayBuffer` as `new_backing_store_with_mode`. The
`kReturnNull` failure mode maps to an `Option` return, so callers can
handle allocation failure instead of taking an OOM crash.

Also added: `ArrayBuffer::maybe_new`,
`ArrayBuffer::new_resizable_backing_store`,
`ArrayBuffer::is_resizable_by_user_javascript`,
`ArrayBuffer::is_immutable`, `ArrayBuffer::MAX_BYTE_LENGTH`,
`BackingStore::max_byte_length`, and `SharedArrayBuffer::data` mirroring
`ArrayBuffer::data`.

Every typed array subclass and `DataView` gained a
`new_with_shared_buffer` constructor for the `SharedArrayBuffer` overload
of `New`.

`ArrayBuffer::Allocator::NewDefaultAllocator(IsolateGroup&)` is bound as
the free function `new_default_allocator_for_group`. V8 only declares that
overload when built with pointer compression in multi-cage mode, which is
exactly the configuration `IsolateGroup::can_create_new_groups` reports on,
so the wrapper returns an `Option` and yields `None` otherwise.

## Scripts, modules and isolates

`ScriptCompiler::CachedData::CompatibilityCheck` is bound as
`compatibility_check`, along with an exported `CompatibilityCheckResult`
enum mirroring V8's. `ScriptCompiler::CompileOptionsIsValid` is exposed as
`CompileOptions::is_valid`.

`Isolate::ContextDisposedNotification` is bound against the current,
non-deprecated signature that takes `v8::ContextDependants`, and that enum
is exported too. `Context::SetTemporalHostSystemUTCEpochNanosecondsCallback`
is bound with a `TemporalHostSystemUTCEpochNanosecondsCallback` type alias.

## Skipped

`PropertyCallbackInfo::This()` does not exist in the pinned V8 headers;
`PropertyCallbackInfo` only offers `Holder()` (and the deprecated
`HolderV2()`), which was already bound. `PropertyCallbackInfo::GetIsolate`
is left unbound as well, since property callbacks already receive a scope
and routing isolate access through it is what the rest of the crate does.

`ArrayBuffer::MAX_BYTE_LENGTH` is derived from the already-exported
`v8__TypedArray__kMaxByteLength`, which V8 defines as exactly
`ArrayBuffer::kMaxByteLength`. This keeps `src/binding.hpp` untouched, so
the bindgen step is unaffected by this change.

## Testing

Tests were added to `tests/test_api.rs` for each new API, including the
dictionary template hole behaviour, typed arrays and `DataView` over a
`SharedArrayBuffer`, both backing store modes with a fallible allocation,
resizable backing stores driven from JavaScript, the cached data
compatibility check against both valid and garbage input, and the Temporal
callback observed through `Temporal.Now`.

Verified locally on aarch64-apple-darwin against a from-source V8 build:
the full `cargo test --release` suite passes (319 tests in `test_api`, all
other test targets green), along with `cargo fmt --check`, `clang-format
--Werror --dry-run src/binding.cc` and `cargo clippy --all-targets --locked
--release -- -D clippy::all`. The local toolchain could not run the bindgen
step, which needs Clang 21, so that path is covered by CI only; since
`src/binding.hpp` is unchanged, the generated bindings should be identical.


bartlomieju:
Rebased onto latest main. The only conflict was in `tests/test_api.rs`, where
both this branch and the recently landed bindings PR appended test functions at
the end of the file; both sets are kept.

This also fixes the CI failures from the previous run. `array_buffer_maybe_new`
and `array_buffer_backing_store_modes` aborted in DCHECK-enabled builds (debug
ptrcomp and asan) because they asked for exactly `kMaxByteLength` bytes, which
V8 accepts as a valid length and then tries to allocate. The tests now ask for
`kMaxByteLength + 1`, so V8 takes its early bail-out branch
(`if (byte_length > max) return {}` in `MaybeNew`, and the equivalent
`kReturnNull` check in `NewBackingStore`) without attempting a huge allocation.
That is the branch these bindings are meant to cover, and it behaves the same
way in every build configuration. The same change was applied to the
`SharedArrayBuffer` variant, which had not been reached before CI bailed out.

